PR #37, opened 2026-02-03
feat: expose workspaceId in project JSON output

rmartins90:
Add workspaceId to PROJECT_ESSENTIAL_FIELDS so that workspace projects include their workspace ID in JSON/NDJSON output. Personal projects won't have this field, matching the implementation from todoist-ai PR #262.

This allows scripts and agents to determine which workspace a project belongs to when using JSON output format.


scottlovegrove (commented):
👍🏻

PR #53, opened 2026-02-10
feat: auto-retry id-like task refs as direct ID lookups

### LINKED ISSUE #46: Unclear param formats for filters can lead to confusion
Edit: There was confusion on my part. `--parent` works, but there's still an issue here IMO

When searching for subtasks of a parent task, the cli help says:
`--parent <ref>      Filter subtasks of a parent task`

This option only works if the `<ref>` is of the form `id:XXXWVCX78mQ7CXXX`.

This is confusing and can lead to misinterpretation by both humans and agents. The help could be more explicit about this, but better would be for refs of the form `XXXWVCX78mQ7CXXX` (without id:) to work throughout.

---

Previous (incorrect) description:
~Client side task list filters (like --parent) may not find relevant tasks #46~

~Task queries that filter client-side are prone to missing expected results. I initially thought this was due to a lack of pagination but I see the code attempts to handle that, so this needs further investigation. I'll try to dig into this more with a concrete test case I have.~

Example (ids obfuscated):
``` sh
% td task list --parent XXXWVCX78mQ7CXXX
Error: TASK_NOT_FOUND
task "XXXWVCX78mQ7CXXX" not found.
```

Contrast with `todoist-ai` MCP server
``` sh
% mcporter call 'todoist.find-tasks(parentId:"XXXWVCX78mQ7CXXX")'
{
  "tasks": [
    {
      "id": "XXXh3WRHVJQw9XXX",
      "content": "test",
      "description": "",
      "recurring": false,
      "priority": "p4",
      "projectId": "XXXCPW8j2R2ccXXX",
      "parentId": "XXXWVCX78mQ7CXXX",
      "labels": [],
      "checked": false
    }
  ],
  "totalCount": 1,
  "hasMore": false,
  "appliedFilters": {
    "parentId": "XXXWVCX78mQ7CXXX",
    "limit": 10
  }
}
```
~

mindsocket:
Fixes #46

- When `resolveTaskRef` fails to match a ref by name search and the ref looks like a raw Todoist ID (per `looksLikeRawId` from #49), automatically retry with `id:` prefix before erroring
- The `idPrefixHint` warnings from #49 no longer apply in this case. Since they still apply for other entity types, it has not been touched.

**Note**: Depends on #52 — this PR is built on top of server-side search/filtering improvements. While it could be made independent if is more easily merged after it. See ab3bdc21086f54dde7b081384a9f8231e00d3dbe for the changes specific to this PR.

## Test plan

- [x] Existing tests pass (675/675)
- [x] New test: alphanumeric ref auto-retries as direct ID lookup
- [x] New test: numeric ref auto-retries as direct ID lookup
- [x] New test: plain text ref does not auto-retry

### DIFF
--- a/src/__tests__/refs.test.ts
+++ b/src/__tests__/refs.test.ts
@@ -146,22 +146,40 @@ describe('resolveTaskRef', () => {
         await expect(resolveTaskRef(api, 'nonexistent')).rejects.toThrow('not found')
     })
 
-    it('hints about id: prefix when ref looks like a raw ID', async () => {
+    it('auto-retries id-like refs as direct ID lookup', async () => {
         const api = createMockApi({
+            getTask: vi.fn().mockResolvedValue(tasks[0]),
             getTasksByFilter: vi.fn().mockResolvedValue({ results: [], nextCursor: null }),
         })
 
-        await expect(resolveTaskRef(api, '6fmg66Fr27R59RPg')).rejects.toThrow('id:6fmg66Fr27R59RPg')
+        // Alphanumeric mix (e.g. Todoist task ID pasted without id: prefix)
+        const result = await resolveTaskRef(api, '6fmg66Fr27R59RPg')
+        expect(result.id).toBe('task-1')
+        expect(api.getTasksByFilter).toHaveBeenCalledWith(
+            expect.objectContaining({ query: 'search: 6fmg66Fr27R59RPg' }),
+        )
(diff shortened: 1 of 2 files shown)


scottlovegrove:
@mindsocket Would you mind rebasing this branch on main, there are merge conflicts.

mindsocket:
@scottlovegrove done!

PR #66, opened 2026-02-13
feat: shell completion support (bash, zsh, fish)

jacobo-doist:
## Summary

- Add shell completion support for bash, zsh, and fish via `td completion install/uninstall`
- Completion engine walks the Commander.js command tree to suggest commands, subcommands, options, and enum values
- Options that accept a fixed set of values (priority, theme, event type, etc.) complete those values at the shell and accept case-insensitive input
- Choice values are derived from the same constants used by parsers, so completions stay in sync with accepted values automatically
- Validate comma-separated values for options like `--role` against their allowed values, rejecting invalid ones at parse time
- Known limitation: quoted multi-word arguments that happen to contain a subcommand name can confuse the tree walker (documented, tested, tracked upstream as https://github.com/pnpm/tabtab/issues/35)

## Test plan

- [x] `npm run build` compiles clean
- [x] `npm test` — all 792 tests pass
- [x] Manual: `td completion install` for bash/zsh/fish, verify completions work in a new shell session
- [x] Manual: verify enum value completions for `--priority`, `--theme`, `--event`, etc.
- [x] Manual: `--flag=<TAB>` completes correctly across all three shells
- [x] Manual: case-insensitive input accepted (`--start-day Sunday`, `--priority P1`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
   ],
   "dependencies": {
     "@doist/todoist-api-typescript": "6.2.1",
-    "@pnpm/tabtab": "^0.5.4",
+    "@pnpm/tabtab": "0.5.4",
     "chalk": "5.6.2",
     "commander": "14.0.2",
     "date-fns": "4.1.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@doist/todoist-api-typescript": "6.2.1",
-        "@pnpm/tabtab": "^0.5.4",
+        "@pnpm/tabtab": "0.5.4",
         "chalk": "5.6.2",
         "commander": "14.0.2",
         "date-fns": "4.1.0",


src/types/pnpm-tabtab.d.ts
```diff
@@ -0,0 +1,48 @@
+declare module '@pnpm/tabtab' {
```
scottlovegrove: What do we need this types file for? Looking at the `@pnpm/tabtab` package, it already comes with its own types.

jacobo-doist: We do not, thanks :)

src/commands/completion.ts
```diff
@@ -0,0 +1,129 @@
+import { existsSync } from 'node:fs'
+import { homedir } from 'node:os'
+import { join } from 'node:path'
+import { Command } from 'commander'
+import { getCompletions, parseCompLine } from '../lib/completion.js'
+
+// pwsh is included because tabtab supports it — installedShells() needs to
+// detect and clean up pwsh completion files even though we don't advertise it.
+const COMPLETION_EXTENSIONS: Record<string, string> = {
```
scottlovegrove:
```suggestion
const COMPLETION_EXTENSIONS: Record<SupportedShell, string> = {
```

`SupportedShell` can be imported from `@pnpm/tabtab`

src/commands/completion.ts
```diff
@@ -0,0 +1,129 @@
...
+                    `Unsupported shell: ${shell}. Supported: ${tabtab.SUPPORTED_SHELLS.join(', ')}`,
+                )
+                process.exitCode = 1
+                return
+            }
+
+            await tabtab.install({
+                name: 'td',
+                completer: 'td',
+                shell: shell as 'bash' | 'zsh' | 'fish' | undefined,
```
scottlovegrove:
```suggestion
                shell: shell as SupportedShell,
```

jacobo-doist:
> Very nice, I also like that when provided with no shell type it gives you a choice

That's a [tabtab](https://github.com/mklabs/tabtab) goodie, and yes it's cool (next level would be to detect the running shell and suggest it as default)

src/lib/skills/content.ts
```diff
@@ -24,6 +24,7 @@ Use this skill when the user wants to interact with their Todoist tasks.
 - \`td reminder add\` - Task reminders
 - \`td stats\` - Productivity stats
 - \`td settings view\` - User settings
+- \`td completion install\` - Install shell completions
```
jacobo-doist, line 27: This is here for the sake of completion (ahem 🤡) but I'm really hesitant to include it because it is not relevant to agents unless to advice on how to set this up for a human operator.

scottlovegrove: I think asking an agent "how do you get shell autocomplete with the Todoist cli" is a valid enough use case and justification for keeping it there 👍🏻

src/commands/project.ts
```diff
@@ -15,6 +16,8 @@ import { LIMITS, paginate } from '../lib/pagination.js'
 import { resolveProjectRef } from '../lib/refs.js'
 import { projectUrl } from '../lib/urls.js'
 
+const VIEW_STYLE_CHOICES: ProjectViewStyle[] = ['list', 'board']
```
scottlovegrove: Any reason we aren't listing `calendar` here?

package.json
```diff
@@ -47,6 +47,7 @@
   ],
   "dependencies": {
     "@doist/todoist-api-typescript": "6.2.1",
+    "@pnpm/tabtab": "^0.5.4",
```
scottlovegrove: nit: We should pin it to this version. If you remove the `^`, then do `npm i` it will copy the change over to `package-lock.json`.

src/commands/project.ts
jacobo-doist: missed it, I checked against the spec and I found other two missing options for event types in activity too. Coming soon.

PR #71, opened 2026-02-14
chore: Ensures CI checks done on *all* PRs

scottlovegrove:
Currently linting and tests are done when: 
- a commit happens on `main`
- A PR with `main` as its base is opened

Neither of these are really right.

- We don't need to run the test/linting on main separately because you can only commit to main via PR, and the CI checks are done on PRs
- A PR whose base is not main would not have the checks done, they should be.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,10 +1,7 @@
 name: Lint
 
 on:
-  push:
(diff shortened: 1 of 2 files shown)


pawelgrimm (commented):
🚀

Issue #84, opened 2026-02-19
feat(activity): enforce hard cap when using --all


gnapse:
## Problem

The activity command currently treats `--all` as effectively unbounded (`Number.MAX_SAFE_INTEGER`) and accumulates all fetched events in memory before printing.

For large histories this can lead to very long runtimes and excessive memory usage.

## Proposed change

Add a hard upper cap for total fetched activity events, enforced even when `--all` is passed.

## Notes

- Keep current behavior for smaller result sets.
- When the cap is hit, return partial results and a clear footer/note indicating truncation and how to narrow filters (e.g. date range, project, event type).
- This should apply consistently across text, markdown, JSON, and NDJSON output modes.

## Context

This came up while shipping `td activity --markdown` and identifying risk around unbounded exports.

PR #85, opened 2026-02-19
feat(activity): add markdown output mode for activity logs

gnapse:
## Summary
- add `--markdown` output mode to `td activity`
- keep existing default text output and `--json` / `--ndjson` behavior unchanged
- enforce output-format exclusivity across `--markdown`, `--json`, and `--ndjson`
- add tests for markdown output and conflicting-format validation
- update skill docs with a `td activity --markdown` example

## Why now

We want a simple LLM-friendly export path immediately, without changing the existing human-scannable terminal output.

## Limitations vs Todoist Web

This CLI implementation intentionally ships a simpler format for now. Compared to the Todoist Web markdown export:

- uses a simpler event/object/content line format instead of the richer sentence-template formatter
- uses YYYY-MM-DD day headers instead of localized/relative day labels
- does not yet mirror all of web's text-normalization nuances

These are known and acceptable for this iteration. We plan to iterate and bring CLI output closer to the web export format in an upcoming follow-up.

## Follow-ups

- hard-cap `--all` to avoid unbounded fetch/memory behavior: #84

## Validation

- `npm test -- src/__tests__/activity.test.ts`
- pre-push hook test suite passed during push

### DIFF
--- a/src/commands/activity.ts
+++ b/src/commands/activity.ts
@@ -26,6 +26,7 @@ interface ActivityOptions {
     limit?: string
     cursor?: string
     all?: boolean
+    markdown?: boolean
     json?: boolean
     ndjson?: boolean
     full?: boolean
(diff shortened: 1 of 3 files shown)


src/commands/activity.ts
```diff
@@ -129,6 +130,75 @@ function formatActivityRow(
...
+    const day = String(date.getDate()).padStart(2, '0')
+    return `${year}-${month}-${day}`
+}
+
+function formatMarkdownTime(eventDate: string): string {
+    const date = new Date(eventDate)
+    const hours = String(date.getHours()).padStart(2, '0')
+    const minutes = String(date.getMinutes()).padStart(2, '0')
+    return `${hours}:${minutes}`
+}
```
scottlovegrove, line 150: Are we just assuming 24hr times here? Or deliberately forgoing am/pm?

gnapse: Nothing deliberate here, just overlooked. We could use the user's preference, although if this is mostly for AI consumption not sure if it's worth the effort. AI does not care.

scottlovegrove: > AI does not care.

It might not care about the user's choice, but it should still be told whether it's am or pm really, if I have a task at 9:32, which is it? The AI will only hallucinate a guess, no?

gnapse: This would output `09:32`, which I think is standard for AM times in 24 hours mode.

PR #88, opened 2026-02-20
feat: add `td update` self-update command

scottlovegrove:
## Summary

- Adds `td update` command that checks the npm registry for the latest version and runs `npm install -g` to self-update
- `--check` flag to see if an update is available without installing
- Shows a spinner while checking the registry so it doesn't look like it's hanging
- Handles permission errors (suggests `sudo`), network failures, and non-zero npm exit codes
- Best-effort pnpm detection (falls back to npm)

## Test plan

- [x] `npm run type-check` passes (pre-existing errors in `project.ts` only)
- [x] `npx biome check` passes on new files
- [x] `npm test` — all 914 tests pass (8 new tests for update command)
- [x] Manual: `npm run build && node dist/index.js update --check` — shows version info

### DIFF
--- a/src/commands/update.ts
+++ b/src/commands/update.ts
@@ -67,7 +67,7 @@ export async function updateAction(options: { check?: boolean }): Promise<void> 
     }
 
     const pm = detectPackageManager()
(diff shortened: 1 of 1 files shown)


frankieyan:
Tested with v1.18.0 ✅

PR #110, opened 2026-03-01
Add a read-only OAuth mode for safe autonomous tool use

dthvt:
### Motivation

- Add an OAuth flow option for read-only integration so users can authenticate with a least-privilege token. 
- Prevent accidental data mutation when the CLI is authenticated with a read-only OAuth token. 

### Description

- Introduces read-only OAuth support by adding a `--read-only` option to `td auth login`, extending `buildAuthorizationUrl` to accept a `readOnly` option, and saving auth metadata (`auth_mode` and `auth_scope`) alongside tokens in the config. 
- Treats externally provided tokens (environment variable or manual `td auth token`) as `unknown` mode and persists `auth_mode: 'unknown'` when saving manual tokens. 
- Adds a `permissions` module that identifies mutating API calls and sync payloads and exposes `ensureWriteAllowed` which throws when the stored auth mode is `read-only`. 
- Wraps the Todoist API client in `createSpinnerWrappedApi` to run `ensureWriteAllowed` before mutating calls, preserve spinner/progress behavior, and emit progress events and errors consistently. 
- Updates CLI docs and help content (`README.md` and skill content) to document the read-only login flow and auth metadata behavior. 
- Updates and adds unit tests to cover the new behavior and metadata handling. 

### Testing

- Ran unit tests with `vitest` including modified `src/__tests__/auth.test.ts` and new tests `src/__tests__/oauth.test.ts` and `src/__tests__/permissions.test.ts`, and all tests passed. 
- Verified the OAuth authorization URL uses `data:read` when `--read-only` is passed via the new `buildAuthorizationUrl` unit test. 
- Verified that write operations are blocked when `getAuthMetadata` reports `authMode: 'read-only'` via the `permissions` unit tests.


scottlovegrove (commented):
@dthvt In general, I approve of this PR, however, we have just merged a change in that makes use of storing auth details at the OS level (#120), as well as how we were doing it before, this might mean some minor changes will be required in this PR to cater for this

Issue #119, opened 2026-03-09
Add --dry-run flag?


nats12:
# Context

Inspired by [Rewrite Your CLI for AI Agents](https://justin.poehnelt.com/posts/rewrite-your-cli-for-ai-agents/?utm_source=tldrdev) - `--dry-run` is a well-established CLI convention that lets users preview what a command would do without actually executing it. With the CLI increasingly being used by AI agents, this becomes especially valuable (agents can preview actions, show the user what would happen, and only execute after confirmation).

# Proposal

Should we add support for the `--dry-run` flag on **mutating** commands (add, update, delete, complete, move)?  When passed, the CLI would run all query/parse logic normally but skip the API call, instead printing a summary of what would happen.

## Why it's useful

  - Catches **misinterpretations** - especially with natural language parsing in `td add`, dry run lets you verify the CLI understood you correctly before creating the task
  - Safe bulk operations - filter-based commands could touch dozens of tasks; dry run acts as a guardrail
  - Agent-friendly - AI agents calling td on a user's behalf can preview actions and get confirmation before mutating state

## Examples

  **Natural language parsing verification**

  `td add "Review Q3 report every Friday at 2pm #Work @high-priority" --dry-run`

  Would create:
  > Content:  Review Q3 report
  > Project:  Work
  > Due:      Every Friday at 14:00 (recurring)
  >  Priority: p1
  > Run without --dry-run to create.

   **Ambiguous task references**

  `td task complete "Review" --dry-run`

   > Multiple matches found. Would complete:
   > Review Q3 report (Project: Work, due: Mar 12)
   > Review PR #482 (Project: Engineering, due: Mar 10)
   > Review Sarah's proposal (Project: Work, due: Mar 15)
   > 3 tasks matched. Run without --dry-run to complete all, or be more specific.

  This is especially useful when you have tasks with similar names, without dry run, `td task complete "Review"` might complete the wrong one (or all of them). With dry run, you see exactly which tasks matched can refine your query.

 ## Open questions

  - Is this worth the added complexity, or is the current workflow (list/filter first, then act) sufficient?
  - Should it apply to all mutating commands or just the high-risk ones (delete, bulk update)?
  - Output format - plain text summary, or structured JSON (via --json --dry-run) so agents can parse it?

PR #156, opened 2026-03-16
chore: Add support for dry-run flag on mutating commands

nats12:
## Summary

- Adds `--dry-run` flag to all mutating commands (add, task, project, label, filter, comment, section, reminder)
- Shows a preview of the action and parameters without executing the mutation
- **On destructive commands that require `--yes` (delete, project move), `--dry-run` takes precedence — the action won’t execute even with `--yes`**
- Updates `SKILL_CONTENT` with `--dry-run` documentation and per-section examples

## Related
https://github.com/Doist/todoist-cli/issues/119

## Test plan

- [ ] Try mutating commands across all models, the command should not action - instead a preview of what is about to be carried out is shown

### `task complete`
Screenshot 2026-03-16 at 15 30 34

### `task add`
Screenshot 2026-03-16 at 15 30 39

### `task delete` (with `--yes`)
Screenshot 2026-03-16 at 16 14 29

### DIFF
--- a/src/commands/comment.ts
+++ b/src/commands/comment.ts
@@ -200,15 +200,15 @@ async function deleteComment(
 ): Promise<void> {
     const id = lenientIdRef(commentId, 'comment')
 
-    if (options.dryRun) {
-        printDryRun('delete comment', { ID: id })
-        return
-    }
-
     const api = await getApi()
     const comment = await api.getComment(id)
     const preview =
         comment.content.length > 50 ? `${comment.content.slice(0, 50)}...` : comment.content
+
+    if (options.dryRun) {
+        printDryRun('delete comment', { Comment: preview })
+        return
+    }
 
     if (!options.yes) {
         console.log(`Would delete comment: ${preview}`)
--- a/src/commands/filter.ts
+++ b/src/commands/filter.ts
@@ -145,12 +145,12 @@ async function deleteFilterCmd(
     nameOrId: string,
     options: { yes?: boolean; dryRun?: boolean },
 ): Promise<void> {
+    const filter = await resolveFilterRef(nameOrId)
+
     if (options.dryRun) {
-        printDryRun('delete filter', { ID: nameOrId })
+        printDryRun('delete filter', { Filter: filter.name, Query: filter.query })
         return
     }
-
-    const filter = await resolveFilterRef(nameOrId)
 
     if (!options.yes) {
         console.log(`Would delete: ${filter.name}`)
--- a/src/commands/label.ts
+++ b/src/commands/label.ts
@@ -135,12 +135,12 @@ async function deleteLabel(
     nameOrId: string,
     options: { yes?: boolean; dryRun?: boolean },
 ): Promise<void> {
+    const label = await resolveLabelRef(nameOrId)
+
     if (options.dryRun) {
-        printDryRun('delete label', { ID: nameOrId })
+        printDryRun('delete label', { Label: `@${label.name}` })
         return
     }
-
-    const label = await resolveLabelRef(nameOrId)
 
     if (!options.yes) {
         console.log(`Would delete: @${label.name}`)
--- a/src/commands/project.ts
+++ b/src/commands/project.ts
@@ -226,11 +226,6 @@ async function deleteProject(
     ref: string,
     options: { yes?: boolean; dryRun?: boolean },
 ): Promise<void> {
-    if (options.dryRun) {
-        printDryRun('delete project', { Project: ref })
-        return
-    }
-
     const api = await getApi()
     const project = await resolveProjectRef(api, ref)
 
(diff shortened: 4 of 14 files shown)


src/lib/skills/content.ts
```diff
@@ -180,6 +185,10 @@ td task move "task name" --parent "Parent task"
 td task move "task name" --no-parent          # Move to project root
 td task move "task name" --no-section         # Remove from section
 
+# Dry run (preview any mutating command without executing)
+td task add "New task" --due "tomorrow" --dry-run  # Preview task creation
+td task delete "task name" --dry-run               # Preview deletion (no --yes needed)
```
scottlovegrove: `(no --yes needed)` sounds messy. Perhaps this could be clarified a bit better. Or remove it entirely (which is probably all that needs to happen).

src/lib/skills/content.ts
```diff
@@ -204,6 +213,8 @@ td project move "Project Name" --to-workspace "Acme" --folder "Engineering"
 td project move "Project Name" --to-workspace "Acme" --visibility team
 td project move "Project Name" --to-personal
 # move requires --yes to confirm (without it, shows a dry-run preview)
+td project create --name "New Project" --dry-run  # Preview project creation
+td project delete "Project Name" --dry-run        # Preview deletion (no --yes needed)
```
scottlovegrove: Ditto here.

src/lib/skills/content.ts
nats12: @scottlovegrove removed 👍🏼

src/commands/comment.ts
```diff
@@ -182,9 +194,18 @@ async function addComment(ref: string, options: AddOptions): Promise<void> {
...
-async function deleteComment(commentId: string, options: { yes?: boolean }): Promise<void> {
-    const api = await getApi()
+async function deleteComment(
+    commentId: string,
+    options: { yes?: boolean; dryRun?: boolean },
+): Promise<void> {
     const id = lenientIdRef(commentId, 'comment')
+
+    if (options.dryRun) {
+        console.log(`Would delete comment: ${id}`)
```
pauloslund:
```suggestion
        printDryRun(`delete comment`, { ID: id })
```

src/commands/filter.ts
```diff
@@ -129,7 +141,15 @@ async function resolveFilterRef(nameOrId: string): Promise<Filter> {
     throw new Error(formatError('FILTER_NOT_FOUND', `Filter "${nameOrId}" not found.`))
 }
 
-async function deleteFilterCmd(nameOrId: string, options: { yes?: boolean }): Promise<void> {
+async function deleteFilterCmd(
+    nameOrId: string,
+    options: { yes?: boolean; dryRun?: boolean },
+): Promise<void> {
+    if (options.dryRun) {
+        console.log(`Would delete filter: ${nameOrId}`)
```
pauloslund:
```suggestion
        printDryRun('delete filter',  {ID: nameOrId })
```

src/commands/label.ts
```diff
@@ -120,7 +131,15 @@ async function createLabel(options: CreateOptions): Promise<void> {
     console.log(chalk.dim(`ID: ${label.id}`))
 }
 
-async function deleteLabel(nameOrId: string, options: { yes?: boolean }): Promise<void> {
+async function deleteLabel(
+    nameOrId: string,
+    options: { yes?: boolean; dryRun?: boolean },
+): Promise<void> {
+    if (options.dryRun) {
+        console.log(`Would delete label: ${nameOrId}`)
```
pauloslund: Before I comment the same suggestion, is there a reason the delete commands aren't using `printDryRun` that I'm missing?

src/commands/project.ts
```diff
@@ -221,7 +222,15 @@ export async function viewProject(ref: string, options: ViewOptions = {}): Promi
     }
 }
 
-async function deleteProject(ref: string, options: { yes?: boolean }): Promise<void> {
+async function deleteProject(
+    ref: string,
+    options: { yes?: boolean; dryRun?: boolean },
+): Promise<void> {
+    if (options.dryRun) {
```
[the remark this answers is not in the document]
nats12: @scottlovegrove @pauloslund regarding this point, I feel that `--dry-run` should stay offline only i.e. exist just to show a preview. Validation should happen after once the user agrees. Do you agree?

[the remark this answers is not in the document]
scottlovegrove: > Do you agree?

Not really. To me, a dry run should do everything a normal run should do, with the exception of the actual final change. Otherwise it's not really a dry run, it's just parrotting what the user has inputted, which could be garbage.

My go to example when I'm thinking about this is when npm, when you want to package the files, or even publish, it does everything (building, processing, whatever) but just doesn't do the last step (create a package, publish the package). https://docs.npmjs.com/cli/v11/commands/npm-publish#files-included-in-package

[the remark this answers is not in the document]
nats12: Fair point 👍🏼

PR #176, opened 2026-03-25
feat: add new agent skills

rfgamaral:
## Overview

Added support for pi and universal agents, and standardized the README structure to be consistent across Doist CLIs. The universal agent installs to `~/.agents/` for compatibility with Amp, OpenCode, and similar tools. The installer skips the agent directory existence check for universal since there's no specific agent to verify.

The `postinstall` script is now guarded against missing `dist/` on fresh checkouts, avoiding noisy errors during `npm install` before the first build.

### Reference
- Doist/bob-cli#17
- Doist/twist-cli#101

## Test plan
- [ ] README renders correctly on GitHub

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint:write": "oxlint src --fix",
     "format": "oxfmt",
     "format:check": "oxfmt --check",
-    "postinstall": "node dist/postinstall.js || true",
+    "postinstall": "test -f dist/postinstall.js && node dist/postinstall.js || true",
     "test": "vitest run",
     "test:watch": "vitest"
   },


package.json
```diff
@@ -16,7 +16,7 @@
     "lint:write": "oxlint src --fix",
     "format": "oxfmt",
     "format:check": "oxfmt --check",
-    "postinstall": "node dist/postinstall.js || true",
+    "postinstall": "test -f dist/postinstall.js && node dist/postinstall.js || true",
```
scottlovegrove, line 19: @rfgamaral I assume this `test` command works on windows too?

rfgamaral: It does not...

Is this something we should be worried about, @scottlovegrove? I mean, who in their right mind actually uses Windows (not WSL)? 😅

Regardless, I can work on fixing this across all our CLIs, if you think it's important. Let me know.

scottlovegrove: > Is this something we should be worried about, @scottlovegrove? I mean, who in their right mind actually uses Windows (not WSL)? 😅

Absolutely something we should be concerned with. Doist OS is meant to run on Windows (not just WSL), and it relies on these CLIs, so we definitely need to make sure these are supported and won't fail.

rfgamaral: Good point. I'll take care of it.

PR #187, opened 2026-03-27
Add notify-uids support for comment creation

kanato-masayoshi:
Add support for notifying specific Todoist users when creating a task/project comment from the CLI.

- adds a `--notify-uids` option to `td comment add`
- routes notified-comment creation through Todoist API v1 so `uids_to_notify` is actually sent
- keeps the existing comment creation path unchanged when no notifications are requested

特に確認してもらいたいところ/懸念点:
- Todoist API v1 の `uids_to_notify` を使う実装にしている。public REST v2 client では直接扱えないため、通知付きコメント時のみ v1 を直接叩く形にした
- 返却 payload は最小限の型付けにしているので、必要なら後で型をもう少し厳密にしてもいい

Auto-generated by OpenClaw
@kanato-masayoshi レビューよろしく

### DIFF
--- a/src/lib/api/comments.ts
+++ b/src/lib/api/comments.ts
@@ -1,0 +1,57 @@
+import { getApiToken } from '../auth.js'
+
+export interface CreateCommentV1Args {
+    taskId?: string
+    projectId?: string
+    content: string
+    attachment?: unknown
+    uidsToNotify?: string[]
+}
+
+export interface CommentV1Response {
+    id: string
+    posted_uid: string
+    content: string
+    file_attachment: unknown
+    uids_to_notify: string[] | null
+    is_deleted: boolean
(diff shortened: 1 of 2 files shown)


src/lib/api/comments.ts
```diff
@@ -0,0 +1,57 @@
...
+        }),
+    })
+
+    const text = await res.text()
+    if (!res.ok) {
+        throw new Error(`API v1 create comment failed: ${res.status} ${text}`)
+    }
+
+    return JSON.parse(text) as CommentV1Response
+}
```
scottlovegrove, line 57: We should really just be using the Todoist SDK for adding comments, but it does look like the `uids_to_notify` field is missing. I shall add this to the SDK now.

PR #193, opened 2026-03-29
chore: add developer skill for adding new CLI commands

scottlovegrove:
## Summary

- Adds an agent-agnostic developer skill (`.agents/skills/add-command/`) with the full checklist for adding new CLI commands: mock API, spinners, implementation, flag conventions, ID resolution patterns, tests, skill content sync, and verification
- Claude Code skill (`.claude/skills/add-command/`) references the shared one so `/add-command` works
- Updates `sync:skill` script to run `npm run build` first, so it's a single command to regenerate `SKILL.md`

## Test plan

- [x] All 1063 tests pass (main branch, no feature changes)
- [x] `npm run sync:skill` now builds before syncing
- [ ] Verify `/add-command` loads in Claude Code

### DIFF
--- a/.agents/skills/add-command/SKILL.md
+++ b/.agents/skills/add-command/SKILL.md
@@ -30,18 +30,19 @@ Add the async function and register the subcommand.
 
 ### Flag conventions
 
-| Command type                   | Flags                               |
-| ------------------------------ | ----------------------------------- |
-| Read-only                      | `--json` (and `--ndjson` for lists) |
-| Mutating (returns entity)      | `--json`, `--dry-run`               |
-| Mutating (no return)           | `--dry-run`                         |
-| Destructive + irreversible     | `--yes`, `--dry-run`                |
-| Reversible (archive/unarchive) | `--dry-run` (no `--yes`)            |
+| Command type                   | Flags                                    |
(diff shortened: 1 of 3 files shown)


.agents/skills/add-command/SKILL.md
```diff
@@ -0,0 +1,103 @@
...
+    .command('my-action [ref]')
+    .description('Do something')
+    .option('--json', 'Output as JSON')
+    .option('--dry-run', 'Preview what would happen without executing')
+    .action((ref, options) => {
+        if (!ref) {
+            myCmd.help()
+            return
+        }
+        return myAction(ref, options)
```
pedroalves0, line 60: Should the skill look into the API SDK repo, or at least check the SDK types, and analyze what the best SDK method is for the command being implemented? Or does it already do that without any additional prompting?

scottlovegrove: Maybe. I've been thinking about shared stuff for the CLIs anyway, because there have been quite a few times where I make a change in one, and it's something that would benefit all of them, and you have to go through and make the changes in each, but I'm unsure the best approach (I did consider a CLI monorepo, but I wasn't wholly convinced).

PR #204, opened 2026-03-30
feat!: migrate completed command to getAllCompletedTasks endpoint

scottlovegrove:
## Summary
- Switches `td completed` from `getCompletedTasksByCompletionDate` (cursor-based) to `getAllCompletedTasks` (offset-based)
- Adds `--label`, `--offset`, `--annotate-notes`, and `--annotate-items` flags
- Uses inline project/section data from the response instead of a separate `getProjects()` call (still fetches full projects only for collaborator resolution when tasks have assignees)

## Breaking Changes
- `--cursor` flag removed — use `--offset` instead
- `--all` flag removed — offset-based pagination risks 429 rate-limit errors with unbounded fetches

## Test plan
- [x] Type-check passes
- [x] All 1129 tests pass (18 in completed suite — 13 updated, 5 new)
- [x] Lint clean
- [x] Format clean
- [ ] Manual: `td completed` shows today's completed tasks
- [ ] Manual: `td completed --label "work"` filters by label
- [ ] Manual: `td completed --offset 10 --limit 5` pagination works
- [ ] Manual: `td completed --annotate-notes --json` includes comment data


scottlovegrove:
Superseded by #XXX — the migration to `getAllCompletedTasks` already landed on main via commit 8de1cb6. Closing as the work here is now redundant.

scottlovegrove:
Reopening — closed this in error. My local branch was stale and I misread the commit graph.

src/commands/completed/list.ts
```diff
@@ -41,34 +46,24 @@ export async function listCompleted(options: CompletedListOptions): Promise<void
 
     const api = await getApi()
 
+    if (isSearch) {
```
doistbot, line 49: <a href="#">P2</a>  `--all` and `--cursor` are silently ignored in the date-based path. Both flags are still defined on the command in `index.ts` and accepted by Commander, but `listByDate` never reads `options.all` or `options.cursor`. A user running `td completed --all` (without `--search`) expects all results, but gets at most the default limit of 30. This is a regression from the old code where `--all` worked correctly via cursor-based pagination. The search path already validates incompatible flags (e.g., `--since` with `--search` throws an error); the date path should follow the same pattern and reject `--all` and `--cursor` when not using `--search`. Alternatively, the flags could be removed from the command definition, but they are still needed for the search path.

src/commands/completed/list.ts
```diff
@@ -174,3 +162,138 @@ export async function listCompleted(options: CompletedListOptions): Promise<void
...
+    if (options.project) {
+        projectId = await resolveProjectId(api, options.project)
+    }
+
+    const limit = options.limit ? parseInt(options.limit, 10) : 30
+    const offset = options.offset ? parseInt(options.offset, 10) : 0
+
+    const resp = await api.getAllCompletedTasks({
+        since: new Date(since + 'T00:00:00'),
+        until: new Date(until + 'T00:00:00'),
```
doistbot, line 180: <a href="#">P2</a>  `until` is set to midnight at the start of the specified date (`new Date(until + 'T00:00:00')`). This creates a range that excludes tasks completed on the `until` date after midnight. For example, `td completed --since 2024-01-08 --until 2024-01-08` sends identical timestamps for both `since` and `until` (Jan 8 00:00:00), producing a zero-width range that returns no tasks. Even for multi-day ranges like `--since 2024-01-01 --until 2024-01-08`, tasks completed on Jan 8 are excluded. The default case works because `until` defaults to tomorrow. Fix: use `T23:59:59` for the `until` timestamp, or advance `until` by one day.

src/commands/completed/list.ts
doistbot, line 49: <a href="#">P2</a>  This early search branch silently ignores the newly accepted `--label`, `--offset`, `--annotate-notes`, and `--annotate-items` options: `listSearchResults` only sends query/cursor/limit. A command such as `td completed --search foo --label urgent` returns unfiltered results. Reject these incompatible combinations (as is already done for date/project options), or implement them for search.

src/commands/completed/list.ts
```diff
@@ -174,3 +162,138 @@ export async function listCompleted(options: CompletedListOptions): Promise<void
...
+        })
+    }
+
+    if (options.json) {
+        const tasksWithAssignee = tasks.map((task) => ({
+            ...task,
+            responsibleName: getAssigneeName(task),
+        }))
+        console.log(
+            formatPaginatedJson(
```
doistbot, line 250: <a href="#">P2</a>  `--annotate-notes --json` does not include the requested annotation data unless the user also adds `--full`. `formatPaginatedJson` strips non-essential task fields by default, and the task essential-field list does not include annotation/comment fields; the same applies to NDJSON and `--annotate-items`. Preserve annotation fields when either annotation flag is set, or make those flags imply full output.

src/commands/completed/list.ts
```diff
@@ -174,3 +162,138 @@ export async function listCompleted(options: CompletedListOptions): Promise<void
...
+        if (!fullProjects || !collaboratorCache) return null
+        return formatAssignee({
+            userId: task.responsibleUid,
+            projectId: task.projectId,
+            projects: fullProjects,
+            cache: collaboratorCache,
+        })
+    }
+
+    if (options.json) {
```
doistbot, line 244: <a href="#">P2</a>  The output blocks in `listByDate` (lines 244–274 for JSON/NDJSON, 276–298 for text) are structurally identical to `listSearchResults` (lines 118–148 for JSON/NDJSON, 150–163 for text). Both functions map tasks to add `responsibleName`, call the same `formatPaginatedJson`/`formatPaginatedNdjson`/`formatTaskRow` functions, and differ only in `nextCursor` (variable vs `null`) and the footer message.

The pre-PR code used a single function with inline `isSearch` branches, which avoided this duplication. Consider extracting a shared helper that accepts the task list, a project-name resolver, an assignee-name resolver, the `nextCursor` value, and a footer string, so that output-format changes only need to be made once.

src/commands/completed/completed.test.ts
```diff
@@ -74,10 +71,10 @@ describe('completed command', () => {
 
         await program.parseAsync(['node', 'td', 'completed'])
 
-        expect(mockApi.getCompletedTasksByCompletionDate).toHaveBeenCalledWith(
+        expect(mockApi.getAllCompletedTasks).toHaveBeenCalledWith(
             expect.objectContaining({
-                since: getToday(),
-                until: getTomorrow(),
+                since: new Date(getToday() + 'T00:00:00'),
```
doistbot, line 76: <a href="#">P2</a>  The default limit changed from `LIMITS.tasks` (300) to a hardcoded `30` in `listByDate`, but no test verifies this default. The `--uses today as default since date` test calls `getAllCompletedTasks` without `--limit` and uses `expect.objectContaining` — it checks `since`/`until` but not `limit`. Since the new offset-hint feature triggers when `tasks.length === limit`, a regression in the default value would silently change hint behavior. Adding `limit: 30` to the existing `expect.objectContaining` in this test would cover it.

doistbot (commented):
This PR migrates the `td completed` command from the cursor-based `getCompletedTasksByCompletionDate` endpoint to the offset-based `getAllCompletedTasks` endpoint, adding `--label`, `--offset`, `--annotate-notes`, and `--annotate-items` flags while removing `--all` and `--cursor` from the date-based path.

Few things worth tightening:

- **Legacy flags silently ignored:** `--all` and `--cursor` are still registered on the command but never read by `listByDate`, so users running `td completed --all` get a truncated default-limit result with no warning. These should either be rejected with an error in the date path (as the search path already does for incompatible flags) or removed from the definition.

- **Date boundary excludes end-date tasks:** `until` is set to midnight (`T00:00:00`) on the specified end date, excluding any tasks completed later that day. Using `T23:59:59` or advancing by one day would fix this.

- **New options silently dropped in search path:** `--label`, `--offset`, `--annotate-notes`, and `--annotate-items` are accepted by Commander but ignored when `--search` is used, returning unfiltered/unannotated results without error.

- **Annotation data stripped in JSON output:** `--annotate-notes --json` (and the same for `--annotate-items` / NDJSON) drops annotation fields unless `--full` is also passed, since `formatPaginatedJson` strips non-essential fields. Annotation flags should preserve their fields or imply full output.

- **Duplicated output blocks:** `listByDate` and `listSearchResults` now have nearly identical JSON/NDJSON/text formatting blocks. Extracting a shared helper would keep output-format changes in one place.

- **Missing default-limit test coverage:** The default limit changed from 300 to a hardcoded `30`, but no test asserts this — adding `limit: 30` to the existing `expect.objectContaining` would guard the offset-hint behavior that depends on it.

- **Minor type-safety and efficiency notes:** The double cast on `resp.projects` could be narrowed to `Record<string, { name: string }>`, and the redundant `getProjects()` call (fired once for name resolution and again for assignee resolution) could be reused.

<details>

- **<a href="#">P3</a> src/commands/completed/list.ts:190**: `resp.projects as Record<string, Record<string, unknown>>` followed by `inlineProjects[pid]?.name as string | undefined` is a double cast that bypasses type safety. If the SDK doesn't type the `projects` field, a narrower cast like `Record<string, { name: string }>` would eliminate the second cast and give compile-time safety on the accessed shape.
- **<a href="#">P3</a> src/commands/completed/list.ts:172**: When `--project` is passed as a project name (not `id:` or URL), `resolveProjectId` → `resolveProjectRef` calls `api.getProjects()` to fuzzy-match the name, then discards the full project list. Later, if any returned tasks have assignees (line 228), `api.getProjects()` is called a second time to build the project map for collaborator resolution. The first call's result could be reused instead of refetching. This only fires in the narrow path of name-based project resolution + assignees, so it's low priority.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F204&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A530030d0-7e9e-11f1-890d-99e5da644e24%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A530030d0-7e9e-11f1-890d-99e5da644e24&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
@scottlovegrove The command doesn't seem to actually work after this PR the command, i.e. returns a bunch of "Invalid input: expected string/object/bool/etc., received undefined" errors.

Should be because `@doist/todoist-sdk`'s `getAllCompletedTasks` runs `validateTaskArray` against the full strict Task schema, but the completed endpoint returns a partial shape (parentId, labels, priority, description, etc. all undefined; addedAt/updatedAt as Invalid Date), so validation throws before anything renders.

I'll let you take a look with no rush.

PR #205, opened 2026-03-31
feat: add read-only OAuth mode for safe autonomous tool use

scottlovegrove:
## Summary

- Adds `td auth login --read-only` flag that authenticates with OAuth scope `data:read`, blocking all mutating API calls at the proxy layer
- Stores auth metadata (mode, scope) in config file alongside the secure token storage introduced in #120
- Creates a permissions module with a default-deny allowlist of ~33 safe API methods — unknown/new methods are blocked by default
- Displays auth mode in `td auth status` (both human-readable and `--json` output)

Supersedes #110, reimplemented against current main to accommodate the secure store refactor (#120) and command directory split (#195).

## Changes

| Area | Files |
|------|-------|
| Auth metadata | `src/lib/auth.ts` — `AuthMode`, `AuthMetadata`, `SaveApiTokenOptions` types; `getAuthMetadata()`; metadata persistence in save/clear |
| OAuth scopes | `src/lib/oauth.ts` — `readOnly` option on `buildAuthorizationUrl` |
| Permissions | `src/lib/permissions.ts` (new) — `KNOWN_SAFE_API_METHODS`, `ensureWriteAllowed()`, `isMutatingApiMethod()`, `isMutatingSyncPayload()` |
| API proxy | `src/lib/api/core.ts` — permission checks before mutating calls |
| Upload guard | `src/lib/api/uploads.ts` — direct `ensureWriteAllowed()` (bypasses proxy) |
| Commands | `src/commands/auth/{index,login,token,status}.ts` — `--read-only` flag, metadata passing, mode display |
| Tests | New: `permissions.test.ts`, `oauth.test.ts`; updated: `auth.test.ts`, `lib-auth.test.ts` |
| Docs | `README.md`, `src/lib/skills/content.ts`, `skills/todoist-cli/SKILL.md` |

## Test plan

- [x] `npm run type-check` passes
- [x] All 1147 tests pass
- [x] `npm run lint:check` — 0 warnings, 0 errors
- [x] `npm run format:check` — all clean
- [x] Manual: `td auth login --read-only` → verify OAuth flow uses `data:read` scope
- [x] Manual: attempt mutating command in read-only mode → verify blocked with clear error
- [x] Manual: `td auth status` → verify mode displayed
- [x] Manual: `td auth login` (without `--read-only`) → verify switches back to read-write

### DIFF
--- a/src/commands/auth/status.ts
+++ b/src/commands/auth/status.ts
@@ -1,8 +1,8 @@
 import chalk from 'chalk'
 import { getApi } from '../../lib/api/core.js'
-import { NoTokenError, getAuthMetadata } from '../../lib/auth.js'
+import { type AuthMode, NoTokenError, getAuthMetadata } from '../../lib/auth.js'
 
-function formatAuthMode(authMode: string, authScope?: string): string {
+function formatAuthMode(authMode: AuthMode, authScope?: string): string {
     if (authMode === 'read-only') {
         return `read-only (OAuth scope ${authScope ?? 'data:read'})`
     }
--- a/src/lib/permissions.ts
+++ b/src/lib/permissions.ts
@@ -55,6 +55,7 @@ const KNOWN_SAFE_API_METHODS = new Set([
 ])
 
 export function isMutatingApiMethod(methodName: string): boolean {
+    if (methodName in Object.prototype) return false
     return !KNOWN_SAFE_API_METHODS.has(methodName)
 }
 
--- a/src/__tests__/permissions.test.ts
+++ b/src/__tests__/permissions.test.ts
@@ -90,6 +90,12 @@ describe('permissions', () => {
         expect(isMutatingApiMethod('brandNewApiMethod')).toBe(true)
     })
 
+    it('exempts Object.prototype methods', () => {
+        expect(isMutatingApiMethod('toString')).toBe(false)
+        expect(isMutatingApiMethod('valueOf')).toBe(false)
+        expect(isMutatingApiMethod('hasOwnProperty')).toBe(false)
+    })
+
     it('detects mutating sync payloads', () => {
         expect(isMutatingSyncPayload([{ commands: [{ type: 'item_add' }] }])).toBe(true)
     })


src/lib/permissions.ts
```diff
@@ -0,0 +1,82 @@
+import { type AuthMetadata, getAuthMetadata } from './auth.js'
+
+export const READ_ONLY_ERROR_MESSAGE =
+    'This CLI is authenticated in read-only mode (OAuth scope data:read). Re-run `td auth login` without --read-only to enable write operations.'
+
+// Default-deny allowlist: any API method NOT listed here is treated as mutating.
+// When adding new read-only SDK methods, add them to this set.
```
scottlovegrove, line 7: We should add this to the developer skill for adding new commands.

scottlovegrove: Done — added as step 3 in the add-command skill (e2ea158). New read-only SDK methods need to be added to `KNOWN_SAFE_API_METHODS`; mutating methods are blocked by default which is the correct behavior.

src/lib/permissions.ts
```diff
@@ -0,0 +1,82 @@
...
+    'exportTemplateAsFile',
+    'exportTemplateAsUrl',
+    // Reminders (REST read)
+    'getReminders',
+    'getLocationReminders',
+    // Sync is handled separately via payload inspection
+    'sync',
+])
+
+export function isMutatingApiMethod(methodName: string): boolean {
```
doistbot, line 57: [P1] Object prototype methods like `toString` or `valueOf` will be incorrectly treated as mutating API methods because they are not in the `KNOWN_SAFE_API_METHODS` allowlist. Since the proxy wraps mutating methods in an `async` function, calling `api.toString()` will return a Promise and throw an error in read-only mode, breaking standard logging and string interpolation.

You can exempt standard prototype methods by checking if they exist on `Object.prototype`:

```typescript
export function isMutatingApiMethod(methodName: string): boolean {
    if (methodName in Object.prototype) return false
    return !KNOWN_SAFE_API_METHODS.has(methodName)
}
```

src/commands/auth/status.ts
```diff
@@ -1,15 +1,34 @@
 import chalk from 'chalk'
 import { getApi } from '../../lib/api/core.js'
-import { NoTokenError } from '../../lib/auth.js'
+import { NoTokenError, getAuthMetadata } from '../../lib/auth.js'
+
+function formatAuthMode(authMode: string, authScope?: string): string {
```
doistbot: [P3] Consider using the explicit `AuthMode` type for the `authMode` parameter instead of `string` to improve type safety.

```typescript
function formatAuthMode(authMode: AuthMode, authScope?: string): string {
```

*(You'll need to export/import `AuthMode` from `../../lib/auth.js` if you make this change)*

src/lib/permissions.ts
scottlovegrove: Good catch — fixed in 212a13e. Added `if (methodName in Object.prototype) return false` guard and a test for toString/valueOf/hasOwnProperty.

src/commands/auth/status.ts
scottlovegrove: Fixed in 212a13e — now uses `AuthMode` type imported from `../../lib/auth.js`.

PR #210, opened 2026-03-31
fix: rename @doist/todoist-api-typescript to @doist/todoist-sdk

scottlovegrove:
## Summary
- Replaces `@doist/todoist-api-typescript@7.8.0` with `@doist/todoist-sdk@8.0.0`
- Updates all 34 source file imports to reference the new package name
- Renames CI workflow file from `update-todoist-api-typescript.yml` to `update-todoist-sdk.yml`
- Handles newly optional `role` field on `SyncWorkspace` type (defaults to `'MEMBER'`)

## Test plan
- [x] All 1148 tests pass
- [x] Build succeeds
- [x] Lint check passes
- [x] Format check passes
- [x] Pre-commit and pre-push hooks pass

### DIFF
--- a/.github/workflows/update-todoist-sdk.yml
+++ b/.github/workflows/update-todoist-sdk.yml
@@ -44,7 +44,7 @@ jobs:
               if: steps.check-updates.outputs.updates-available == 'true'
               run: |
                   echo "Updating @doist/todoist-sdk from ${{ steps.check-updates.outputs.current-version }} to ${{ steps.check-updates.outputs.latest-version }}"
-                  npm install @doist/todoist-sdk@latest
+                  npm install --save-exact @doist/todoist-sdk@latest
 
             - name: Upload updated package files
               if: steps.check-updates.outputs.updates-available == 'true'
--- a/src/lib/api/workspaces.ts
+++ b/src/lib/api/workspaces.ts
@@ -4,7 +4,7 @@ import { getApi } from './core.js'
 export interface Workspace {
     id: string
     name: string
-    role: 'ADMIN' | 'MEMBER' | 'GUEST'
+    role: 'ADMIN' | 'MEMBER' | 'GUEST' | null
     plan: string
     domainName: string | null
     currentMemberCount: number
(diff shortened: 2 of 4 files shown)


src/lib/api/workspaces.ts
```diff
@@ -31,7 +31,7 @@ function toWorkspace(w: SyncWorkspace): Workspace {
     return {
         id: String(w.id),
         name: w.name,
-        role: w.role,
+        role: w.role ?? 'MEMBER',
```
doistbot: [P1] Defaulting a missing workspace role to `MEMBER` fabricates permissions. `SyncWorkspace.role` is optional in the new SDK, so when the API omits it this line will make downstream text/JSON output report `MEMBER` even though no role was provided. Please preserve the missing state in `Workspace` (for example `null`) and handle rendering separately instead of inventing a role here.

.github/workflows/update-todoist-sdk.yml
```diff
@@ -37,14 +37,14 @@ jobs:
...
                   fi
 
-            - name: Update @doist/todoist-api-typescript to latest
+            - name: Update @doist/todoist-sdk to latest
               if: steps.check-updates.outputs.updates-available == 'true'
               run: |
-                  echo "Updating @doist/todoist-api-typescript from ${{ steps.check-updates.outputs.current-version }} to ${{ steps.check-updates.outputs.latest-version }}"
-                  npm install @doist/todoist-api-typescript@latest
+                  echo "Updating @doist/todoist-sdk from ${{ steps.check-updates.outputs.current-version }} to ${{ steps.check-updates.outputs.latest-version }}"
+                  npm install @doist/todoist-sdk@latest
```
doistbot: [P2] This update step no longer preserves the repo's exact dependency pins. `npm install @doist/todoist-sdk@latest` uses npm's default save prefix, so the workflow will typically rewrite `package.json` to `^8.x.y` even though this repo currently checks in exact versions. Adding `--save-exact` here keeps published builds from accepting newer SDK releases that were never tested in CI.

PR #211, opened 2026-04-01
feat: show workspace name after joining a project

scottlovegrove:
## Summary
- `td project join <id>` now displays the workspace name after joining a project
- JSON output (`--json`) includes the full workspace object alongside the join response
- Changed argument from optional `[ref]` to required `<id>`
- Added `getWorkspace` SDK method support (spinner, permissions, mock)

## Test plan
- [x] All 1148 tests pass
- [x] Type-check passes
- [x] Lint and format checks pass
- [ ] Manual test: `td project join id:<workspace-project-id>` shows project name and workspace name
- [ ] Manual test: `td project join id:<workspace-project-id> --json` includes workspace in output
- [ ] Manual test: `td project join id:<id> --dry-run` previews without calling API

### DIFF
--- a/src/commands/project/join.ts
+++ b/src/commands/project/join.ts
@@ -24,13 +24,17 @@ export async function joinProjectCmd(
     }
     const response = await api.joinProject(id)
     const { project } = response
-    const workspace = await api.getWorkspace(project.workspaceId)
+    const workspace = project.workspaceId ? await api.getWorkspace(project.workspaceId) : null
 
     if (options.json) {
-        console.log(JSON.stringify({ ...response, workspace }, null, 2))
+        console.log(JSON.stringify(workspace ? { project, workspace } : { project }, null, 2))
         return
     }
 
     console.log(`Joined: ${project.name}`)
-    console.log(chalk.dim(`Workspace: ${workspace.name}`))
+    if (workspace) {
+        console.log(chalk.dim(`Workspace: ${workspace.name}`))
+    } else {
+        console.log(chalk.dim(`ID: ${project.id}`))
+    }
 }
--- a/src/__tests__/project.test.ts
+++ b/src/__tests__/project.test.ts
@@ -1847,6 +1847,34 @@ describe('project join', () => {
             project: { id: 'proj-123', name: 'Shared Project' },
             workspace: { id: 'ws-1', name: 'Acme Corp' },
         })
+        expect(parsed.tasks).toBeUndefined()
+    })
+
+    it('handles personal shared project without workspace', async () => {
+        const program = createProgram()
+
+        mockApi.joinProject.mockResolvedValue({
+            project: {
+                id: 'proj-456',
+                name: 'Personal Shared',
+                color: 'blue',
+                isFavorite: false,
+            },
+            tasks: [],
+            sections: [],
+            comments: [],
+            collaborators: [],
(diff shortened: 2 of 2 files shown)


src/commands/project/join.ts
```diff
@@ -15,13 +15,15 @@ export async function joinProjectCmd(
     }
 
     const api = await getApi()
-    const project = await api.joinProject(id)
+    const response = await api.joinProject(id)
+    const { project } = response
+    const workspace = await api.getWorkspace(project.workspaceId)
```
doistbot: [P1] Crash when joining a personal shared project

Shared projects are not always part of a workspace (e.g., personal shared projects). When `project.workspaceId` is missing (`null` or `undefined`), passing it to `api.getWorkspace` will cause an invalid API call or crash. We should verify `workspaceId` exists before fetching it.

```suggestion
    const workspace = project.workspaceId ? await api.getWorkspace(project.workspaceId) : null
```

src/commands/project/join.ts
```diff
@@ -15,13 +15,15 @@ export async function joinProjectCmd(
...
 
     if (options.json) {
-        console.log(formatJson(project, 'project'))
+        console.log(JSON.stringify({ ...response, workspace }, null, 2))
```
doistbot: [P2] Excessive payload in JSON output

The updated `joinProject` SDK method returns a full sync payload, including `tasks`, `sections`, `comments`, and more. Spreading `...response` dumps all of this into the JSON output, massively inflating the size and introducing unintended noise compared to the previous behavior.

We should limit the output to just the `project` (and `workspace` if available) to keep the JSON output focused.

```suggestion
        console.log(JSON.stringify(workspace ? { project, workspace } : { project }, null, 2))
```

src/commands/project/join.ts
```diff
@@ -15,13 +15,15 @@ export async function joinProjectCmd(
...
         return
     }
 
     console.log(`Joined: ${project.name}`)
-    console.log(chalk.dim(`ID: ${project.id}`))
+    console.log(chalk.dim(`Workspace: ${workspace.name}`))
```
doistbot: [P1] Crash when accessing `workspace.name` on null

If the joined project does not belong to a workspace, `workspace` will be null or undefined. Accessing `workspace.name` will throw a runtime error. We should gracefully fall back to displaying the project ID like the previous behavior.

```suggestion
    if (workspace) {
        console.log(chalk.dim(`Workspace: ${workspace.name}`))
    } else {
        console.log(chalk.dim(`ID: ${project.id}`))
    }
```

src/commands/project/join.ts
scottlovegrove: Fixed — now guards with `project.workspaceId ? await api.getWorkspace(project.workspaceId) : null` so personal shared projects won't crash.

src/commands/project/join.ts
scottlovegrove: Good catch — JSON output now only includes `{ project, workspace }` (or just `{ project }` when no workspace). No more tasks/sections/comments noise.

src/commands/project/join.ts
scottlovegrove: Fixed — falls back to showing `ID: project.id` when workspace is null.

PR #212, opened 2026-04-01
fix: suppress credential manager warning on every command

scottlovegrove:
## Summary
- Removes the "system credential manager unavailable; using plaintext token from ..." warning that printed to stderr on **every** command invocation when the secure store isn't available
- Users are still warned once at token save time (`td auth login` / `td auth token`), so they know their token is stored in plaintext — they just won't be nagged about it on every run
- Removes the now-unused `warnSecureStoreFallback` helper

## Test plan
- [x] All existing auth tests pass (37 auth tests, 1150 total)
- [ ] Verify `td inbox` no longer prints the warning when using config-file fallback
- [ ] Verify `td auth login` / `td auth token` still warns about plaintext storage at save time

### DIFF
--- a/src/__tests__/lib-auth.test.ts
+++ b/src/__tests__/lib-auth.test.ts
@@ -295,6 +295,7 @@ describe('lib/auth', () => {
         expect(readConfig()).toEqual({ api_token: 'fallback-token-123456' })
 
         await expect(getApiToken()).resolves.toBe('fallback-token-123456')
+        expect(errorSpy).not.toHaveBeenCalled()
 
         setConfig({
             api_token: 'fallback-token-123456',


src/__tests__/lib-auth.test.ts
```diff
@@ -295,9 +295,6 @@ describe('lib/auth', () => {
         expect(readConfig()).toEqual({ api_token: 'fallback-token-123456' })
 
         await expect(getApiToken()).resolves.toBe('fallback-token-123456')
-        expect(errorSpy).toHaveBeenCalledWith(
```
doistbot, line 298: [P2] Keep an explicit assertion that `getApiToken()` stays silent here. Removing this check means the test no longer verifies the regression this PR is fixing, so a future change could reintroduce stderr output on normal command runs without any failure.

Please replace it with `expect(errorSpy).not.toHaveBeenCalled()` after `await expect(getApiToken()).resolves.toBe(...)` so the test still guards the new behavior.

scottlovegrove: Good catch — added `expect(errorSpy).not.toHaveBeenCalled()` after the `getApiToken()` call so the test explicitly guards the new silent behavior.

PR #213, opened 2026-04-02
ci: add next branch for pre-release support

scottlovegrove:
## Summary
- Adds `next` branch support to semantic-release config and release workflow
- Merges into `next` will publish pre-release versions to npm under the `next` dist-tag
- Changelog and version commit-back are skipped for pre-releases
- Twist release announcements only fire for stable releases on `main`
- Adds CONTRIBUTING.md with pre-release workflow documentation

## Post-merge setup
After merging, create the `next` branch:
```bash
git checkout -b next main && git push -u origin next
```

## Test plan
- [ ] Verify existing `main` branch releases still work (including Twist announcement)
- [ ] Create `next` branch and merge a `feat:` commit — verify pre-release published with `next` dist-tag
- [ ] Verify Twist announcement does NOT fire for pre-releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,7 @@ permissions:
 # Release involves crucial steps that shouldn't be cancelled mid-run,
 # so new workflow runs are queued until the previous one finishes.
 concurrency:
-    group: ${{ github.workflow }}
+    group: ${{ github.workflow }}-${{ github.ref }}
     cancel-in-progress: false
 
 jobs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,7 +57,7 @@ Each commit message consists of a **header**, an **optional body**, and an **opt
 [optional footer(s)]
 ```
 
-Commit types such as `feat:` and `fix:` affect versioning and changelog generation. Other types like `build:`, `chore:`, `ci:`, `docs:`, `perf:`, `refactor:`, `revert:`, `style:` and `test:` are valid but will not trigger a release.
+Commit types such as `feat:` and `fix:` affect versioning and changelog generation. Other types like `build:`, `chore:`, `ci:`, `docs:`, `perf:`, `refactor:`, `style:` and `test:` are valid but will not trigger a release.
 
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -68,6 +69,7 @@ jobs:
               run: npm ci
 
             - name: Capture previous tag
+              if: github.ref_name == 'main'
```
doistbot, line 72: [P1] Now that `next` can create prerelease tags, this step can capture `vX.Y.Z-next.N` as `PREVIOUS_TAG` after `next` is merged back into `main`, because `git describe --tags --abbrev=0` just returns the closest reachable tag. The later announcement diff then becomes `vX.Y.Z-next.N..vX.Y.Z`, which usually drops the actual release commits from the Twist message. Please restrict this lookup to stable tags only (or explicitly exclude `-next.`) before deriving the announcement range.

.github/workflows/release.yml
```diff
@@ -4,6 +4,7 @@ on:
     push:
         branches:
             - main
+            - next
```
doistbot, line 7: [P2] Adding `next` here means prerelease and stable publishes now share the same workflow-level concurrency key (`${{ github.workflow }}`). A long-running prerelease on `next` will queue a `main` release behind it even though the two release lanes are independent. Making the concurrency group branch-specific would avoid unnecessary release delays.

CONTRIBUTING.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+<type>[optional scope]: <description>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+Commit types such as `feat:` and `fix:` affect versioning and changelog generation. Other types like `build:`, `chore:`, `ci:`, `docs:`, `perf:`, `refactor:`, `revert:`, `style:` and `test:` are valid but will not trigger a release.
```
doistbot: [P2] `revert:` is not in the non-releasing bucket with the current `@semantic-release/commit-analyzer` setup. Under the conventional commits preset it triggers a patch release, so this sentence currently documents the opposite behavior and can lead to accidental publishes. Please remove `revert:` from this list or call out that it still releases.

PR #215, opened 2026-04-02
feat: allow switching between stable and pre-release update channels

scottlovegrove:
## Summary
- Adds `td update switch --stable` and `td update switch --pre-release` commands so users can opt into pre-release (next) updates from within the CLI
- `td update` now reads the configured channel from `~/.config/todoist-cli/config.json` and fetches from the appropriate npm dist-tag (`latest` vs `next`)
- Extracts generic config read/write into `src/lib/config.ts`, keeping auth-specific logic in `auth.ts`
- Converts the `update` command from a flat file to a folder-based command group
- Skips the `td changelog` hint after pre-release updates (since pre-releases don't update the changelog)

## Test plan
- [x] 18 update tests pass (existing + new channel/switch tests)
- [x] Full suite: 1159 tests pass
- [x] Type-check, lint, skill-sync all clean
- [x] Manual: `td update switch --pre-release` writes config and shows warning
- [x] Manual: `td update switch --stable` writes config
- [x] Manual: `td update --check` respects configured channel

### DIFF
--- a/src/commands/update/action.ts
+++ b/src/commands/update/action.ts
@@ -14,6 +14,38 @@ function getInstallTag(channel: UpdateChannel): string {
     return channel === 'pre-release' ? 'next' : 'latest'
 }
 
+interface ParsedVersion {
+    major: number
+    minor: number
+    patch: number
+    prerelease: string | undefined
+}
+
+function parseVersion(version: string): ParsedVersion {
+    const [core, ...rest] = version.split('-')
+    const [major, minor, patch] = core.split('.').map(Number)
+    return { major, minor, patch, prerelease: rest.length > 0 ? rest.join('-') : undefined }
+}
+
+/** Returns true when `candidate` is strictly newer than `current` per semver. */
+function isNewer(current: string, candidate: string): boolean {
+    const a = parseVersion(current)
+    const b = parseVersion(candidate)
+
+    // Compare major.minor.patch
+    for (const key of ['major', 'minor', 'patch'] as const) {
+        if (b[key] !== a[key]) return b[key] > a[key]
+    }
+
+    // Equal core: release > pre-release
+    if (!a.prerelease && b.prerelease) return false
+    if (a.prerelease && !b.prerelease) return true
+
+    // Both pre-release: lexicographic (handles "next.1" vs "next.2" etc.)
+    if (a.prerelease && b.prerelease) return b.prerelease > a.prerelease
+    return false
+}
+
 async function fetchVersion(channel: UpdateChannel): Promise<string> {
     const url = `https://registry.npmjs.org/${PACKAGE_NAME}/${getInstallTag(channel)}`
     const response = await fetch(url)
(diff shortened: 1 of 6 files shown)


src/lib/config.ts
```diff
@@ -0,0 +1,49 @@
...
+import { dirname, join } from 'node:path'
+
+export const CONFIG_PATH = join(homedir(), '.config', 'todoist-cli', 'config.json')
+
+export type UpdateChannel = 'stable' | 'pre-release'
+
+export interface Config extends Record<string, unknown> {
+    api_token?: string
+    pendingSecureStoreClear?: boolean
+    auth_mode?: string
```
doistbot: Moved `AuthMode` to `config.ts` so it owns the type directly — no circular dependency.

src/commands/update/action.ts
```diff
@@ -43,14 +47,23 @@ function runInstall(pm: string): Promise<{ exitCode: number; stderr: string }> {
...
 }
 
+function channelLabel(channel: UpdateChannel): string {
+    return channel === 'pre-release' ? ` ${chalk.magenta('(pre-release)')}` : ''
+}
+
 export async function updateAction(options: { check?: boolean }): Promise<void> {
+    const config = await readConfig()
+    const channel: UpdateChannel = config.update_channel ?? 'stable'
+    const tag = getInstallTag(channel)
```
doistbot, line 89: [P1] This new channel-aware path can downgrade users. `next` legitimately lags behind `latest` between pre-release cycles, so a user on stable `1.36.0` with `update_channel: 'pre-release'` will fetch `1.36.0-next.1`, fail the equality check, and be prompted to "update" to an older build. Please compare semver precedence before printing/installing and skip the install when the channel tag resolves to a version older than the currently installed one.

src/commands/update/index.ts
```diff
@@ -0,0 +1,18 @@
+import { Command } from 'commander'
+import { updateAction } from './action.js'
+import { switchChannel } from './switch.js'
+
+export function registerUpdateCommand(program: Command): void {
+    const update = program
+        .command('update')
+        .description('Update the CLI to the latest version')
```
doistbot: [P3] The help text is now misleading. After this change, `td update` follows the configured release channel and may install `@next`, so "latest version" is no longer always accurate. Rewording this to mention the configured channel would keep `--help` aligned with the new behavior.

src/commands/update/action.ts
scottlovegrove: Added semver-aware comparison via `isNewer()`. When the channel version is older than current (e.g. stable 1.36.0 → pre-release 1.35.1-next.1), we show a note ("v1.35.1-next.1 (pre-release) is older than your current v1.36.0") but still proceed with the install. For `--check`, it correctly shows "Already up to date" without the downgrade note.

src/commands/update/index.ts
scottlovegrove: Updated to "Update the CLI to the latest version for the configured channel".

skills/todoist-cli/SKILL.md
```diff
@@ -446,7 +446,10 @@ td view <url> --limit 25 --ndjson              # Passthrough list options where
 ### Update
 ```bash
 td update                                    # Update CLI to latest version
-td update --check                            # Check for updates without installing
+td update --check                            # Check for updates without installing, show channel
+td update channel                            # Show current update channel
```
frankieyan, line 450: Minor nit: this command sounds like it's updating the channel, rather than being a show command, and using a flag might seem more appropriate. Not sure if it is worth changing, though.

src/commands/update/action.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+function detectPackageManager(): string {
+    const execPath = process.env.npm_execpath ?? ''
+    if (execPath.includes('pnpm')) return 'pnpm'
+    return 'npm'
+}
+
+function runInstall(pm: string, tag: string): Promise<{ exitCode: number; stderr: string }> {
+    const command = pm === 'pnpm' ? 'add' : 'install'
```
frankieyan, line 66: No love for yarn and bun? 😁

frankieyan (commented):
✅

PR #218, opened 2026-04-02
feat: td goal CLI command

domjost:
## Summary
- Add `td goal` command with 9 subcommands: list, view (default), create, update, delete, complete, uncomplete, link, unlink
- Add `resolveGoalRef()` for name/ID/URL resolution
- Add `GOAL_ESSENTIAL_FIELDS` for JSON output
- Add goal fixtures, mock API methods, and 12 tests
- Update skill content with goal command reference

Part of bringing the `Goals` feature to the REST API, CLI, and MCP. See [spec](https://docs.google.com/document/d/1hxC_PW7nbEvwtUpcyMf-3Oa1spL5GvxugLPGts9E5O4/edit?tab=t.qpdgp7ny8amt).

Depends on SDK PR (Doist/todoist-sdk-typescript#548).

| PR | Repo | URL | Base |
  |----|------|-----|------|
  | 1.1 Pagination | Todoist | [#27272](https://github.com/Doist/Todoist/pull/27272) | `main` |
  | 1.2 CRUD + lifecycle | Todoist | [#27273](https://github.com/Doist/Todoist/pull/27273) | PR 1.1 |
  | 1.3 Search | Todoist | [#27274](https://github.com/Doist/Todoist/pull/27274) | PR 1.2 |
  | 1.4 Item linking | Todoist | [#27275](https://github.com/Doist/Todoist/pull/27275) | PR 1.3 |
  | 1.5 goal_id on tasks | Todoist | [#27276](https://github.com/Doist/Todoist/pull/27276) | PR 1.4 |
  | 2 SDK types + methods | todoist-sdk-typescript | [#548](https://github.com/Doist/todoist-sdk-typescript/pull/548) | `main` |
  | 3 CLI `td goal` | todoist-cli | [#218](https://github.com/Doist/todoist-cli/pull/218) | `main` |
  | 4 MCP tools | todoist-ai | [#433](https://github.com/Doist/todoist-ai/pull/433) | `main` |

  **Merge order:** 1.1 → 1.2 → 1.3 → 1.4 → 1.5 → 2 → (3 + 4 in parallel)

## Test plan
- [ ] `npm test` — all 1162 tests pass
- [ ] `npm run type-check` — clean
- [ ] `npm run check:skill-sync` — in sync


src/commands/goal.ts
```diff
@@ -0,0 +1,414 @@
...
+async function listGoals(options: PaginatedViewOptions & { workspace?: string }): Promise<void> {
+    const api = await getApi()
+
+    const targetLimit = options.all
+        ? Number.MAX_SAFE_INTEGER
+        : options.limit
+          ? parseInt(options.limit, 10)
+          : LIMITS.goals
+
+    const ownerType = options.workspace ? 'WORKSPACE' : undefined
```
doistbot: [P1] The `--workspace <ref>` flag takes a specific workspace reference, but here it is merely evaluated as a boolean to set `ownerType = 'WORKSPACE'`. As a result, `td goal list --workspace "Work"` will incorrectly return goals from *all* workspaces rather than filtering to the requested one.

You should resolve the specific workspace using `resolveWorkspaceRef(options.workspace)` and either pass `workspaceId` to `api.getGoals` (if supported by the SDK) or filter the returned goals client-side.

src/lib/output.ts
```diff
@@ -217,6 +217,14 @@ const LOCATION_REMINDER_ESSENTIAL_FIELDS = [
     'locTrigger',
     'radius',
 ] as const
+const GOAL_ESSENTIAL_FIELDS = [
+    'id',
+    'name',
+    'ownerType',
```
doistbot, line 221: [P2] Consider including `'ownerId'` alongside `'ownerType'` in `GOAL_ESSENTIAL_FIELDS`. If a goal belongs to a workspace (`ownerType === 'WORKSPACE'`), the `ownerId` contains the actual workspace ID. Without it, the JSON output won't indicate which workspace the goal is attached to.

src/lib/refs.ts
```diff
@@ -303,6 +303,16 @@ export async function resolveProjectId(api: TodoistApi, ref: string): Promise<st
     return project.id
 }
 
+export async function resolveGoalRef(api: TodoistApi, ref: string): Promise<Goal> {
+    return resolveRef(
+        ref,
+        (id) => api.getGoal(id),
+        () => api.getGoals(),
```
doistbot: [P1] `api.getGoals()` returns a paginated list of goals (only the first page). If the goal the user is trying to resolve by name is not on the first page, `resolveRef` will fail to find it.

Use `paginate` to fetch all goals, similar to how it is done for tasks:
```suggestion
        () =>
            paginate((cursor, limit) => api.getGoals({ cursor: cursor ?? undefined, limit }), {
                limit: Number.MAX_SAFE_INTEGER,
            }),
```

src/commands/goal.ts
```diff
@@ -0,0 +1,414 @@
...
+        : options.limit
+          ? parseInt(options.limit, 10)
+          : LIMITS.tasks
+
+    const { results: tasks, nextCursor } = await paginate(
+        (cursor, limit) => api.getTasks({ goalId: goal.id, cursor: cursor ?? undefined, limit }),
+        { limit: targetLimit },
+    )
+
+    if (options.json) {
```
doistbot, line 99: [P2] `goal view --json`/`--ndjson` drops the goal itself and only serializes the linked tasks. That makes the machine-readable output inconsistent with the human view and with other entity-view commands, and there is no way to retrieve the goal fields from this command. Consider emitting the goal alongside the task list (for example `{ goal, tasks, nextCursor }`) instead of only `results: tasks`.

src/commands/goal.ts
```diff
@@ -0,0 +1,414 @@
...
+    const api = await getApi()
+    const goal = await resolveGoalRef(api, ref)
+
+    const args: Record<string, string | null | undefined> = {}
+    if (options.name !== undefined) args.name = options.name
+    if (options.description !== undefined) args.description = options.description
+    if (options.deadline !== undefined) args.deadline = options.deadline
+    if (options.responsible !== undefined) args.responsibleUid = options.responsible
+
+    const updated = await api.updateGoal(goal.id, args)
```
doistbot, line 256: [P2] This command never checks whether any update fields were supplied before calling `updateGoal`. Running `td goal update <ref>` with no flags will send an empty body and either produce a misleading `Updated:` message or bubble up a server-side validation error. The other update commands guard this locally with a `NO_CHANGES` error, so goal should do the same here.

doistbot (commented):
This PR successfully introduces the `td goal` command alongside its subcommands, reference resolution, and test coverage, significantly expanding the CLI's capabilities. The implementation is well-structured and aligns closely with existing CLI patterns to provide a seamless experience for the new SDK features. There are just a few logic and consistency details to refine, specifically regarding workspace flag evaluation during listing, pagination in reference resolution, empty update handling, and ensuring JSON outputs include both complete goal objects and their associated `ownerId` fields.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F218&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3ecb2d10-2ebf-11f1-8d3a-8e2d05b3df7b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3ecb2d10-2ebf-11f1-8d3a-8e2d05b3df7b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/__tests__/helpers/fixtures.ts
```diff
@@ -269,6 +269,52 @@ export const fixtures = {
             itemOrder: 1,
         } as Filter,
     },
+    goals: {
+        shipV2: {
+            id: '550e8400-e29b-41d4-a716-446655440000',
+            ownerType: 'USER',
```
scottlovegrove, line 275: Just for completeness, we should probably have this fixture as a workspace owner, right? Rather than having both of these as being user owned.

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    let { results: goals, nextCursor } = await paginate(
+        (cursor, limit) => api.getGoals({ ownerType, cursor: cursor ?? undefined, limit }),
+        { limit: targetLimit },
+    )
+
+    // If a specific workspace was requested, filter to just that workspace's goals
+    if (workspaceId) {
+        goals = goals.filter((g) => g.ownerId === workspaceId)
+    }
```
scottlovegrove: I wonder if this should be added to the backend so that _it_ can do the filtering for us. So when you make a call to `api.getGoals()` you include the workspaceId as well. @chiara-doist would that be doable?

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    if (goals.length === 0) {
+        console.log('No goals found.')
+        return
+    }
+
+    for (const goal of goals) {
+        const id = chalk.dim(goal.id.slice(0, 8))
+        const name = goal.isCompleted ? chalk.strikethrough(goal.name) : goal.name
+        const progress = formatProgressBar(goal.progress.percentage, 10)
```
scottlovegrove: We need to factor in accessibility here. We have precendent in this CLI for this when it comes to a progress marker (I think in the project health).

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+            }),
+        )
+        console.log(formatPaginatedNdjson({ results: tasks, nextCursor }, 'task', options.full))
+        return
+    }
+
+    // Header
+    console.log(chalk.bold(goal.name))
+    console.log(chalk.dim(`ID:       ${goal.id}`))
+    console.log(chalk.dim(`Owner:    ${goal.ownerType}`))
```
scottlovegrove: Given that ownerType can be shouty (`USER` or `WORKSPACE`), I wonder if we should have a formatting helper that can print a nicer `User` or `Workspace`. Probably only needs to be a ternary check.

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+    if (goal.description) console.log(`Desc:     ${goal.description}`)
+    if (goal.deadline) console.log(`Deadline: ${chalk.green(goal.deadline)}`)
+    console.log(
+        `Progress: ${formatProgressBar(goal.progress.percentage)} (${goal.progress.completedItemCount}/${goal.progress.totalItemCount})`,
+    )
```
scottlovegrove, line 127: Again, we need to factor in accessibility here for the progress bar.

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+    if (goal.isCompleted) console.log(chalk.green('✓ Completed'))
```
scottlovegrove: Accessibility here might be to not have the tick, but to have `is` instead. So if accessibility is on, it would display "Is completed". Or maybe better it should say "Goal completed".

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+    dryRun?: boolean
+}
+
+async function updateGoal(ref: string, options: UpdateOptions): Promise<void> {
+    if (
+        options.name === undefined &&
+        options.description === undefined &&
+        options.deadline === undefined &&
+        options.responsible === undefined
+    ) {
```
scottlovegrove: Might be better to spread the `options`, then just do `if (!name && !description && !deadline && !responsible)`

The rest of the function could then use the resulting spread items rather than `options.name` etc.

src/commands/goal.ts
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    const api = await getApi()
+    const goal = await resolveGoalRef(api, ref)
+
+    const args: Record<string, string | null | undefined> = {}
+    if (options.name !== undefined) args.name = options.name
+    if (options.description !== undefined) args.description = options.description
+    if (options.deadline !== undefined) args.deadline = options.deadline
+    if (options.responsible !== undefined) args.responsibleUid = options.responsible
```
scottlovegrove: These could all then be `if (!name) ....` etc.

scottlovegrove (changes requested):

src/commands/goal.ts
chiara-doist: Absolutely it should yes! This should happen at API / DB level

src/__tests__/helpers/fixtures.ts
domjost: Fixed — `learnRust` fixture is now `ownerType: 'WORKSPACE', ownerId: 'ws-1'`.

src/commands/goal.ts
domjost: Agreed this should be backend-side. For now keeping the client-side filter since the API doesn't support `workspace_id` on `GET /goals` yet. Chiara confirmed she'll add it — we can remove the client-side filter once that's in.

src/commands/goal.ts
domjost: Fixed — `formatProgressBar()` already returns plain text (`40%`) in accessible mode, so the list and view progress bars are covered.

src/commands/goal.ts
domjost: Fixed — added `formatOwnerType()` helper that returns `User`/`Workspace` instead of the shouty enum values.

src/commands/goal.ts
domjost: Fixed — now shows `Goal completed` in accessible mode instead of the `✓` tick.

src/commands/goal.ts
domjost: Fixed — destructured options at the top of the function. The check is now `if (!name && !description && !deadline && !responsible)` and the rest uses the spread vars.

src/commands/goal.ts
scottlovegrove: I think we can remove this already. We won't be merging any of this until the backend is ready, which includes the SDK, so might as well just remove it now in preparation as we know the workspace filtering will be done on the backend.

domjost: Removed the client-side workspace filtering and ref resolution. The `--workspace` flag now just sets `ownerType: 'WORKSPACE'`. The actual `workspaceId` filtering will be added when the backend supports it (before merge).

src/commands/goal.ts
domjost: `formatProgressBar()` already returns plain text (`40%`) when accessibility is on — no visual bar. The `(2/5)` count is plain text so it's also accessible. Let me know if you want something different here.

src/commands/goal.ts
domjost: Fixed — now uses `if (name)` instead of `if (name !== undefined)` since we already destructured.

src/commands/goal.ts
scottlovegrove: Remember to add this to the SDK on the `getGoals()` function.

scottlovegrove (approved):
Approving, but we still need to add the `workspaceId` to the `getGoals()` call. We can do that now even though the SDK isn't ready for it, because there is no SDK update for this to validate against just yet 😄

src/commands/goal.ts
chiara-doist: Folks, I'm unresolving this, otherwise pretty sure we'll forget.  

We are still in time to add it: https://github.com/Doist/Todoist/pull/27273

skills/todoist-cli/SKILL.md
```diff
@@ -324,6 +328,25 @@ td filter create --name "Urgent work" --query "p1 & #Work" --dry-run  # Preview
...
+td goal create --name "Ship v2" --json       # Return created goal as JSON
+td goal create --name "Ship v2" --dry-run    # Preview creation
+td goal update "Ship v2" --name "Ship v3"
+td goal update "Ship v2" --description "New desc" --json
+td goal delete "Ship v2" --yes
+td goal complete "Ship v2"                   # Mark goal as completed
+td goal uncomplete "Ship v2"                 # Reopen a completed goal
+td goal link "Ship v2" --task "Buy milk"     # Link a task to a goal
+td goal unlink "Ship v2" --task "Buy milk"   # Unlink a task from a goal
+```
```
gnapse, line 163: It would be good if the interface to link tasks to a goal would support passing multiple tasks at once. Ask the coding agent that you're using what would be a good way to support that in the command line interface.

And maybe this is something we should support at the SDK level as well.

domjost: Good idea! Here are the options for supporting multiple tasks:

**Option 1: Repeated flag (recommended)**
```bash
td goal link "Ship v2" --task "Buy milk" --task "Write tests"
```
Commander supports this natively with variadic options. Idiomatic and consistent with how `--labels` works on `td task add`.

**Option 2: Comma-separated**
```bash
td goal link "Ship v2" --tasks "id:abc,id:def"
```
Simple but breaks if task names contain commas. Works well with IDs only.

**Option 3: Positional args**
```bash
td goal link "Ship v2" "Buy milk" "Write tests"
```
Clean but ambiguous — hard to tell where the goal ref ends and task refs begin.

At the SDK level, `linkTaskToGoal` currently takes a single `taskId`. The CLI can loop over multiple tasks client-side for now. A batch SDK method or backend endpoint could be added later if needed.

What's your preference?

scottlovegrove: > And maybe this is something we should support at the SDK level as well.

We should, but it would require backend updates as well in order for the SDK to not just be making multiple calls. 

> What's your preference?

Option 1 I think would be my preference on this.

gnapse: yeah, option 1, more explicit, no big deal since it will be the agents generating the command lines they want to execute.

domjost: Option 1 it is. Should I also tackle support at the SDK at backend level?

chiara-doist: Seeing this; is this also creating new tasks? 

```
td goal link "Ship v2" --task "Buy milk" --task "Write tests"
```

scottlovegrove: > Seeing this; is this also creating new tasks?

No, it wouldn't be. But that does raise a question of whether we are supporting adding a goalId when creating a task. If we're not, should we? @domjost question for the backend work and the SDK changes.

gnapse: > Seeing this; is this also creating new tasks?

Another question that it raises is: what happens if some of the task references are not resolved to actual tasks, but some are. Do we fail without linking any task? (that would be my preference)

domjost: > Scott: whether we are supporting adding a goalId when creating a task. If we're not, should we?

We should. Adding a task to revisit this later.

> Ernesto: what happens if some of the task references are not resolved to actual tasks, but some are?

I'd prefer linking what was resolved, and returning a list of taskId's that weren't able to be linked. This would give the user/agent enough information for the next steps.

Thoughts?

src/commands/goal.ts
chiara-doist: Confirmed: this was added, closing!

skills/todoist-cli/SKILL.md
scottlovegrove: > I'd prefer linking what was resolved, and returning a list of taskId's that weren't able to be linked. This would give the user/agent enough information for the next steps.

Sounds good to me 👍🏻

gnapse: Yes, sounds good.

scottlovegrove (approved):

domjost:
@scottlovegrove `workspaceId` is being passed to `getGoals()` — and with the SDK now at `8.4.0-next.2`, it's properly typed (removed the type assertion). All checks pass.

skills/todoist-cli/SKILL.md
domjost: @scottlovegrove @gnapse I'll tackle this extension in a separate PR. does that make sense?

scottlovegrove: That's fine 👍🏻

PR #220, opened 2026-04-03
refactor: change update channel from subcommand to flag

scottlovegrove:
## Summary
- Changes `td update channel` subcommand to `td update --channel` flag, addressing [review feedback](https://github.com/Doist/todoist-cli/pull/215#discussion_r3031392657) that the subcommand sounded like it was updating the channel rather than showing it
- Inlines the show-channel logic into `updateAction`, removes `channel.ts`
- Updates tests and skill docs

## Test plan
- [x] 25 update tests pass (including 3 new `--channel` flag tests)
- [x] Full suite: 1166 tests pass
- [x] Type-check, lint clean

### DIFF
--- a/src/commands/update/action.ts
+++ b/src/commands/update/action.ts
@@ -85,6 +85,12 @@ function channelLabel(channel: UpdateChannel): string {
 }
 
 export async function updateAction(options: { check?: boolean; channel?: boolean }): Promise<void> {
+    if (options.check && options.channel) {
+        console.error(chalk.red('Error:'), 'Specify either --check or --channel, not both.')
+        process.exitCode = 1
+        return
+    }
+
     const config = await readConfig()
     const channel: UpdateChannel = config.update_channel ?? 'stable'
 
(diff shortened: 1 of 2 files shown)


src/commands/update/action.ts
```diff
@@ -84,9 +84,19 @@ function channelLabel(channel: UpdateChannel): string {
     return channel === 'pre-release' ? ` ${chalk.magenta('(pre-release)')}` : ''
 }
 
-export async function updateAction(options: { check?: boolean }): Promise<void> {
+export async function updateAction(options: { check?: boolean; channel?: boolean }): Promise<void> {
     const config = await readConfig()
     const channel: UpdateChannel = config.update_channel ?? 'stable'
+
+    if (options.channel) {
```
doistbot, line 97: [P2] `--channel` can now be combined with `--check`, and this early return makes `td update --check --channel` silently skip the registry lookup and behave as “show channel only”. Before this refactor those modes were separate commands, so this ambiguous combination was impossible.

Please make the two flags conflict or emit an explicit error when both are set, so users do not get a different action than `--check` promises.

scottlovegrove: Fixed — `td update --check --channel` now errors with "Specify either --check or --channel, not both." Added a test for this case too.

PR #221, opened 2026-04-03
feat: add --json flag to project collaborators, comment view, and workspace view

scottlovegrove:
## Summary
- Add `--json`, `--ndjson`, and `--full` flags to `td project collaborators`
- Add `--json` and `--full` flags to `td comment view`
- Add `--json` and `--full` flags to `td workspace view`
- Update SKILL_CONTENT with new flag examples

These three data-output commands were the only list/view commands missing JSON support.

## Test plan
- [x] All 1174 existing tests pass
- [x] 7 new tests added (3 for project collaborators, 2 for comment view, 2 for workspace view)
- [x] Type-check passes
- [x] Skill sync check passes
- [ ] Manual: `td project collaborators <ref> --json`
- [ ] Manual: `td comment view id:<id> --json`
- [ ] Manual: `td workspace view <name> --json`

### DIFF
--- a/src/commands/comment/view.ts
+++ b/src/commands/comment/view.ts
@@ -12,7 +12,7 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
     const comment = await api.getComment(id)
 
     if (options.json) {
-        console.log(formatJson(comment, 'comment', options.full))
+        console.log(formatJson(comment, 'comment', options.full, true))
         return
     }
 
--- a/src/lib/skills/content.ts
+++ b/src/lib/skills/content.ts
@@ -214,6 +214,7 @@ td project list --personal                    # Personal projects only
 td project view "Project Name"
 td project collaborators "Project Name"
 td project collaborators "Project Name" --json   # JSON output
+td project collaborators "Project Name" --json --full  # JSON with all fields
 td project collaborators "Project Name" --ndjson  # NDJSON output
 td project create --name "New Project" --color "blue"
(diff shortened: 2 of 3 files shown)


src/commands/comment/view.ts
```diff
@@ -11,6 +11,11 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
     const id = lenientIdRef(commentId, 'comment')
     const comment = await api.getComment(id)
 
+    if (options.json) {
+        console.log(formatJson(comment, 'comment', options.full))
```
doistbot: [P2] The new JSON path drops the comment URL entirely. The text output still prints `URL:` and `output.ts` already knows how to add `webUrl` for `comment` entities, so this branch should pass the `showUrl` argument when calling `formatJson` so `td comment view --json` exposes the same navigational data.

src/lib/skills/content.ts
```diff
@@ -213,6 +213,8 @@ td project list
 td project list --personal                    # Personal projects only
 td project view "Project Name"
 td project collaborators "Project Name"
+td project collaborators "Project Name" --json   # JSON output
```
doistbot, line 216: [P2] This update adds examples for `--json`/`--ndjson`, but the new `--full` flags introduced in this PR still aren’t reflected in `SKILL_CONTENT`. Since this file is the source of truth for installed agent skills, the generated skill docs will remain out of sync with the CLI surface until those new flags are documented here as well.

PR #223, opened 2026-04-03
fix: trim Todoist CLI skill content

gnapse:
## Summary
This trims the Todoist CLI skill documentation by consolidating repeated guidance into shared sections and keeping only representative command examples.

It rewrites `src/lib/skills/content.ts` to emphasize core patterns, shared flags, and grouped command families instead of exhaustive per-command repetition.

It also regenerates `skills/todoist-cli/SKILL.md` so the published skill stays in sync with the source.

### DIFF
--- a/src/lib/skills/content.ts
+++ b/src/lib/skills/content.ts
@@ -80,7 +80,7 @@ td task browse "Plan sprint"
 \`\`\`
 
 Useful task flags:
-- \`--stdin\` reads description or comment content from stdin.
+- \`--stdin\` reads the task description from stdin.
 - \`--parent\`, \`--section\`, \`--project\`, \`--workspace\`, \`--assignee\`, \`--labels\`, \`--due\`, \`--deadline\`, \`--duration\`, and \`--priority\` cover most task workflows.
 - \`td task complete --forever\` stops recurrence; \`td task update --no-deadline\` clears deadlines; \`td task move --no-parent\` and \`--no-section\` detach from hierarchy.
 
@@ -117,12 +117,14 @@ td label view "urgent"
 td label create --name "urgent" --color red
 td label update "urgent" --color orange
 td label delete "urgent" --yes
+td label browse "urgent"
 
 td filter list
 td filter view "Urgent work"
 td filter create --name "Urgent work" --query "p1 & #Work"
 td filter update "Urgent work" --query "p1 & #Work & today"
 td filter delete "Urgent work" --yes
+td filter browse "Urgent work"
 
 td section list "Roadmap"
 td section create --project "Roadmap" --name "In Progress"
@@ -130,6 +132,7 @@ td section update id:123 --name "Done"
 td section archive id:123
 td section unarchive id:123
 td section delete id:123 --yes
+td section browse id:123
 \`\`\`
 
 Shared labels can appear in \`td label list\` and \`td label view\`, but standard update and delete actions only work for labels with IDs.
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -81,7 +81,7 @@ td task browse "Plan sprint"
 ```
 
 Useful task flags:
-- `--stdin` reads description or comment content from stdin.
+- `--stdin` reads the task description from stdin.
 - `--parent`, `--section`, `--project`, `--workspace`, `--assignee`, `--labels`, `--due`, `--deadline`, `--duration`, and `--priority` cover most task workflows.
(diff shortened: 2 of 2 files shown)


src/lib/skills/content.ts
```diff
@@ -6,91 +6,45 @@ export const SKILL_CONTENT = `# Todoist CLI (td)
...
-- \`--cursor <cursor>\` - Continue from pagination cursor
-- \`--show-urls\` - Show web app URLs for each item
-
-## Global Options
-
-- \`--no-spinner\` - Disable loading animations
-- \`--progress-jsonl\` - Machine-readable progress events (JSONL to stderr)
-- \`-v, --verbose\` - Verbose output to stderr (repeat: -v info, -vv detail, -vvv debug, -vvvv trace)
-- \`--accessible\` - Add text labels to color-coded output (due:/deadline:/~ prefixes, ★ for favorites). Also: \`TD_ACCESSIBLE=1\`
+- Read and list commands commonly support \`--json\`. List-style commands usually also support \`--ndjson\`, \`--full\`, \`--raw\`, \`--limit <n>\`, \`--all\`, \`--cursor <cursor>\`, and \`--show-urls\`.
```
doistbot: [P2] This shared-flags summary is broader than the actual command surface. Several list/read families now contradict it: `workspace list` has no `--limit/--all/--cursor/--show-urls`, `notification list` paginates with `--offset` instead of `--cursor`, and `activity` supports `--cursor` but not `--all`.

Because the per-command sections are now much shorter, this bullet becomes the main guidance agents rely on. Tighten the wording or call out the common exceptions so the skill doesn't steer callers toward invalid flags.

src/lib/skills/content.ts
```diff
@@ -99,393 +53,143 @@ Tasks, projects, labels, and filters can be referenced by:
...
+td project permissions
+td workspace list
+td workspace view "Acme"
+td workspace projects "Acme"
+td workspace users "Acme" --role ADMIN,MEMBER
+td workspace insights "Acme" --project-ids "id1,id2"
 \`\`\`
 
-### Labels
+### Labels, Filters, And Sections
```
doistbot, line 113: [P2] This condensed family block drops some real subcommands from the shipped reference. `label browse`, `filter browse`, and `section browse` all exist in the CLI, but after this rewrite they no longer appear anywhere in the skill content.

Since this file is the installed command reference for agents, removing those names makes those capabilities effectively undiscoverable unless the model already guesses to inspect help. Consider listing the browse subcommands explicitly here or in the quick reference.

src/lib/skills/content.ts
```diff
@@ -99,393 +53,143 @@ Tasks, projects, labels, and filters can be referenced by:
...
+td task reschedule "Plan sprint" 2026-03-20T14:00:00
+td task move "Plan sprint" --project "Personal" --no-section
+td task complete "Plan sprint"
+td task uncomplete id:123456
+td task delete "Plan sprint" --yes
+td task browse "Plan sprint"
+\`\`\`
+
+Useful task flags:
+- \`--stdin\` reads description or comment content from stdin.
```
doistbot: [P3] `--stdin` is described too broadly here. For task commands it only reads the task description from stdin; “comment content” belongs to `td comment add --stdin`, not `td task add/update --stdin`.

Narrowing this to task description input would avoid suggesting a task flag behavior that the CLI does not support.

src/lib/skills/content.ts
henningmu, line 5: strange that we have to manage the content twice. Did we consider consuming the SKILL file here and stripping the frontmatter instead? 🤔

henningmu (commented):
Nice 👏

Would be amazing if we had some evals to assert this actually kept / improved the accuracy of the skill 🤔

src/lib/skills/content.ts
gnapse: I'll leave that for @scottlovegrove to answer. He was the one who looked into having the skill as markdown so it could be installed through npx.

scottlovegrove: Yes, we need an actual skills folder, with relevant files in it in order for the `npx skills add` command to work. The reason it's all duplicated is because we need the skill content in a javascript file for easier distribution via the npm package (although I remain unconvinced by this as we can package the skill.md 🤷🏻‍♂️). 

That said, in general, we don't need to manage it twice ourselves, it's all automated, and if you are updating via an agent, it already knows that whatever goes in the content, needs to go in the skills.md as well. And in case it forgets, there's a script to keep it up to date, and CI checks as well.

PR #224, opened 2026-04-03
feat: add td doctor command

gnapse:
## Summary

Add a new top-level `td doctor` command that checks config health, Todoist authentication, and npm update status, with `--offline` and `--json` modes.

Refactor update version-checking into a shared helper so `td update` and `td doctor` use the same npm registry and semver logic.

Add tests for the new diagnostics flow and update the generated skill docs to document the new command.

## Example

```bash
node dist/index.js doctor --offline
```

```text
SKIP Auth validation skipped (--offline); credentials found via secure-store
SKIP Skipped npm registry check (--offline); current version is v1.38.2
Doctor summary: 0 passed, 2 skipped
```

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -4,6 +4,7 @@ import chalk from 'chalk'
 import { Command } from 'commander'
 import packageJson from '../../package.json' with { type: 'json' }
 import { CONFIG_PATH, NoTokenError, TOKEN_ENV_VAR, probeApiToken } from '../lib/auth.js'
+import { validateConfigForDoctor } from '../lib/config.js'
 import {
     compareVersions,
     fetchLatestVersion,
@@ -116,24 +117,16 @@ async function checkConfigFile(): Promise<DoctorCheck | null> {
             }
         }
 
-        const warnings: string[] = []
-        const config = parsed as Record<string, unknown>
-        if (
-            config.update_channel !== undefined &&
-            config.update_channel !== 'stable' &&
-            config.update_channel !== 'pre-release'
-        ) {
-            warnings.push(`has unknown update channel "${String(config.update_channel)}"`)
-        }
+        const issues = validateConfigForDoctor(parsed)
 
         return {
             name: 'config',
-            status: warnings.length > 0 ? 'warn' : 'pass',
+            status: issues.length > 0 ? 'warn' : 'pass',
             message:
-                warnings.length > 0
-                    ? `Config file is readable but ${warnings.join('; ')} (${CONFIG_PATH})`
+                issues.length > 0
+                    ? `Config file is readable but ${issues.join('; ')} (${CONFIG_PATH})`
                     : `Config file is readable (${CONFIG_PATH})`,
-            details: { path: CONFIG_PATH, exists: true },
+            details: { path: CONFIG_PATH, exists: true, issues },
         }
     } catch (error) {
         if (error instanceof Error && 'code' in error && error.code === 'ENOENT') {
--- a/src/__tests__/doctor.test.ts
+++ b/src/__tests__/doctor.test.ts
@@ -155,6 +155,44 @@ describe('doctor command', () => {
         expect(process.exitCode).toBeUndefined()
     })
 
+    it('warns when config fields are invalid or unrecognized', async () => {
+        mockReadFile.mockResolvedValue(
+            JSON.stringify({
+                pendingSecureStoreClear: 'yes',
+                auth_mode: 'admin',
+                update_channel: 'beta',
+                extra_setting: true,
+            }),
+        )
+        mockFetch('1.0.0')
+
+        const program = createProgram()
+        await program.parseAsync(['node', 'td', 'doctor'])
+
+        const configWarning = consoleSpy.mock.calls.find(
+            (call: unknown[]) =>
+                typeof call[0] === 'string' &&
+                (call[0] as string).includes('WARN Config file is readable but'),
+        )?.[0]
+
+        expect(configWarning).toEqual(expect.any(String))
+        expect(consoleSpy).toHaveBeenCalledWith(
+            expect.stringContaining('WARN Config file is readable but'),
+        )
+        expect(configWarning).toContain('contains unrecognized key "extra_setting"')
+        expect(configWarning).toContain('pendingSecureStoreClear must be a boolean')
+        expect(configWarning).toContain('auth_mode must be one of: read-only, read-write, unknown')
+        expect(configWarning).toContain('update_channel must be one of: stable, pre-release')
+        expect(consoleSpy).toHaveBeenCalledWith(
+            expect.stringContaining('PASS Authenticated as person@example.com via secure-store'),
+        )
+        expect(consoleSpy).toHaveBeenCalledWith(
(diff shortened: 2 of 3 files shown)


src/lib/update.ts
```diff
@@ -0,0 +1,58 @@
...
+    return channel === 'pre-release' ? 'next' : 'latest'
+}
+
+export function parseVersion(version: string): ParsedVersion {
+    const [core, ...rest] = version.replace(/^v/, '').split('-')
+    const [major, minor, patch] = core.split('.').map(Number)
+    return { major, minor, patch, prerelease: rest.length > 0 ? rest.join('-') : undefined }
+}
+
+/** Returns true when `candidate` is strictly newer than `current` per semver. */
```
doistbot: [P2] Duplicate version comparison logic

The `isNewer` logic replicates the semver parsing logic in `compareVersions` introduced in `src/commands/doctor.ts`. Consider exporting a single `compareVersions` function here to prevent divergence, and use it to implement `isNewer`.

```typescript
export function compareVersions(a: string, b: string): number {
    const left = parseVersion(a)
    const right = parseVersion(b)

    for (const key of ['major', 'minor', 'patch'] as const) {
        if (left[key] !== right[key]) return left[key] > right[key] ? 1 : -1
    }

    if (!left.prerelease && right.prerelease) return 1
    if (left.prerelease && !right.prerelease) return -1

    if (left.prerelease && right.prerelease) {
        return left.prerelease.localeCompare(right.prerelease, undefined, { numeric: true })
    }

    return 0
}

/** Returns true when `candidate` is strictly newer than `current` per semver. */
export function isNewer(current: string, candidate: string): boolean {
    return compareVersions(candidate, current) > 0
}
```

src/commands/doctor.ts
```diff
@@ -0,0 +1,336 @@
...
+            status: 'fail',
+            message: `Could not read config file ${CONFIG_PATH}: ${message}`,
+            details: { path: CONFIG_PATH },
+        }
+    }
+}
+
+async function checkAuthentication(offline: boolean): Promise<DoctorCheck> {
+    try {
+        await getApiToken()
```
doistbot: [P2] `getApiToken()` is not a read-only probe here. When a user still has a legacy `api_token` in the config file and the secure store is available, this call migrates the token and rewrites the config before `getAuthMetadata()` runs, so a single `td doctor` invocation can warn that the config contains a plaintext token and then immediately report auth `via secure-store`. That makes the diagnostics self-contradictory and gives `doctor` an unexpected side effect.

Please use a non-mutating credential presence/metadata check for this path so `td doctor` reports the current state without changing it.

src/lib/update.ts
```diff
@@ -0,0 +1,58 @@
+import { readConfig, type UpdateChannel } from './config.js'
+
+const PACKAGE_NAME = '@doist/todoist-cli'
```
doistbot: [P2] The package name is now duplicated between this helper and `src/commands/update/action.ts`. If the install target ever changes, `td doctor` / `td update --check` can end up querying one npm package while `td update` installs another.

It would be safer to share a single exported constant (or pass the package name into the helper) so the registry lookup and install command cannot drift apart.

doistbot (commented):
This PR introduces a helpful new `td doctor` command for diagnosing runtime, configuration, and authentication issues, while also thoughtfully refactoring version-checking logic to be shared with `td update`. These additions will significantly improve the troubleshooting experience and make version management more consistent across the CLI. A few minor adjustments are noted to ensure the doctor command doesn't inadvertently mutate legacy token state during its checks, and to consolidate duplicate version-comparison logic and hardcoded package names for better long-term maintainability.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F224&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A43e5bba0-2fa3-11f1-97fe-facb4ed63a38%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A43e5bba0-2fa3-11f1-97fe-facb4ed63a38&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

gnapse:
> > PASS Node.js v22.22.0 satisfies >=20.18.1
> 
> I don't understand the point of this? If they are not on a supported node version, wouldn't the package either not install, or not run?

Good call. I changed `doctor` so the Node version check only shows up when it’s actionable, and we no longer print a success line for the normal case.

I kept the check because it can still catch edge cases like older global installs, `--ignore-engines`, or someone downgrading Node after install, but I agree it shouldn’t take up space in the happy path.

src/commands/doctor.ts
```diff
@@ -0,0 +1,309 @@
...
+            return {
+                name: 'config',
+                status: 'fail',
+                message: `Config file must contain a JSON object (${CONFIG_PATH})`,
+                details: { path: CONFIG_PATH },
+            }
+        }
+
+        const warnings: string[] = []
+        const config = parsed as Record<string, unknown>
```
scottlovegrove: Don't we have a config type we can use here instead of a Record? Would make the checks in the next lines more type safe.

src/lib/update.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export async function fetchLatestVersion(channel: UpdateChannel): Promise<string> {
+    const url = `https://registry.npmjs.org/${PACKAGE_NAME}/${getInstallTag(channel)}`
+    const response = await fetch(url)
+    if (!response.ok) {
+        throw new Error(`Registry request failed (HTTP ${response.status})`)
+    }
+    const data = (await response.json()) as RegistryResponse
+    return data.version
+}
```
scottlovegrove, line 57: We already have all this code in `update/actions.ts`

gnapse: `update/actions.ts` already imports this helper function 🤔

src/commands/doctor.ts
gnapse: Fixing it.

gnapse: This made me not only fix the typecast issue, but I added a validation of the config file object shape. Since we do not have `zod` as a dependency here, I created an ad-hoc config object shape validation function that's a bit more verbose, but I thought that was better than introducing `zod` if we have not needed it already.

PR #225, opened 2026-04-05
feat: structured JSON errors, --quiet flag, and IDs in mutation output

scottlovegrove:
## Summary
- Add `CliError` class with structured error codes and hints, with JSON error output on stderr when `--json` is active
- Wrap Todoist SDK errors (401/403/404/429) with context-aware `CliError` codes (`AUTH_ERROR`, `NOT_FOUND`, `RATE_LIMITED`, `API_ERROR`)
- Add global `--quiet`/`-q` flag: create commands print bare ID, other mutations produce zero stdout
- Include `(id:...)` in all mutation success messages (update, delete, archive, complete, move, etc.)

Inspired by [7 Principles for Agent-Friendly CLIs](https://trevinsays.com/p/7-principles-for-agent-friendly-clis). This addresses Principles 2 (Structured Output), 3 (Actionable Errors), 4 (Safe Retries), and 7 (Bounded Responses).

## Test plan
- [x] All 1174 tests pass
- [x] Type check clean
- [x] Lint clean
- [x] Format clean
- [ ] Manual: `td task add "test" --json` with invalid token → JSON error on stderr
- [ ] Manual: `td task add "test" --quiet` → prints only task ID
- [ ] Manual: `td task delete "test" --yes` → message includes `(id:...)`

### DIFF
--- a/src/commands/task/complete.ts
+++ b/src/commands/task/complete.ts
@@ -29,7 +29,7 @@ export async function completeTask(
 
     if (options.forever) {
         const isRecurring = task.due?.isRecurring ?? false
-        if (!isRecurring) {
+        if (!isRecurring && !isQuiet()) {
             console.log('Task is not recurring, completing normally.')
         }
         await completeTaskForever(task.id)
--- a/src/index.ts
+++ b/src/index.ts
@@ -193,8 +193,10 @@ initializeLogger()
 program
     .parseAsync()
     .catch((err: Error) => {
-        if (err instanceof CliError && isJsonMode()) {
-            console.error(formatErrorJson(err.code, err.message, err.hints))
+        if (isJsonMode()) {
+            const code = err instanceof CliError ? err.code : 'INTERNAL_ERROR'
+            const hints = err instanceof CliError ? err.hints : undefined
+            console.error(formatErrorJson(code, err.message, hints))
         } else {
             console.error(err.message)
         }
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -22,7 +22,8 @@ Use this skill when the user wants to interact with their Todoist tasks.
 - Create and update commands commonly support `--json` to return the created or updated entity.
 - Mutating commands support `--dry-run` to preview actions without executing them.
 - Destructive commands typically require `--yes`.
-- Global flags: `--no-spinner`, `--progress-jsonl`, `-v/--verbose`, `--accessible`.
+- `--quiet` / `-q` suppresses success messages. Create commands still print the bare ID for scripting (e.g. `id=$(td task add "Buy milk" --quiet)`).
+- Global flags: `--no-spinner`, `--progress-jsonl`, `-v/--verbose`, `--accessible`, `--quiet`.
 
 ## Authentication
 
--- a/src/lib/skills/content.ts
+++ b/src/lib/skills/content.ts
@@ -21,7 +21,8 @@ Use this skill when the user wants to interact with their Todoist tasks.
 - Create and update commands commonly support \`--json\` to return the created or updated entity.
 - Mutating commands support \`--dry-run\` to preview actions without executing them.
 - Destructive commands typically require \`--yes\`.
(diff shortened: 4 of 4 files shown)


src/index.ts
```diff
@@ -14,13 +16,15 @@ program
     .option('--progress-jsonl [path]', 'Output progress events as JSONL to stderr or file')
     .option('-v, --verbose', 'Increase output verbosity (repeat up to 4x: -v, -vv, -vvv, -vvvv)')
     .option('--accessible', 'Add text labels to color-coded output (also: TD_ACCESSIBLE=1)')
+    .option('-q, --quiet', 'Suppress success messages (create commands still print the ID)')
```
doistbot, line 19: [P1] Missing documentation in skill content

The `--quiet` flag is missing from `src/lib/skills/content.ts`. The repository guidelines require that all new flags be documented in the skill content file to keep AI agents informed of CLI capabilities and to ensure the CI synchronization check passes.

src/index.ts
```diff
@@ -189,7 +193,11 @@ initializeLogger()
 program
     .parseAsync()
     .catch((err: Error) => {
-        console.error(err.message)
+        if (err instanceof CliError && isJsonMode()) {
+            console.error(formatErrorJson(err.code, err.message, err.hints))
+        } else {
+            console.error(err.message)
```
doistbot, line 201: [P2] Non-CliError exceptions bypass JSON output mode

When `--json` is active, non-`CliError` exceptions (such as local validation errors or network failures) are printed as plain text. To ensure agents reliably receive structured output on stderr for all failures, consider formatting unhandled errors as JSON when `isJsonMode()` is true.

```suggestion
            console.error(isJsonMode() ? formatErrorJson('INTERNAL_ERROR', err.message) : err.message)
```

src/lib/output.ts
```diff
@@ -372,6 +372,18 @@ export function formatError(code: string, message: string, hints?: string[]): st
...
+export function formatErrorJson(code: string, message: string, hints?: string[]): string {
+    return JSON.stringify({ error: { code, message, hints } })
+}
+
+export function isJsonMode(): boolean {
+    return process.argv.includes('--json')
+}
+
+export function isQuiet(): boolean {
+    return process.argv.includes('--quiet') || process.argv.includes('-q')
```
doistbot, line 384: [P2] Fragile argument parsing for short flags

Relying on exact string matches in `process.argv` will fail to detect grouped short flags (e.g., `-vq`). It also produces false positives if `--quiet` or `-q` is passed as an option value (e.g., `td task add "--quiet"`). Consider checking Commander's parsed global options instead, or parsing more defensively.

src/commands/task/complete.ts
```diff
@@ -33,10 +33,10 @@ export async function completeTask(
             console.log('Task is not recurring, completing normally.')
         }
         await completeTaskForever(task.id)
-        console.log(`Completed forever: ${task.content}`)
+        if (!isQuiet()) console.log(`Completed forever: ${task.content} (id:${task.id})`)
```
doistbot, line 36: [P2] Quiet mode is still noisy on the non-recurring `--forever` path. When `options.forever` is used on a normal task, the preceding `Task is not recurring, completing normally.` log still goes to stdout, so `td task complete ... --forever --quiet` does not satisfy the new “zero stdout for non-create mutations” behavior. Guard that informational line with `!isQuiet()` as well, or move it under the same quiet check.

src/index.ts
scottlovegrove: Fixed — added `--quiet` to skill content and ran `sync:skill`.

src/index.ts
scottlovegrove: Fixed — all errors now output JSON when `--json` is active, using `INTERNAL_ERROR` code for non-CliError exceptions.

src/lib/output.ts
scottlovegrove: Acknowledged. This is consistent with how `isAccessible()`, spinner, logger, and progress all check flags via `process.argv.includes()` in the codebase. Opened #229 to track improving this across the board.

src/commands/task/complete.ts
scottlovegrove: Fixed — guarded with `!isQuiet()`.

PR #253, opened 2026-04-11
feat: add backup list and download commands

scottlovegrove:
## Summary

- Add `td backup list` and `td backup download <version> --output-file <path>` commands using the SDK's `getBackups()` and `downloadBackup()` methods
- Add `backups:read` OAuth scope to both read-write and read-only scope sets so new logins get backup access
- Existing tokens missing the scope get a clear error message directing users to re-authenticate via `td auth login`

## Test plan

- [x] `npm run type-check` passes
- [x] `npm test` — all 1281 tests pass (7 new backup tests)
- [x] `npm run check` — lint and format clean
- [x] `npm run sync:skill` — skill file regenerated
- [x] Verified `td backup list` with existing token shows actionable scope error
- [ ] After re-auth with new scopes, verify `td backup list` returns backups
- [ ] Verify `td backup download <version> --output-file backup.zip` saves a valid ZIP

### DIFF
--- a/src/commands/backup/download.ts
+++ b/src/commands/backup/download.ts
@@ -1,0 +1,51 @@
+import fs from 'node:fs'
+import path from 'node:path'
+import { getApi } from '../../lib/api/core.js'
+import { CliError } from '../../lib/errors.js'
+import { isQuiet } from '../../lib/global-args.js'
+
+interface DownloadBackupOptions {
+    outputFile: string
+}
+
+export async function downloadBackup(
+    version: string,
+    options: DownloadBackupOptions,
+): Promise<void> {
+    const api = await getApi()
+
+    let backups
+    try {
+        backups = await api.getBackups()
+    } catch (error) {
+        if (error instanceof CliError && error.code === 'AUTH_ERROR') {
+            throw new CliError(
+                'AUTH_ERROR',
+                'Failed to access backups. Your token may be missing the backups:read scope.',
+                [
+                    'Re-authenticate to grant backup access: td auth login',
+                    'If using an API token, ensure it has the backups:read scope',
+                ],
+            )
+        }
+        throw error
+    }
+
+    const backup = backups.find((b) => b.version === version)
+    if (!backup) {
+        throw new CliError('NOT_FOUND', `Backup version "${version}" not found`, [
+            'Run `td backup list` to see available backup versions',
+        ])
+    }
+
+    const response = await api.downloadBackup({ file: backup.url })
+    const arrayBuffer = await response.arrayBuffer()
+    const buffer = Buffer.from(arrayBuffer)
+
+    const outputPath = path.resolve(options.outputFile)
+    fs.writeFileSync(outputPath, buffer)
+
+    if (!isQuiet()) {
+        console.log(`Backup written to ${outputPath}`)
+    }
+}
--- a/src/commands/backup/download.ts
+++ b/src/commands/backup/download.ts
@@ -3,6 +3,7 @@ import path from 'node:path'
 import { getApi } from '../../lib/api/core.js'
 import { CliError } from '../../lib/errors.js'
 import { isQuiet } from '../../lib/global-args.js'
+import { fetchBackups } from './helpers.js'
 
 interface DownloadBackupOptions {
     outputFile: string
@@ -13,23 +14,7 @@ export async function downloadBackup(
     options: DownloadBackupOptions,
 ): Promise<void> {
     const api = await getApi()
-
-    let backups
-    try {
-        backups = await api.getBackups()
-    } catch (error) {
-        if (error instanceof CliError && error.code === 'AUTH_ERROR') {
-            throw new CliError(
-                'AUTH_ERROR',
-                'Failed to access backups. Your token may be missing the backups:read scope.',
-                [
-                    'Re-authenticate to grant backup access: td auth login',
-                    'If using an API token, ensure it has the backups:read scope',
-                ],
-            )
-        }
-        throw error
-    }
+    const backups = await fetchBackups(api)
 
     const backup = backups.find((b) => b.version === version)
     if (!backup) {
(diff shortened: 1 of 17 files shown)


src/commands/backup/download.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export async function downloadBackup(
+    version: string,
+    options: DownloadBackupOptions,
+): Promise<void> {
+    const api = await getApi()
+
+    let backups
+    try {
```
doistbot: [P3] The error handling logic that detects missing `backups:read` scopes and throws a helpful `CliError` is duplicated identically here and in `list.ts`. Consider extracting a shared helper function (e.g., `async function fetchBackups(api)`) to centralize this check.

src/commands/backup/download.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    const backup = backups.find((b) => b.version === version)
+    if (!backup) {
+        throw new CliError('NOT_FOUND', `Backup version "${version}" not found`, [
+            'Run `td backup list` to see available backup versions',
+        ])
+    }
+
+    const response = await api.downloadBackup({ file: backup.url })
```
doistbot, line 26: [P1] Check the download response before writing it to disk. This path consumes `arrayBuffer()` unconditionally, so a 403/404/5xx response (expired backup URL, transient CDN error, etc.) will be saved as `backup.zip` and then reported as success. `attachment view` already guards `response.ok`; this command should do the same before reading the body.

src/commands/backup/list.ts
```diff
@@ -0,0 +1,50 @@
...
+    json?: boolean
+    ndjson?: boolean
+}
+
+export async function listBackups(options: ListBackupsOptions): Promise<void> {
+    const api = await getApi()
+
+    let backups
+    try {
+        backups = await api.getBackups()
```
doistbot: [P1] This should follow the CLI's normal API contract and destructure `getBackups()` as a paginated response. Repository guidance says SDK list endpoints return `{ results, nextCursor }`, so treating the return value as a bare array will make `backups.length` here and `.find()` in `download.ts` fail at runtime. Please update these calls and the new mocks/tests to use the real response shape.

src/commands/backup/list.ts
```diff
@@ -0,0 +1,50 @@
...
+    } catch (error) {
+        if (error instanceof CliError && error.code === 'AUTH_ERROR') {
+            throw new CliError(
+                'AUTH_ERROR',
+                'Failed to access backups. Your token may be missing the backups:read scope.',
```
doistbot: [P2] This rewrites every `AUTH_ERROR` into a missing-scope diagnosis, but `wrapApiError()` uses `AUTH_ERROR` for any 401/403, including invalid or revoked tokens. In those cases the user gets pointed at the wrong fix. It would be safer to preserve the original auth error unless the response specifically indicates a missing `backups:read` scope.

src/commands/backup/list.ts
```diff
@@ -0,0 +1,50 @@
...
+                [
+                    'Re-authenticate to grant backup access: td auth login',
+                    'If using an API token, ensure it has the backups:read scope',
+                ],
+            )
+        }
+        throw error
+    }
+
+    if (options.json) {
```
doistbot, line 16: [P2] `backup list` is bypassing the shared JSON/NDJSON list helpers, so its machine-readable contract now differs from the rest of the CLI: JSON omits `nextCursor`, and NDJSON adds a bespoke `_type` wrapper. Reusing `formatPaginatedJson` / `formatPaginatedNdjson` would keep scripting output consistent with other list commands.

src/lib/oauth.ts
```diff
@@ -4,8 +4,8 @@ import { DEFAULT_PORT, getRedirectUri } from './oauth-server.js'
 const TODOIST_CLIENT_ID = '04863cc1e3584830a578622f50224d5b'
 const OAUTH_AUTHORIZE_URL = 'https://todoist.com/oauth/authorize'
 const OAUTH_TOKEN_URL = 'https://todoist.com/oauth/access_token'
-export const READ_WRITE_SCOPES = 'data:read_write,data:delete,project:delete'
-export const READ_ONLY_SCOPES = 'data:read'
+export const READ_WRITE_SCOPES = 'data:read_write,data:delete,project:delete,backups:read'
+export const READ_ONLY_SCOPES = 'data:read,backups:read'
```
doistbot, line 8: [P3] Changing the scope constants here makes some existing auth copy stale immediately: `td auth login --help` and the read-only fallback text in `td auth status` still say `data:read`. Please update those user-facing strings alongside this scope change so the CLI reports the scopes it actually requests.

doistbot (commented):
This PR introduces helpful new `td backup list` and `td backup download` commands while updating OAuth scopes to properly include `backups:read`. These additions will provide a valuable data recovery workflow for users, though a few implementation details need refinement to ensure runtime stability. Some adjustments are needed to correctly destructure the paginated SDK responses, properly validate HTTP download responses before writing to disk, ensure scope-specific error handling doesn't mask general auth failures, and align the output formatting and help text with the rest of the CLI.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F253&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0d1be4e0-35e9-11f1-93bf-a5512e6fe8ea%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0d1be4e0-35e9-11f1-93bf-a5512e6fe8ea&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

PR #255, opened 2026-04-12
feat: add workspace CRUD and reporting commands

scottlovegrove:
## Summary
- Adds five new workspace subcommands wrapping previously-unused SDK endpoints: `create`, `update`, `delete`, `user-tasks`, and `activity`.
- `update` and `delete` run an `assertWorkspaceAdmin` check **before** the `--dry-run` short-circuit, so a non-admin's preview fails loudly with `NOT_ADMIN` instead of silently claiming the operation would succeed.
- Follows existing CRUD conventions: `--dry-run`, `--yes` on destructive ops, `--json`/`--ndjson`/`--full` on reporting commands, `isQuiet()` → bare ID for scripting.

## Commands

| Subcommand | SDK call | Admin-only |
|---|---|---|
| `td workspace create` | `addWorkspace` | — |
| `td workspace update <ref>` | `updateWorkspace` | **yes** |
| `td workspace delete <ref>` | `deleteWorkspace` | **yes** (requires `--yes`) |
| `td workspace user-tasks <ref> --user <ref>` | `getWorkspaceUserTasks` | — |
| `td workspace activity <ref>` | `getWorkspaceMembersActivity` | — |

`resolveWorkspaceUserRef` accepts `id:xxx`, email, or full name (fuzzy substring fallback) to make `--user` ergonomic.

## Test plan
- [x] `npm run type-check`
- [x] `npm test` — 1310/1310 pass (58 in workspace suite, including explicit tests that `NOT_ADMIN` fires on `--dry-run` for both update and delete)
- [x] `npm run sync:skill` — SKILL.md regenerated
- [ ] Manual smoke against a live token:
  - `td workspace create --name "Test WS" --dry-run`
  - `td workspace update "Some WS" --description "x" --dry-run` (non-admin should see `NOT_ADMIN`)
  - `td workspace delete "Some WS" --dry-run` (non-admin should see `NOT_ADMIN`)
  - `td workspace user-tasks "Acme" --user me@example.com --json`
  - `td workspace activity "Acme" --json`


src/commands/workspace/helpers.ts
```diff
@@ -1 +1,106 @@
+import type { WorkspaceUser } from '@doist/todoist-sdk'
+import { getApi } from '../../lib/api/core.js'
+import type { Workspace } from '../../lib/api/workspaces.js'
+import { CliError } from '../../lib/errors.js'
+import { extractId, isIdRef } from '../../lib/refs.js'
+
 export const WORKSPACE_ROLES = ['ADMIN', 'MEMBER', 'GUEST']
```
scottlovegrove: I know this isn't part of this PR, but we should just use the const for this from the Todoist SDK

scottlovegrove: Good call — swapped to re-export `WORKSPACE_ROLES` from `@doist/todoist-sdk` in f3d0bd3. Kept the re-export typed as `readonly string[]` so existing callers (`users.ts` `.includes(string)` check, `index.ts` spread into Commander's choices) work without casts.

src/commands/workspace/update.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export async function updateWorkspaceCommand(
+    ref: string,
+    options: UpdateWorkspaceOptions,
+): Promise<void> {
+    const workspace = await resolveWorkspaceRef(ref)
+    // Admin check must run BEFORE dry-run short-circuit so a non-admin's
+    // preview still fails loudly — otherwise a dry-run would lie about
+    // what would happen.
+    assertWorkspaceAdmin(workspace, 'update')
```
doistbot: [P2] This permission check runs before we've validated that any update fields were provided. As a result, `td workspace update <non-admin-ws>` returns `NOT_ADMIN` instead of `NO_CHANGES` when the user forgot all update flags, even though no mutation would be attempted. Build `args` and do the no-op validation first, then keep the admin check ahead of the dry-run/API branch.

src/commands/workspace/user-tasks.ts
```diff
@@ -0,0 +1,93 @@
...
+            : tasks.map((t) => ({
+                  id: t.id,
+                  content: t.content,
+                  projectId: t.projectId,
+                  projectName: t.projectName,
+                  priority: t.priority,
+                  due: t.due?.date ?? null,
+                  isOverdue: t.isOverdue,
+              }))
+        console.log(JSON.stringify({ results: output }, null, 2))
```
doistbot: [P2] This new collection JSON shape omits `nextCursor`, while the CLI's other list/report commands emit `{ results, nextCursor }` even when pagination is not supported (`nextCursor: null`). That makes `workspace user-tasks` and `workspace activity` inconsistent for generic consumers; please include `nextCursor: null` here (or route both through `formatPaginatedJson`).

src/commands/workspace/create.ts
```diff
@@ -0,0 +1,70 @@
...
+        })
+        return
+    }
+
+    const api = await getApi()
+    const workspace = await api.addWorkspace(args)
+    clearWorkspaceCache()
+
+    if (options.json) {
+        console.log(JSON.stringify(workspace, null, 2))
```
doistbot: [P2] `--json` here serializes the raw SDK workspace object, which gives a different contract than the curated workspace JSON used by `td workspace view/list --json`. Scripts that create or update a workspace and then inspect it via the read commands will see two different shapes for the same entity. Please add a workspace formatter (or reuse the existing workspace field mapping) and use it here and in `update.ts`.

doistbot (commented):
This PR successfully introduces five new workspace CRUD and reporting subcommands to the CLI, wrapping previously unused SDK endpoints and adding helpful pre-flight administrative checks. Expanding workspace management capabilities is a great addition, and the comprehensive test coverage provides solid confidence in these new tools. There are just a few minor consistency adjustments to address regarding the execution order of the update command's validation, as well as ensuring the JSON outputs for collections and workspace objects align with the CLI's standard data shapes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F255&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A8a7fcd60-3649-11f1-88e7-0d1e0cdad4c1%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A8a7fcd60-3649-11f1-88e7-0d1e0cdad4c1&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/workspace/update.ts
scottlovegrove: Good catch — reordered in eeca98c. Now `args` is built and the NO_CHANGES guard runs first, so a user who forgets all flags on a non-admin workspace sees "No changes specified" instead of NOT_ADMIN. The admin check is kept ahead of the dry-run/API branch, so non-admins with real intent still fail loudly as originally required. Added an explicit test for the precedence.

src/commands/workspace/user-tasks.ts
scottlovegrove: Added `nextCursor: null` to both `workspace user-tasks` and `workspace activity` `--json` output in eeca98c, matching the `{ results, nextCursor }` shape emitted by other list/report commands. Updated tests to assert it.

src/commands/workspace/create.ts
scottlovegrove: Fixed in eeca98c. Added a `formatWorkspaceJson` helper that emits the same curated shape used by `workspace view --json` / `workspace list --json` (`id, name, plan, role, domainName, memberCount, projectCount`). Both `create` and `update` now do a post-mutation sync refresh to populate the fields the REST response lacks (memberCount/projectCount/domainName), then run it through the helper. Also added a `--full` flag for raw output, with a fallback to the raw SDK response if sync hasn't yet caught up. Tests updated to assert the curated shape.

PR #262, opened 2026-04-13
feat: add td task quickadd subcommand with qa alias

scottlovegrove:
## Summary
- New `td task quickadd` subcommand (alias `td task qa`) that pipes natural-language text through Todoist's Quick Add parser via `api.quickAddTask`. Supports `--stdin`, `--json`, and `--dry-run`.
- Refactors top-level `td add` to delegate to the same shared `quickaddTask` implementation. Drops the post-create `--assignee` fallback (Quick Add parses `+Person Name` inline on shared projects) and gains `--stdin` / `--json` in return.
- Updates `SKILL_CONTENT` with examples plus agent guidance on choosing between `task add` (structured flags: `--deadline`, `--parent`, `--duration`, etc.) and `task quickadd` (inline natural-language).

`td task add` itself (the detailed add with `--assignee`, `--project`, `--due`, ...) is unchanged.

## Test plan
- [x] `npm run type-check`
- [x] `npm test` — 1343/1343 passing, including 8 new tests for `task quickadd` and updated tests for top-level `add`
- [x] `npm run sync:skill`
- [x] Manual: `td task quickadd --help`, `td task qa "Buy milk tomorrow p1 #Shopping" --dry-run`, `echo "Buy milk tomorrow" | td task qa --stdin --dry-run`, `td add "Buy milk" --dry-run`

### DIFF
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -41,7 +41,7 @@ Tokens are stored in the OS credential manager when available, with fallback to 
 ## Quick Reference
 
 - Daily views: `td today`, `td inbox`, `td upcoming`, `td completed`, `td activity`
-- Task lifecycle: `td task list/view/add/update/reschedule/move/complete/uncomplete/delete/browse`
+- Task lifecycle: `td task list/view/add/quickadd/update/reschedule/move/complete/uncomplete/delete/browse` (alias: `td task qa` for `quickadd`)
 - Projects: `td project list/view/create/update/archive/unarchive/archived/delete/move/join/browse/collaborators/permissions`
 - Project analytics: `td project progress/health/health-context/activity-stats/analyze-health`
 - Organization: `td label ...`, `td filter ...`, `td section ...`, `td workspace ...`
@@ -75,6 +75,8 @@ td activity --type task --event completed
 ### Tasks
 ```bash
 td task add "Buy milk" --due tomorrow
+td task quickadd "Buy milk tomorrow p1 #Shopping"
+td task qa "Review PR @urgent +Alice"
 td task list --project "Work" --label "urgent" --priority p1
 td task view "Buy milk"
 td task add "Plan sprint" --project "Work" --section "Planning" --labels "urgent,review"
@@ -87,8 +89,13 @@ td task delete "Plan sprint" --yes
 td task browse "Plan sprint"
 ```
 
+Choosing between `task add` and `task quickadd`:
+- `td task quickadd` (alias `td task qa`) uses Todoist's natural-language parser. Inline syntax covers dates ("tomorrow at 2pm"), priority (`p1`–`p4`), project (`#Project`), labels (`@label`), sections (`/Section`), and assignee (`+Person` on shared projects). **Prefer `quickadd` when all task attributes can be expressed inline and you do not need to set additional structured fields** — it's one call and no name-resolution lookups are required.
+- Use `td task add` when you need flags that Quick Add syntax can't express (`--deadline`, `--description`, `--parent`, `--duration`, `--uncompletable`, `--order`), when the text is being composed programmatically, or when you need explicit `id:` / URL references for project/section/parent.
+- `td task quickadd` supports `--stdin`, `--json`, and `--dry-run` only; everything else is embedded in the text.
+
 Useful task flags:
-- `--stdin` reads the task description from stdin.
+- `--stdin` on `task add` reads the task description from stdin; on `task quickadd` it reads the full natural-language text from stdin.
 - `--parent`, `--section`, `--project`, `--workspace`, `--assignee`, `--labels`, `--due`, `--deadline`, `--duration`, and `--priority` cover most task workflows.
 - `td task complete --forever` stops recurrence; `td task update --no-deadline` clears deadlines; `td task move --no-parent` and `--no-section` detach from hierarchy.
 
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -93,6 +93,7 @@ Choosing between `task add` and `task quickadd`:
 - `td task quickadd` (alias `td task qa`) uses Todoist's natural-language parser. Inline syntax covers dates ("tomorrow at 2pm"), priority (`p1`–`p4`), project (`#Project`), labels (`@label`), sections (`/Section`), and assignee (`+Person` on shared projects). **Prefer `quickadd` when all task attributes can be expressed inline and you do not need to set additional structured fields** — it's one call and no name-resolution lookups are required.
 - Use `td task add` when you need flags that Quick Add syntax can't express (`--deadline`, `--description`, `--parent`, `--duration`, `--uncompletable`, `--order`), when the text is being composed programmatically, or when you need explicit `id:` / URL references for project/section/parent.
 - `td task quickadd` supports `--stdin`, `--json`, and `--dry-run` only; everything else is embedded in the text.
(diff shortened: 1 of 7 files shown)


skills/todoist-cli/SKILL.md
```diff
@@ -87,8 +89,13 @@ td task delete "Plan sprint" --yes
...
 ```
 
+Choosing between `task add` and `task quickadd`:
+- `td task quickadd` (alias `td task qa`) uses Todoist's natural-language parser. Inline syntax covers dates ("tomorrow at 2pm"), priority (`p1`–`p4`), project (`#Project`), labels (`@label`), sections (`/Section`), and assignee (`+Person` on shared projects). **Prefer `quickadd` when all task attributes can be expressed inline and you do not need to set additional structured fields** — it's one call and no name-resolution lookups are required.
+- Use `td task add` when you need flags that Quick Add syntax can't express (`--deadline`, `--description`, `--parent`, `--duration`, `--uncompletable`, `--order`), when the text is being composed programmatically, or when you need explicit `id:` / URL references for project/section/parent.
+- `td task quickadd` supports `--stdin`, `--json`, and `--dry-run` only; everything else is embedded in the text.
+
 Useful task flags:
-- `--stdin` reads the task description from stdin.
+- `--stdin` on `task add` reads the task description from stdin; on `task quickadd` it reads the full natural-language text from stdin.
```
scottlovegrove: It's not just `task quickadd`, it's when the user does `td add --stdin` as well.

scottlovegrove: Good catch — fixed in 4a580fd. Updated the line to call out that `--stdin` reads the full natural-language text on both `td task quickadd` and the top-level `td add`.

doistbot (commented):
This PR introduces the `td task quickadd` subcommand while nicely refactoring the top-level `td add` to share the same underlying Quick Add implementation. These changes successfully unify the natural-language parsing paths and expand input flexibility with `--stdin` and `--json` support. A few minor adjustments are suggested to tighten input validation around whitespace and empty strings, clarify the shared-project limitation for inline assignees in the CLI help, and ensure the top-level `td add` flag changes are fully reflected in the skill content.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F262&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3d8292e0-3743-11f1-83ee-f7fa2a13ac68%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3d8292e0-3743-11f1-83ee-f7fa2a13ac68&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/task/quickadd.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export async function quickaddTask(options: QuickaddOptions): Promise<void> {
+    if (options.text && options.stdin) {
+        throw new CliError(
+            'CONFLICTING_OPTIONS',
+            'Cannot specify text both as argument and --stdin',
+        )
+    }
+
+    const text = options.stdin ? (await readStdin()).trim() : options.text
```
doistbot: [P3] Consider trimming `options.text` as well. If a user passes a string of only whitespace (e.g., `td add "   "`), it currently bypasses the `!text` check and hits the API, likely resulting in an API error. Trimming both sources ensures the `MISSING_CONTENT` error is thrown consistently and avoids an unnecessary network call.

```typescript
    const text = options.stdin ? (await readStdin()).trim() : options.text?.trim()
```

src/commands/task/quickadd.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface QuickaddOptions {
+    text?: string
+    stdin?: boolean
+    json?: boolean
+    dryRun?: boolean
+}
+
+export async function quickaddTask(options: QuickaddOptions): Promise<void> {
+    if (options.text && options.stdin) {
```
doistbot: [P2] This conflict check only fires for truthy positional text. `td task qa "" --stdin` currently skips the error and reads from stdin, even though both input sources were provided. Switching this to an explicit `options.text !== undefined` check would keep the behavior consistent for scripted callers that may pass empty strings.

src/commands/task/index.ts
```diff
@@ -155,6 +158,23 @@ Examples:
             return addTask({ ...options, content })
         })
 
+    const quickaddCmd = task
+        .command('quickadd [text]')
+        .alias('qa')
+        .description(
+            'Quick add a task using natural language (e.g. "Buy milk tomorrow p1 #Shopping +John")',
```
doistbot: [P3] The example here implies `+John` is generally supported, but Todoist only parses assignees inline on shared projects. The skill docs already carry that caveat, so the CLI help should match to avoid sending users down a non-working path in personal projects.

src/commands/add.ts
```diff
@@ -1,47 +1,20 @@
...
 
 export function registerAddCommand(program: Command): void {
     const addCmd = program
         .command('add [text]')
         .description(
-            'Quick add with natural language (human shorthand; agents should use "task add")',
+            'Quick add with natural language (human shorthand for "td task quickadd" / "td task qa")',
         )
-        .option('--assignee <ref>', 'Assign to user (name, email, id:xxx, or "me")')
+        .option('--stdin', 'Read text from stdin')
```
doistbot, line 10: [P2] This changes the top-level `td add` contract by adding `--stdin`/`--json` and removing `--assignee`, but the skill content updates only describe `task quickadd`. Since the installed skill docs are meant to stay aligned with command/flag changes, this top-level shorthand should also be mentioned there so agents do not get stale guidance.

src/commands/task/quickadd.ts
scottlovegrove: Done in 556f782. Both sources are now trimmed (`const raw = options.stdin ? await readStdin() : options.text ?? ''; const text = raw.trim()`), so whitespace-only input errors with `MISSING_CONTENT` before any API call. Added a regression test for this too.

src/commands/task/quickadd.ts
scottlovegrove: Done in 556f782. Switched to `options.text !== undefined && options.stdin` so `td task qa "" --stdin` now errors consistently. Added a test for the empty-string case.

src/commands/task/index.ts
scottlovegrove: Good point — dropped `+John` from the CLI help in 556f782. The example now reads `"Buy milk tomorrow p1 #Shopping"`, which works in both personal and shared projects. Shared-project caveat still lives in the skill docs.

src/commands/add.ts
scottlovegrove: Done in 556f782. Added a line to the skill content explicitly describing the top-level `td add` as the human shorthand for `td task quickadd`, with the same flag surface (`--stdin`, `--json`, `--dry-run`), so agents won't get stale `--assignee` guidance.

rfgamaral (commented):
✅

PR #264, opened 2026-04-13
feat(apps): add td auth login --app-management and td apps list/view

scottlovegrove:
## Summary

Adds opt-in support for managing the user's registered Todoist developer apps. Three commits so each piece is reviewable in isolation:

1. **feat(auth): add --app-management flag to td auth login** (`91fefeb`)
   - Bumps `@doist/todoist-sdk` to 9.1.1.
   - New `--app-management` flag on `td auth login` requests the `dev:app_console` OAuth scope alongside the standard data scopes.
   - Composable with `--read-only` so users can keep data access read-only while still managing their apps.
   - Off by default — granting this scope lets the token rotate client secrets, edit webhooks, delete apps, etc.

2. **feat(apps): add td apps list** (`dc457c5`)
   - New top-level `apps` command surface backed by the SDK's new `AppClient`.
   - Plain output: `<displayName> (id:N)` then a dim description. The `(id:N)` form keeps rows self-describing in `--accessible` mode (no reliance on dim colour as the only id/name separator).
   - `--json` / `--ndjson` dump the full app payload — surface is small and stable enough to round-trip cleanly.
   - Wires a generic missing-scope detector into `wrapApiError`: any 403 with `error_tag: AUTH_INSUFFICIENT_TOKEN_SCOPE` is now surfaced as a friendly `MISSING_SCOPE` error pointing at `td auth login --app-management`. Endpoint-agnostic — any future scope-gated API gets the same hint for free.

3. **feat(apps): add td apps view <ref>** (`b7c2cdb`)
   - Resolves a name (fuzzy/case-insensitive substring), `id:N`, or raw numeric id, then prints the detail record with a labelled key/value block (id, status, users, created date, service URL, OAuth redirect, token scopes, icon URL).
   - `--json` returns the full `AppWithUserCount` payload.
   - Dedicated `resolveAppRef` (not the shared `resolveRef`) because app IDs are pure-numeric — the shared `looksLikeRawId` fallback would 400 on alphanumeric refs.
   - Single fetch on `id:N` / numeric paths, two fetches on the name path (list + detail enrichment for `userCount`).

## What is *not* in this PR

Write operations (`add`, `update`, `delete`, webhooks, installations, user authorizations, app secrets / tokens). The SDK exposes the full surface; this PR is deliberately the read-only foundation. Follow-ups will build on top.

## Test plan

- [x] `npm test` — 1368 tests pass (20 new for apps + scope-detection)
- [x] `npm run check` — lint and format clean
- [x] `npm run check:skill-sync` — SKILL.md in sync
- [x] **Live verified end-to-end against a real account:**
  - [x] `td auth login --app-management` → authorize URL contains `scope=…,dev:app_console`; persisted `auth_scope` reflects it; `GET /apps` returns 200
  - [x] `td auth login --app-management --read-only` → combined scope `data:read,backups:read,dev:app_console`
  - [x] `td apps list` / `--json` / `--ndjson` against an account with 16 registered apps
  - [x] `td apps view "Mark as Completed"` / `id:19550` / `19550` / `--json` / `--accessible`
  - [x] `td apps view "doesnotexist-xyz123"` → friendly `APP_NOT_FOUND` (no spurious 400)
  - [x] `td apps view 99999999` → friendly `APP_NOT_FOUND` (wrapped 404 caught)
  - [x] With an insufficient-scope token: SDK call → friendly `MISSING_SCOPE` with `td auth login --app-management` hint

### DIFF
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -52,6 +52,7 @@ Tokens are stored in the OS credential manager when available, with fallback to 
 - Collaboration: `td comment ...`, `td notification ...`, `td reminder ...`
 - Templates and files: `td template ...`, `td attachment view <file-url>`, `td backup ...`
 - Account and tooling: `td stats`, `td settings ...`, `td completion ...`, `td view <todoist-url>`, `td doctor`, `td update`, `td changelog`
+- Developer apps: `td apps list` (requires `td auth login --app-management`)
 
 ## References
 
@@ -214,6 +215,14 @@ td template import-id "Roadmap" --template-id product-launch --locale fr
 td backup list
 td backup download "2024-01-15_12:00" --output-file backup.zip
 ```
+
+### Developer Apps
+```bash
+td apps list
+td apps list --json
+```
+
+The `apps` command surface manages the user's registered Todoist developer apps (integrations). All `apps` subcommands require the `dev:app_console` OAuth scope — re-run `td auth login --app-management` to grant it. Without the scope, calls fail with a `MISSING_SCOPE` error pointing at the same hint. `td apps list` in plain mode shows id and display name with a dim description; `--json` / `--ndjson` dump the full app payload (id, displayName, status, userId, createdAt, serviceUrl, oauthRedirectUri, description, icons, appTokenScopes).
 
 ### Settings, Stats, And Utilities
 ```bash
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -31,10 +31,14 @@ Use this skill when the user wants to interact with their Todoist tasks.
 ```bash
 td auth login
 td auth login --read-only
+td auth login --app-management
+td auth login --app-management --read-only
 td auth token
 td auth status
 td auth logout
 ```
+
+`--app-management` adds the `dev:app_console` OAuth scope to the requested grant. Combine with `--read-only` to keep data access read-only while still gaining app-management access. Granting this scope is opt-in because it allows the token to manage your registered Todoist apps (rotate secrets, edit webhooks, etc.).
 
 Tokens are stored in the OS credential manager when available, with fallback to `~/.config/todoist-cli/config.json`. `TODOIST_API_TOKEN` takes precedence over stored credentials.
 
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -52,7 +52,7 @@ Tokens are stored in the OS credential manager when available, with fallback to 
 - Collaboration: `td comment ...`, `td notification ...`, `td reminder ...`
 - Templates and files: `td template ...`, `td attachment view <file-url>`, `td backup ...`
 - Account and tooling: `td stats`, `td settings ...`, `td completion ...`, `td view <todoist-url>`, `td doctor`, `td update`, `td changelog`
-- Developer apps: `td apps list` (requires `td auth login --app-management`)
+- Developer apps: `td apps list/view` (requires `td auth login --app-management`)
 
 ## References
 
@@ -220,9 +220,17 @@ td backup download "2024-01-15_12:00" --output-file backup.zip
 ```bash
 td apps list
 td apps list --json
+td apps view "Todoist for VS Code"
+td apps view id:9909
+td apps view 9909
+td apps view id:9909 --json
 ```
 
-The `apps` command surface manages the user's registered Todoist developer apps (integrations). All `apps` subcommands require the `dev:app_console` OAuth scope — re-run `td auth login --app-management` to grant it. Without the scope, calls fail with a `MISSING_SCOPE` error pointing at the same hint. `td apps list` in plain mode shows id and display name with a dim description; `--json` / `--ndjson` dump the full app payload (id, displayName, status, userId, createdAt, serviceUrl, oauthRedirectUri, description, icons, appTokenScopes).
+The `apps` command surface manages the user's registered Todoist developer apps (integrations). All `apps` subcommands require the `dev:app_console` OAuth scope — re-run `td auth login --app-management` to grant it. Without the scope, calls fail with a `MISSING_SCOPE` error pointing at the same hint.
+
+`td apps list` plain output leads with the display name and follows it with `(id:N)` (self-describing in `--accessible` mode). `--json` / `--ndjson` dump the full app payload (id, displayName, status, userId, createdAt, serviceUrl, oauthRedirectUri, description, icons, appTokenScopes).
+
+`td apps view <ref>` accepts a name (fuzzy/case-insensitive), `id:N`, or a raw numeric id. Plain output shows display name as a header, then a labelled key/value block (id, status, users, created date, service URL, OAuth redirect, token scopes, icon URL) followed by the description. `--json` returns the AppWithUserCount payload (App + `userCount`).
 
 ### Settings, Stats, And Utilities
 ```bash
--- a/src/commands/apps/index.ts
+++ b/src/commands/apps/index.ts
@@ -26,14 +26,13 @@ Requires authenticating with the dev:app_console scope:
         .option('--ndjson', 'Output as newline-delimited JSON')
         .action(listApps)
 
-    const viewCmd = apps
-        .command('view [ref]')
+    apps.command('view [ref]', { isDefault: true })
         .description('View details for a single app (by name, id:N, or raw id)')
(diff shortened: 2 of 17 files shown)


src/commands/apps/index.ts
```diff
@@ -0,0 +1,41 @@
...
+        )
+
+    apps.command('list')
+        .description('List your registered Todoist apps')
+        .option('--json', 'Output as JSON')
+        .option('--ndjson', 'Output as newline-delimited JSON')
+        .action(listApps)
+
+    const viewCmd = apps
+        .command('view [ref]')
```
doistbot: [P2] Missing implicit default command. To align with the "implicit view subcommand" pattern described in `AGENTS.md` (e.g., `td project <ref>` defaulting to `td project view <ref>`), `apps view` should be registered with `{ isDefault: true }`. Additionally, when no reference is provided, it should display the parent command's help (`apps.help()`) rather than `viewCmd.help()` so users see all available subcommands.

```typescript
    const viewCmd = apps
        .command('view [ref]', { isDefault: true })
        .description('View details for a single app (by name, id:N, or raw id)')
        .option('--ndjson', 'Output as newline-delimited JSON')
        .action((ref, options) => {
            if (!ref) {
                apps.help()
                return
            }
            return viewApp(ref, options)
        })
```

src/lib/refs.ts
```diff
@@ -282,6 +282,55 @@ export async function resolveProjectRef(api: TodoistApi, ref: string): Promise<P
...
+ *  3. Otherwise → `getApps()` then exact / substring match on `displayName`
+ *     via the in-memory `resolveFromList`, then `getApp(matchedId)` to
+ *     enrich with `userCount`
+ */
+export async function resolveAppRef(api: TodoistApi, ref: string): Promise<AppWithUserCount> {
+    if (!ref.trim()) {
+        throw new CliError('INVALID_APP', 'app reference cannot be empty.')
+    }
+
+    if (isIdRef(ref)) {
```
doistbot: [P2] Missing 404 error transformation for explicit ID references. When a reference uses the `id:` prefix (e.g., `id:9999`), a 404 from `api.getApp()` will bubble up as a generic `NOT_FOUND` error because it returns directly and skips the try/catch block used for raw numeric IDs below. You can combine the logic for both ID paths so they share the same friendly `APP_NOT_FOUND` error handling.

```typescript
    const explicitId = isIdRef(ref) ? extractId(ref) : null
    const idToFetch = explicitId ?? (/^\d+$/.test(ref) ? ref : null)

    if (idToFetch !== null) {
        try {
            return await api.getApp(idToFetch)
        } catch (error) {
            // The api Proxy in core.ts already wraps TodoistRequestError into a
            // generic CliError('NOT_FOUND', …). Catch both shapes so the
            // user-facing error names the entity (App "…" not found) instead
            // of the raw HTTP message.
            if (error instanceof CliError && error.code === 'NOT_FOUND') {
                throw new CliError('APP_NOT_FOUND', `App "${ref}" not found.`)
            }
            if (error instanceof TodoistRequestError && error.httpStatusCode === 404) {
                throw new CliError('APP_NOT_FOUND', `App "${ref}" not found.`)
            }
            throw error
        }
    }
```

doistbot (commented):
This PR introduces excellent new functionality for managing Todoist developer apps via the CLI, adding the `--app-management` auth flag alongside the new `apps list` and `apps view` commands. The implementation is clean and thoughtfully expands the tool's utility for developers while ensuring robust JSON output support. A few adjustments are noted around the generic 403 scope error to prevent conflicts with existing scope-gated commands like backups, alongside minor updates to wire the implicit view subcommand and standardize 404 error handling for explicit ID references.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F264&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A130a86b0-3759-11f1-9a47-a0236bdca6a9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A130a86b0-3759-11f1-9a47-a0236bdca6a9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/api/core.ts
```diff
@@ -161,10 +164,26 @@ function createSpinnerWrappedApi(api: TodoistApi): TodoistApi {
...
+export function wrapApiError(error: unknown): Error {
     if (error instanceof CliError) return error
     if (error instanceof TodoistRequestError) {
         const status = error.httpStatusCode
+        if (status === 403 && isInsufficientScopeError(error)) {
+            return new CliError(
+                'MISSING_SCOPE',
+                'Your token is missing the OAuth scope required for this command.',
+                [
+                    'For app management commands: re-run `td auth login --app-management` (combine with --read-only if desired).',
```
doistbot: [P1] This shared 403 handler now always tells users to re-run `td auth login --app-management`, but `AUTH_INSUFFICIENT_TOKEN_SCOPE` is already reachable from non-app commands. In particular, `td backup ...` only pre-checks `backups:read` when `authScope` metadata is known; users coming from `TODOIST_API_TOKEN` or `td auth token` fall through to the API, and a missing backup scope will now get an app-management hint that cannot fix the problem.

Please keep the transport-layer error generic here, or derive the remediation from the command/method so existing scope-gated commands do not show the wrong re-auth flow.

scottlovegrove: Good catch — fixed in 26dc1f1.

`wrapApiError` now takes the SDK method name and looks up the right remediation via a `METHOD_SCOPE_GROUP` registry: only `getApps`/`getApp` are mapped to the `app-management` group. Everything else (including `getBackups`/`downloadBackup` for env-var token users) falls through to a generic `td auth login` hint, so the wrong re-auth flow is no longer surfaced.

src/commands/apps/index.ts
scottlovegrove: Done in 26dc1f1 — `view` is now `{ isDefault: true }` and the empty-ref branch shows `apps.help()`. Live verified `td apps id:9909` works as shorthand and `td apps view` (no ref) prints the parent help with both `list` and `view` listed.

src/lib/refs.ts
scottlovegrove: Fixed in 26dc1f1. Combined the `id:N` and pure-numeric paths into a single id-bearing branch with shared try/catch, so both surface `APP_NOT_FOUND` on miss. Added a regression test and live-verified `td apps view id:99999999` now produces the friendly error.

skills/todoist-cli/SKILL.md
```diff
@@ -31,11 +31,15 @@ Use this skill when the user wants to interact with their Todoist tasks.
 ```bash
 td auth login
 td auth login --read-only
+td auth login --app-management
```
henningmu, line 34: I wonder if this comes back to haunt us down the road when we add more optional scopes (e.g. `--billing`). Would a generic `--aditional-scopes=app-management,billing` be better? 🤔

scottlovegrove: That's why I did https://github.com/Doist/todoist-cli/pull/265 so that we can keep track of it better

scottlovegrove: Or do you just mean the API for adding scopes in general?

henningmu: The API for adding scopes. My proposal was one argument accepting an array (comma-separated list of scopes) vs. many arguments.

scottlovegrove: ok, i'll update it to a single argument 👍🏻

PR #267, opened 2026-04-14
feat(apps): enrich td apps view + surface client_id in list/view

scottlovegrove:
## Summary

Brings `td apps view <ref>` closer to parity with the web Settings → App Management detail pane, and surfaces each app's OAuth `client_id` on both `td apps list` and `td apps view` now that SDK 9.3.0 carries `clientId` on the `App` schema directly (backend Doist/Todoist#27400).

### `td apps view <ref>`

Alongside the existing `getApp` fetch, the command now pulls the webhook configuration (`getAppWebhook`) by default. With `--include-secrets` it additionally fetches the app's secrets (`getAppSecrets`), verification token, test token, and distribution token.

The OAuth `client_id` is **public** and always shown — it comes straight off the `App` payload, no extra fetch. The four sensitive credentials — client secret, verification token, test access token, distribution token — are **hidden by default**: plain output renders `(hidden — pass --include-secrets to reveal)`, and `--json` / `--ndjson` omit the keys entirely (no `null` sentinel, no placeholder string). With `--include-secrets` the values are rendered / emitted normally; a non-existent test token reads as `(not created)` in that mode.

**Call budget:** default path fires one extra fetch after `resolveAppRef` — `getAppWebhook` only. Secret-bearing endpoints are never touched without the explicit opt-in, satisfying secret-minimization (see earlier review on #267 and the Doist secrets-management handbook).

### `td apps list`

Each row now renders as:

```
Todoist for VS Code (id:9909)
   Client ID: f259fe20fbe24ac88650dcc08cbd5672
   (no description)
```

Zero extra API calls — `client_id` arrives as part of the `getApps` response thanks to SDK 9.3.0. Restores the feature dropped from #264, where fetching the client id per-row would have cost N+1 `getAppSecrets` calls.

## What is *not* in this PR

- No write operations (create/update/delete, reset secrets, create test tokens, update webhooks) — those can land in follow-ups.

## Test plan

- [x] `npm test` — 1396 tests pass (10 new for the enrichment + Client ID behaviour)
- [x] `npm run check:skill-sync` — SKILL.md in sync
- [x] **Live verified end-to-end against a real account with `dev:app_console`:**
  - [x] `td apps list` — 16 apps, each row showing `Client ID: …` inline (dim), no extra API calls
  - [x] `td apps view id:19550` (default) — Client ID shown; four `(hidden …)` lines; `Webhook: (not configured)`. Only `getAppWebhook` fired beyond resolve.
  - [x] `td apps view id:19550 --include-secrets` — all sensitive values rendered; test token reads `(not created)` when null server-side. All 5 enrichment calls fire.
  - [x] `td apps view id:19550 --json` — payload has `clientId` + `webhook`; no `clientSecret` / `verificationToken` / `distributionToken` / `testToken` keys.
  - [x] `td apps view id:19550 --json --include-secrets` — all four sensitive keys present with real values.
  - [x] `td apps view "Mark as Completed"` — name-match path still works and triggers the same enrichment.
  - [x] `td apps view --accessible id:19550` — rendering stays self-describing without colour cues.

## Notes for reviewers

- All 5 new SDK methods (`getAppSecrets`, `getAppVerificationToken`, `getAppTestToken`, `getAppDistributionToken`, `getAppWebhook`) are in the read-only allowlist and mapped to the `app-management` scope group — a 403 from any of them surfaces the `--app-management` remediation hint.
- Spinner label for all 5 is the same `'Loading app...'` as `getApp`, so the concurrent enrichment batch reads as one operation rather than five labels flickering past.


src/commands/apps/view.ts
```diff
@@ -5,19 +5,45 @@ import { resolveAppRef } from '../../lib/refs.js'
...
 
-    if (options.json) {
-        console.log(JSON.stringify(app, null, 2))
-        return
-    }
+    // Always fetch: secrets (for the public clientId) and webhook config.
+    // Verification / test / distribution tokens are only useful when reveal is
+    // on, so gate the fetch to avoid paying for data we can't display.
+    const [secrets, webhook, verification, testToken, distribution] = await Promise.all([
+        api.getAppSecrets(app.id),
```
doistbot: [P2] The default path now downloads `clientSecret` even when the user did not opt into revealing secrets. `getAppSecrets()` returns both `clientId` and `clientSecret`, so `td apps view` will still pull the secret onto the local machine while claiming those values are hidden by default.

If we want that guarantee to apply to data access instead of just rendering, this call needs to stay behind `--include-secrets` (or `Client ID` needs to move behind the flag until the API exposes a client-id-only endpoint). This also goes against our general secret-minimization guidance: https://handbook.doist.com/doc/standard-secrets-management-wbKmIfrtgr

src/commands/apps/index.ts
```diff
@@ -30,6 +30,10 @@ Requires authenticating with the dev:app_console scope:
         .description('View details for a single app (by name, id:N, or raw id)')
         .option('--json', 'Output as JSON')
         .option('--ndjson', 'Output as newline-delimited JSON')
+        .option(
```
doistbot, line 39: [P3] This adds a new user-facing flag, but the `td apps --help` example block above still only shows the old `apps view` invocations. That makes `--include-secrets` discoverable in SKILL.md but not in the CLI's built-in help.

Please add at least one example using this flag to the `.addHelpText()` block here.

doistbot (commented):
This PR nicely enhances the `td apps view` command by bringing it closer to web parity with the addition of OAuth credentials and webhook configurations. The careful attention to formatting and balancing API call budgets is a great touch for maintainability and the overall CLI user experience. A few minor adjustments are needed, specifically ensuring that the client secret is not fetched at all without explicit opt-in to align with secret-minimization guidelines, and adding the new flag to the built-in help text examples for better discoverability.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F267&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac6ce2ba0-37d8-11f1-9fec-b3fdb15189eb%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac6ce2ba0-37d8-11f1-9fec-b3fdb15189eb&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/apps/view.ts
scottlovegrove: Good catch — fixed in 56c5c82.

`getAppSecrets` is now gated behind `--include-secrets` too, so the default path no longer transports `client_secret` onto the user's machine. The `Client ID` line is gated under the same flag as a consequence (it renders `(hidden — pass --include-secrets to reveal)` in plain output and is absent from `--json` by default). Once the backend exposes a client-id-only endpoint we can lift just the Client ID back out.

Default path is now a single extra call after `resolveAppRef`: `getAppWebhook` only. With `--include-secrets`, all five enrichment calls fire.

src/commands/apps/index.ts
scottlovegrove: Done in 56c5c82 — two `--include-secrets` examples added to the `.addHelpText()` block plus a short inline explanation of what the flag does. `td apps --help` now makes the flag (and the fact that secrets aren't pulled without it) discoverable from the CLI itself.

src/commands/apps/view.ts
```diff
@@ -5,19 +5,45 @@ import { resolveAppRef } from '../../lib/refs.js'
...
+const HIDDEN_HINT = '(hidden — pass --include-secrets to reveal)'
+
+function hiddenLine(label: string): string {
+    return `  ${label}${chalk.dim(HIDDEN_HINT)}`
 }
 
 export async function viewApp(ref: string, options: ViewAppOptions = {}): Promise<void> {
     const api = await getApi()
     const app = await resolveAppRef(api, ref)
+    const reveal = Boolean(options.includeSecrets)
```
pedroalves0:
```suggestion
    const revealSecrets = Boolean(options.includeSecrets)
```

super nit-pick

pedroalves0 (approved):
✅

PR #270, opened 2026-04-14
fix(goal): include completed tasks in goal view via --include-completed

lmjabreu:
## Summary

`td goal view <id> --json --full --all` silently drops completed tasks from a goal's linked items. On a goal with `progress.totalTaskCount: 30 / completedTaskCount: 29`, the current command returns `tasks.results` with a single entry — the only *open* task. Verified live against 14 goals: 5 return `[]` (all tasks completed), 3 return `totalTaskCount` only when it is ≤3 (nothing completed yet), rest return the open-count. The pagination loop in `goal view` is not missing — the REST `GET /tasks?goal_id=…` filter is open-only by design.

### Root cause (confirmed against the API)

- `GET /tasks?goal_id=X` only returns open tasks — no `include_completed`, no alternate flag.
- `/goals/{id}/items` (GET) → `405 Method Not Allowed`.
- `/tasks/completed?goal_id=X` and `/tasks/completed/by_completion_date?goal_id=X` both **ignore** the `goal_id` query parameter (verified by inspecting `goal_ids` on the returned items).
- The task filter DSL has no `@goal:` or `goal_id:` operator.

The only signal that survives is the `goal_ids: string[]` field on each completed item — which means client-side filtering, walked over the 3-month-windowed completion-date endpoint.

### Fix

New opt-in `--include-completed` flag on `td goal view`. When set, after the existing open-task pagination, the command walks `/tasks/completed/by_completion_date` in 3-month windows up to `MAX_WINDOWS` (8, ~2 years) back, filters items client-side by `goal_ids.includes(goalId)`, and short-circuits as soon as `goal.progress.completedTaskCount` matches are found. Smoke-tested against the bug-reporter's goals:

| Goal | Default (open) | `--include-completed` | Wall clock |
|---|---|---|---|
| Goals v1.0 (30/29) | 1 task | 30 tasks | ~3.5s |
| Activity Log reboot (17/12) | 5 tasks | 17 tasks | ~3.7s |

### Why this is additive, not a behavior change

Per Luis's direction: existing consumers of `td goal view` — including anyone scripting against `--all` — see zero change. No default fetches change, no JSON fields change, no exit codes change. The completed-task walk only runs when the caller passes `--include-completed`, which also adds two new fields to JSON output (`completedTaskCount`, `completedTasksTruncated`) and a yellow truncation warning in text mode when the walk exhausts the lookback horizon without collecting every match.

`--limit` and `--all` remain the ceiling on the *combined* result set when `--include-completed` is used, so `--limit 5 --include-completed` never exceeds 5 total tasks.

### Scope notes / follow-ups

- **The walk is O(thousands of completed items) for goals with historical completions.** For the "Goals v1.0" test goal the walker scanned ~270 items across 1 window to find 29 matches; goals with older linked tasks can scan ~2,400 items over 12+ windows. Capped at 8 windows (~2 years) to keep latency bounded — beyond that we return `truncated: true`. **The real fix is server-side:** either a `GET /goals/{id}/items` endpoint, or honoring `goal_id` on the completed-task endpoints. Happy to file that upstream if someone points me at the right repo.
- **Unrelated bug surfaced during investigation:** `td workspace list` fails with `Invalid input: expected string, received null` on `workspaces[].domainName`. Root cause is `z.string().optional()` in `@doist/todoist-sdk`'s `WorkspaceSchema` (`dist/esm/types/sync/resources/workspaces.js:53`) — it should be `.nullish()` since the API returns `null` for workspaces without a custom domain. The CLI's own `toWorkspace` already handles null; the error is thrown inside `api.sync()` before we see the payload. **Not addressed in this PR** — it belongs upstream in `@doist/todoist-sdk`.

## Test plan

- [x] `npm test` — 1314/1314 pass (11 new tests: 7 helper unit tests, 4 command-level integration tests covering default-off behavior, merge, `--limit` capping, and truncation flag)
- [x] `npm run check` — oxlint + oxfmt clean
- [x] `npm run check:skill-sync` — skill doc in sync
- [x] Live smoke: `td goal view <id> --json --full --all` (unchanged) vs `--include-completed` (returns full `totalTaskCount`) against 2 real workspace goals
- [ ] Second reviewer pass on the raw-fetch shape mapping in `src/lib/api/goal-tasks.ts` — I'm reproducing the SDK's `Task` shape by hand because the SDK's `validateTaskArray` strips `goal_ids` (the zod schema doesn't declare it), which would kill our client-side filter. Worth a sanity check that I haven't missed any fields that downstream renderers rely on.

### DIFF
--- a/src/commands/goal.ts
+++ b/src/commands/goal.ts
@@ -1,6 +1,7 @@
 import chalk from 'chalk'
 import { Command } from 'commander'
 import { getApi, type Project } from '../lib/api/core.js'
+import { fetchCompletedTasksForGoal } from '../lib/api/goal-tasks.js'
 import { CollaboratorCache, formatAssignee } from '../lib/collaborators.js'
 import { isAccessible } from '../lib/global-args.js'
 import type { PaginatedViewOptions } from '../lib/options.js'
@@ -81,7 +82,9 @@ async function listGoals(options: PaginatedViewOptions & { workspace?: string })
 
 // ── View ──
 
-async function viewGoal(ref: string, options: PaginatedViewOptions): Promise<void> {
+type ViewGoalOptions = PaginatedViewOptions & { includeCompleted?: boolean }
+
+async function viewGoal(ref: string, options: ViewGoalOptions): Promise<void> {
     const api = await getApi()
     const goal = await resolveGoalRef(api, ref)
 
@@ -91,17 +94,42 @@ async function viewGoal(ref: string, options: PaginatedViewOptions): Promise<voi
           ? parseInt(options.limit, 10)
           : LIMITS.tasks
 
-    const { results: tasks, nextCursor } = await paginate(
+    const { results: openTasks, nextCursor } = await paginate(
         (cursor, limit) => api.getTasks({ goalId: goal.id, cursor: cursor ?? undefined, limit }),
         { limit: targetLimit },
     )
+
+    // The REST `/tasks?goal_id=…` endpoint only returns *open* tasks, so for
+    // goals with completed work those items are silently missing. Opt in to
+    // the (more expensive) completed-task walk with `--include-completed`.
+    let completedTasks: typeof openTasks = []
+    let completedTruncated = false
+    if (options.includeCompleted) {
+        const remaining = targetLimit - openTasks.length
+        if (remaining > 0) {
+            const { tasks: fetched, truncated } = await fetchCompletedTasksForGoal({
+                goalId: goal.id,
+                limit: remaining,
+                expectedCount: goal.progress?.completedTaskCount,
+            })
+            completedTasks = fetched
+            completedTruncated = truncated
+        }
+    }
+
+    const tasks = [...openTasks, ...completedTasks]
 
     if (options.json) {
         const goalJson = JSON.parse(formatJson(goal, 'goal', options.full))
         const tasksJson = JSON.parse(
             formatPaginatedJson({ results: tasks, nextCursor }, 'task', options.full),
         )
-        console.log(JSON.stringify({ goal: goalJson, tasks: tasksJson }, null, 2))
+        const payload: Record<string, unknown> = { goal: goalJson, tasks: tasksJson }
+        if (options.includeCompleted) {
+            payload.completedTaskCount = completedTasks.length
+            payload.completedTasksTruncated = completedTruncated
+        }
+        console.log(JSON.stringify(payload, null, 2))
         return
     }
 
@@ -160,6 +188,13 @@ async function viewGoal(ref: string, options: PaginatedViewOptions): Promise<voi
             }),
         )
         console.log('')
+    }
+    if (completedTruncated) {
+        console.log(
+            chalk.yellow(
+                'Note: some completed tasks are older than the lookback window and were not fetched.',
+            ),
+        )
     }
     console.log(formatNextCursorFooter(nextCursor))
 }
@@ -331,7 +366,11 @@ export function registerGoalCommand(program: Command): void {
     goal.command('view [ref]', { isDefault: true })
         .description('View goal details and linked tasks')
         .option('--limit <n>', 'Limit number of results (default: 300)')
-        .option('--all', 'Fetch all results (no limit)')
+        .option('--all', 'Fetch all open linked tasks (no limit)')
+        .option(
+            '--include-completed',
+            'Also fetch completed linked tasks (slower; makes multiple API calls)',
+        )
         .option('--json', 'Output as JSON')
         .option('--ndjson', 'Output as newline-delimited JSON')
         .option('--full', 'Include all fields in JSON output')
--- a/src/commands/goal.ts
+++ b/src/commands/goal.ts
@@ -104,16 +104,26 @@ async function viewGoal(ref: string, options: ViewGoalOptions): Promise<void> {
     // the (more expensive) completed-task walk with `--include-completed`.
     let completedTasks: typeof openTasks = []
     let completedTruncated = false
+    let completedLimitReached = false
     if (options.includeCompleted) {
         const remaining = targetLimit - openTasks.length
         if (remaining > 0) {
-            const { tasks: fetched, truncated } = await fetchCompletedTasksForGoal({
+            const {
+                tasks: fetched,
+                truncated,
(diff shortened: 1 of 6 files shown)


src/commands/goal.ts
```diff
@@ -161,6 +189,13 @@ async function viewGoal(ref: string, options: PaginatedViewOptions): Promise<voi
         )
         console.log('')
     }
+    if (completedTruncated) {
```
doistbot: [P1] This warning is unreachable when the merged task list is empty. If a goal only has completed linked tasks outside the lookback horizon, `fetchCompletedTasksForGoal()` can return `{ tasks: [], truncated: true }`, but the earlier `tasks.length === 0` branch still prints `No linked tasks.` and returns before this note runs. Please emit the truncation note in that empty-state path as well (or before the early return), otherwise `--include-completed` can produce a false negative.

src/lib/api/goal-tasks.ts
```diff
@@ -0,0 +1,194 @@
...
+            if (pagesFetched > MAX_PAGES_SAFETY) {
+                throw new CliError('API_ERROR', 'Completed-task pagination exceeded safety limit')
+            }
+        } while (cursor && results.length < ceiling)
+
+        windowUntil = windowSince
+        windowsRun += 1
+    }
+
+    const truncated = results.length < ceiling && windowsRun >= MAX_WINDOWS
```
doistbot: [P2] `truncated` only tracks lookback exhaustion, so hitting the caller's `limit` is treated as a complete result. For example, with `--limit 5 --include-completed` on a goal that has 1 open task and 29 completed tasks, this helper returns 4 completed tasks with `truncated: false`, and `goal view` emits `nextCursor: null` / no footer even though 25 linked tasks were omitted. Consider returning a separate `hasMore`/`limitReached` signal when `expectedCount > results.length` (or when a cursor/window remains) so the view layer can surface that the combined result set is incomplete.

doistbot (commented):
This PR introduces the `--include-completed` flag to `td goal view`, allowing users to fetch and client-side filter completed tasks across historical windows to present a complete picture of goal progress. The workaround elegantly circumvents the API's current limitations while safely preserving the default behavior for existing consumers. There are a few edge cases to address regarding truncation signaling, specifically ensuring the truncation warning remains reachable when the merged task list is entirely empty, and differentiating between exhausting the historical lookback versus simply hitting the requested result limit.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F270&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab33279f0-3843-11f1-8978-5ee3c5b39fc8%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab33279f0-3843-11f1-8978-5ee3c5b39fc8&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

lmjabreu:
Addressed both review comments in babebce:

- **[P1]** Moved the truncation warning into a `printCompletedWarnings()` helper that runs on both the empty and non-empty branches, so goals with only out-of-window completed tasks still show the note instead of a bare `No linked tasks.`
- **[P2]** Added a separate `limitReached` signal to `fetchCompletedTasksForGoal` (distinct from `truncated`) and surfaced it as `completedTasksLimitReached` in JSON + a yellow text-mode hint. Also handles the case where the open-task page alone saturates `--limit`, so e.g. `--limit 2` on a goal with 2 open + 2 completed now correctly flags that completed tasks were skipped. New unit + command-level tests cover both paths.

scottlovegrove (changes requested):
The Todoist SDK already has a `getCompletedTasksByCompletionDate()` function that handles all of the API side of things. 

That said, doing it this way feels a little clunky to me. I wonder if there's a change on the backend we can make that will allow for fetching completed tasks specifically for goals (cc @chiara-doist). If we can't do that, then we should at the very least switch to using the actual SDK's functionality for fetching completed tasks.

chiara-doist:
Completed tasks are a can of worms, across Todoist in general, not just limited to goals 🪱

About these; 

> ### Root cause (confirmed against the API)
> 
> - `GET /tasks?goal_id=X` only returns open tasks — no `include_completed`, no alternate flag.
> - `/goals/{id}/items` (GET) → `405 Method Not Allowed`.
> - `/tasks/completed?goal_id=X` and `/tasks/completed/by_completion_date?goal_id=X` both **ignore** the `goal_id` query parameter (verified by inspecting `goal_ids` on the returned items).
> - The task filter DSL has no `@goal:` or `goal_id:` operator.


These are just not supported, because we haven't had a reason to... And I don't think we should introduce very intensive commands / queries before we have a legit use case for them (and since this is for prototype week, I don't recommend us doing so). 

@lmjabreu can you handle it a different way for the purposes of the prototype week?

scottlovegrove:
> @lmjabreu can you handle it a different way for the purposes of the prototype week?

@lmjabreu you should be able to do it how you have been, just tell your agent to use the todoist sdk, rather than handroll the API calls :)

lmjabreu:
Thanks both,

I can use a workaround for the prototyping week 👍

I think we'll need to fetch completed tasks associated with a goal at some point, e.g. when we create a dedicated goal page, rather than use filterist like we do at the moment.

PR #272, opened 2026-04-15
feat(auth): replace per-scope flags with --additional-scopes

scottlovegrove:
## Summary

Addresses [@henningmu's review feedback on #264](https://github.com/Doist/todoist-cli/pull/264#discussion_r3077573810): each new opt-in OAuth scope was getting its own top-level login flag (`--app-management`, `--backups`). That pattern doesn't scale. This PR replaces the two flags with one `--additional-scopes=<list>` (comma-separated), backed by a small registry so future scopes are additive — no new command surface when a new scope lands.

Two commits:

1. **feat(auth): replace per-scope flags with --additional-scopes** (`80b0099`)
   - New `src/lib/oauth-scopes.ts` registry with `ADDITIONAL_SCOPES`, `parseScopesOption`, and `formatScopesHelp` — one place to add future scopes.
   - `td auth login --help` now enumerates every supported scope with a one-line summary.
   - Unknown scopes fail fast with a friendly error listing valid options, *before* any browser opens.
   - `buildReloginCommand` emits the new syntax (preserves `--read-only` separately, bundles opt-ins into one `--additional-scopes=` arg).
   - SKILL.md, command help text (`td apps`, `td backup`), and the `MISSING_SCOPE` remediation strings all speak the new language.
   - `--read-only` stays separate — it swaps the base grant (`data:read_write` → `data:read`), it's not an additional scope.
   - Breaking change: `--app-management` and `--backups` are removed. Persisted `auth_flags` identifiers are unchanged, so existing configs keep working.

2. **feat(auth): personalise MISSING_SCOPE hint from stored auth flags** (`11af719`)
   - When a 403 `AUTH_INSUFFICIENT_TOKEN_SCOPE` fires, `wrapApiError` now reads `auth_flags` via `getAuthMetadata` and threads the required flag through `buildReloginCommand` — the same helper `fetchBackups` already uses for its proactive pre-flight check. Both paths now produce identical advice that preserves whichever scopes the user already holds.
   - Previously the hint was a static string: a read-only user hitting `td backup list` saw `td auth login --additional-scopes=backups`, silently dropping `--read-only`. Now they see `td auth login --read-only --additional-scopes=backups`.

## Examples

```
$ td auth login --help
...
  --read-only                   Authenticate with read-only scope (data:read)
  --additional-scopes <list>    Comma-separated opt-in OAuth scopes (see list below)

Available scopes (comma-separated for --additional-scopes):
  app-management  Manage your registered Todoist developer apps (rotate secrets, edit webhooks).
  backups         List and download your Todoist backups.

Examples:
  td auth login
  td auth login --read-only
  td auth login --additional-scopes=app-management
  td auth login --read-only --additional-scopes=backups
  td auth login --additional-scopes=app-management,backups
```

## Test plan

- [x] \`npm test\` — 1400 tests pass (up from 1368; new \`src/__tests__/oauth-scopes.test.ts\` covers \`parseScopesOption\` edge cases; existing tests updated for the new signatures)
- [x] \`npm run check\` — lint / format / type-check clean
- [x] \`npm run check:skill-sync\` — \`skills/todoist-cli/SKILL.md\` regenerated from \`content.ts\`
- [ ] Live smoke against a real account:
  - [ ] \`td auth login --help\` shows the scopes block
  - [ ] \`td auth login --additional-scopes=app-management\` → authorize URL contains \`dev:app_console\`; persisted \`auth_flags: ['app-management']\`
  - [ ] \`td auth login --read-only --additional-scopes=backups\` → combined scope \`data:read,backups:read\`
  - [ ] \`td auth login --additional-scopes=nonsense\` → friendly error, browser not opened
  - [ ] Space form also accepted: \`td auth login --additional-scopes app-management\` works
  - [ ] On a token missing the scope: \`td apps list\` surfaces a \`MISSING_SCOPE\` hint that preserves any prior flags

### DIFF
--- a/src/commands/auth/index.ts
+++ b/src/commands/auth/index.ts
@@ -13,7 +13,12 @@ export function registerAuthCommand(program: Command): void {
         .option('--read-only', 'Authenticate with read-only scope (data:read)')
         .option(
             '--additional-scopes <list>',
-            'Comma-separated opt-in OAuth scopes (see list below)',
+            'Comma-separated opt-in OAuth scopes (see list below). The flag may be repeated; every occurrence is merged.',
+            // Commander treats this as a scalar by default, so repeated uses
+            // (`--additional-scopes=a --additional-scopes=b`) would silently
+            // drop earlier values. Concatenate into one comma-separated string
+            // and let parseScopesOption split/dedupe/validate as usual.
+            (value: string, prev: string | undefined) => (prev ? `${prev},${value}` : value),
         )
         .addHelpText('after', formatScopesHelp())
         .action(loginWithOAuth)
--- a/src/commands/auth/index.ts
+++ b/src/commands/auth/index.ts
@@ -1,4 +1,5 @@
 import { Command } from 'commander'
+import { formatScopesHelp } from '../../lib/oauth-scopes.js'
 import { loginWithOAuth } from './login.js'
 import { logout } from './logout.js'
 import { showStatus } from './status.js'
@@ -11,13 +12,10 @@ export function registerAuthCommand(program: Command): void {
         .description('Authenticate with Todoist via OAuth')
         .option('--read-only', 'Authenticate with read-only scope (data:read)')
         .option(
-            '--app-management',
-            'Also request the dev:app_console scope (manage your Todoist apps). Combine with --read-only if desired.',
+            '--additional-scopes <list>',
+            'Comma-separated opt-in OAuth scopes (see list below)',
         )
-        .option(
-            '--backups',
-            'Also request the backups:read scope (list/download Todoist backups). Combine with --read-only if desired.',
-        )
+        .addHelpText('after', formatScopesHelp())
         .action(loginWithOAuth)
 
     auth.command('token [token]')
--- a/src/lib/api/core.ts
+++ b/src/lib/api/core.ts
@@ -9,7 +9,9 @@ import {
     WorkspaceProject,
     type DueDate,
 } from '@doist/todoist-sdk'
-import { getApiToken } from '../auth.js'
+import { buildReloginCommand } from '../auth-flags.js'
+import { getApiToken, getAuthMetadata } from '../auth.js'
+import type { AuthFlag } from '../config.js'
 import { CliError } from '../errors.js'
 import { ensureWriteAllowed, isMutatingApiMethod, isMutatingSyncPayload } from '../permissions.js'
 import { getProgressTracker } from '../progress.js'
@@ -157,7 +159,7 @@ function createSpinnerWrappedApi(api: TodoistApi): TodoistApi {
                             (error as Error).message,
                         )
                     }
-                    throw wrapApiError(error, property)
+                    throw await wrapApiError(error, property)
                 }
             }
         },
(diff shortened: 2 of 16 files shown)


src/commands/auth/index.ts
```diff
@@ -11,13 +12,10 @@ export function registerAuthCommand(program: Command): void {
         .description('Authenticate with Todoist via OAuth')
         .option('--read-only', 'Authenticate with read-only scope (data:read)')
         .option(
-            '--app-management',
-            'Also request the dev:app_console scope (manage your Todoist apps). Combine with --read-only if desired.',
-        )
-        .option(
-            '--backups',
-            'Also request the backups:read scope (list/download Todoist backups). Combine with --read-only if desired.',
+            '--additional-scopes <list>',
```
doistbot, line 15: [P2] Repeated `--additional-scopes` flags will silently drop earlier values. Commander treats this as a scalar option, so `td auth login --additional-scopes=app-management --additional-scopes=backups` ends up passing only `backups` to `loginWithOAuth`. That is an easy migration path from the old per-scope flags, and it produces a token missing one requested scope without any warning.

Consider either collecting multiple occurrences and merging them with the comma-split parser, or explicitly rejecting duplicate uses of the option.

src/lib/api/core.ts
```diff
@@ -178,43 +180,47 @@ function isInsufficientScopeError(error: TodoistRequestError): boolean {
...
 }
 
-const SCOPE_REMEDIATION: Record<ScopeGroup, string> = {
-    'app-management':
-        'This command requires the dev:app_console scope. Re-run `td auth login --app-management` (combine with --read-only if desired).',
-    backups:
-        'This command requires the backups:read scope. Re-run `td auth login --backups` (combine with --read-only if desired).',
-    standard:
-        'Re-authenticate with `td auth login` (or `td auth login --read-only`) to refresh your token with the standard scopes.',
+const SCOPE_DESCRIPTIONS: Record<AuthFlag, string> = {
```
doistbot: [P2] This reintroduces a second source of truth for additional scopes next to `ADDITIONAL_SCOPES`. If a future scope is added to the registry and wired into `METHOD_REQUIRED_FLAG` but this table is missed, the remediation string becomes `This command requires the undefined scope` at runtime.

Using `AuthFlag` here also forces the fake `'read-only': ''` entry. Deriving the OAuth scope name from the registry (for example via `oauthScopeFor`) and narrowing this path to additional-scope flags would keep the new registry genuinely additive.

src/lib/oauth-scopes.ts
```diff
@@ -0,0 +1,112 @@
...
+ * de-duplicated list of scope flags. Emits a canonical order matching
+ * `AUTH_FLAG_ORDER` (so persisted `auth_flags` stays stable regardless of
+ * user input order). Throws `CliError('INVALID_OPTIONS')` on empty entries
+ * or unknown scope names.
+ */
+export function parseScopesOption(raw: string): AdditionalScopeFlag[] {
+    const parts = raw
+        .split(',')
+        .map((p) => p.trim())
+        .filter((p) => p.length > 0)
```
doistbot: [P3] Empty list elements are currently accepted even though the function comment says empty entries should fail. Inputs like `--additional-scopes=app-management,` or `app-management,,backups` parse successfully because blank segments are filtered out.

That makes a typo in the list easy to miss. Either reject empty segments before filtering, or update the contract/tests to match the permissive behavior.

doistbot (commented):
This PR cleanly refactors the authentication flow by consolidating individual opt-in scope flags into a single, scalable `--additional-scopes` flag and improving the personalization of missing scope hints. Centralizing these scopes into a single registry significantly improves maintainability and ensures the CLI command surface stays clean as new features are added. A few edge cases should be reviewed before merging, specifically regarding how repeated flag occurrences and empty list segments are handled, as well as ensuring the new registry acts as the strict single source of truth for scope names to prevent accidental runtime errors.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F272&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aa43070d0-38ff-11f1-9a8d-f12c88d72241%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aa43070d0-38ff-11f1-9a8d-f12c88d72241&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

henningmu (commented):
Code looks good, and I don't think we need to iterate further ✅

One potential problem we might need to solve in the future: what if some scopes are read-only and others aren't? Since `--read-only` is applied globally we cannot represent that. Again, let's solve that once we need to 🙌

PR #279, opened 2026-04-20
feat: Add Help Center CLI commands

gnapse:
## Summary

- add a new `td hc` top-level command with `locales`, `search`, and default `view` subcommands for Todoist Help Center articles
- keep article lookup stateless by resolving `view` refs from `id:N`, raw numeric article IDs, or Help Center URLs, and expose supported locales via `td hc locales --json`
- add Help Center fetch/formatting helpers, test coverage, and the corresponding skill/help documentation updates

## Demo

Opening the content of https://www.todoist.com/help/articles/use-task-quick-add-in-todoist-va4Lhpzz

CleanShot 2026-04-20 at 08 48 29

### DIFF
--- a/src/commands/hc/index.ts
+++ b/src/commands/hc/index.ts
@@ -1,0 +1,62 @@
+import { Command } from 'commander'
+import { listHelpCenterLocales } from './locales.js'
+import { searchHelpCenterArticles } from './search.js'
+import { viewHelpCenterArticle } from './view.js'
+
+const LOCALE_OPTION_DESCRIPTION = 'Help Center locale (default: en-us)'
+
+export function registerHelpCenterCommand(program: Command): void {
+    const hc = program
+        .command('hc')
+        .description('Search Todoist Help Center articles')
+        .addHelpText(
+            'after',
+            `
+Examples:
+  td hc locales
+  td hc locales --json
+  td hc search "notifications"
+  td hc search "reminders" --locale pt-br --json
+  td hc view id:360000269065
+  td hc view 360000269065
+  td hc view https://get.todoist.help/hc/en-us/articles/360000269065
+
+Notes:
+  search prints real Help Center article IDs
+  view accepts id:N, raw article IDs, and Help Center URLs`,
+        )
+
+    hc.command('locales')
+        .description('List supported Help Center locales')
+        .option('--json', 'Output as JSON')
+        .action(listHelpCenterLocales)
+
+    const searchCmd = hc
+        .command('search [query]')
+        .description('Search Todoist Help Center articles')
+        .option('--locale <locale>', LOCALE_OPTION_DESCRIPTION)
+        .option('--limit <n>', 'Number of results to return (default: 10, max: 25)')
+        .option('--json', 'Output as JSON')
+        .action((query, options) => {
+            if (!query) {
+                searchCmd.help()
+                return
+            }
+            return searchHelpCenterArticles(query, options)
+        })
+
+    const viewCmd = hc
+        .command('view [ref]', { isDefault: true })
+        .description('View a Help Center article by id:N, raw article ID, or URL')
+        .option('--locale <locale>', LOCALE_OPTION_DESCRIPTION)
+        .option('--browser', 'Open the article in your browser')
+        .option('--json', 'Output as JSON')
+        .option('--html', 'Output the raw HTML article body')
+        .action((ref, options) => {
+            if (!ref) {
+                viewCmd.help()
+                return
+            }
+            return viewHelpCenterArticle(ref, options)
+        })
+}
--- a/src/lib/help-center.ts
+++ b/src/lib/help-center.ts
@@ -1,0 +1,631 @@
+import { CliError } from './errors.js'
+
+const HELP_CENTER_SEARCH_URL = 'https://todoist.zendesk.com/api/v2/help_center/articles/search'
+const HELP_CENTER_ARTICLE_URL_BASE = 'https://todoist.zendesk.com/api/v2/help_center'
+const HELP_CENTER_LOCALES_URL = 'https://todoist.zendesk.com/api/v2/help_center/locales'
+const ACCOUNT_LOCALES_URL = 'https://todoist.zendesk.com/api/v2/locales.json'
+const HELP_CENTER_LOCALE_PATTERN = /^[a-z]{2}(?:-[a-z]{2})?$/
+
+export const DEFAULT_HELP_CENTER_LOCALE = 'en-us'
+export const DEFAULT_HELP_CENTER_LIMIT = 10
+export const MAX_HELP_CENTER_LIMIT = 25
+
+interface RawHelpCenterArticle {
+    id?: number | string
+    title?: string
+    html_url?: string
+    snippet?: string
+    body?: string
+    updated_at?: string
+    label_names?: unknown[]
+    locale?: string
+}
+
+interface RawHelpCenterSearchResponse {
+    results?: RawHelpCenterArticle[]
+}
+
+interface RawHelpCenterArticleResponse {
+    article?: RawHelpCenterArticle
+}
+
+interface RawHelpCenterLocalesResponse {
+    locales?: string[]
+    default_locale?: string
+}
+
+interface RawLocaleDefinition {
+    locale?: string
+    name?: string
+    native_name?: string
+    presentation_name?: string
+    rtl?: boolean
+}
+
+interface RawLocalesResponse {
+    locales?: RawLocaleDefinition[]
+}
+
+export interface HelpCenterSearchResult {
+    id: string
+    title: string
+    htmlUrl: string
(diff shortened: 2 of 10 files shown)


src/lib/help-center.ts
```diff
@@ -0,0 +1,631 @@
...
+async function fetchHelpCenterJson<T>(url: string): Promise<{ response: Response; data: T }> {
+    let response: Response
+    try {
+        response = await fetch(url)
+    } catch (error) {
+        throw new CliError('FETCH_FAILED', `Help Center request failed: ${errorMessage(error)}`)
+    }
+
+    let data: T
+    try {
```
doistbot, line 160: [P2] Handle non-JSON HTTP errors gracefully

If the API returns a non-OK status with an HTML payload (e.g., a 404 page or a 503 proxy error), `response.json()` will fail. This masks the actual HTTP status code and throws a generic `invalid JSON` error before the calling functions can evaluate `response.status === 404` or log the exact status code. Returning an empty data object when `!response.ok` allows the caller to evaluate the status code correctly.

```suggestion
    try {
        data = (await response.json()) as T
    } catch (error) {
        if (!response.ok) {
            return { response, data: {} as T }
        }
        throw new CliError(
            'FETCH_FAILED',
            `Help Center request returned invalid JSON: ${errorMessage(error)}`,
        )
    }
```

src/lib/help-center.ts
```diff
@@ -0,0 +1,631 @@
...
+            'Use values like `en-us`, `es`, `de`, `fr`, `ja`, or `pt-br`.',
+        ])
+    }
+    return normalized
+}
+
+export function parseHelpCenterLimit(limit: number | string | undefined): number {
+    if (limit === undefined) return DEFAULT_HELP_CENTER_LIMIT
+
+    const parsed = typeof limit === 'number' ? limit : Number.parseInt(String(limit).trim(), 10)
```
doistbot, line 326: [P2] `--limit` validation is more permissive than the help text implies. `Number.parseInt()` will accept values like `1.5`, `08foo`, or `10abc` and silently coerce them to `1`, `8`, and `10`, so malformed CLI input changes behavior instead of failing. A stricter parse (`Number()` plus an integer check, or a `^\d+$` guard before parsing) would match the documented contract.

src/lib/help-center.ts
```diff
@@ -0,0 +1,631 @@
...
+}
+
+export interface ResolvedHelpCenterRef {
+    articleId: string
+    locale: string
+    htmlUrl?: string
+    source: 'id' | 'url'
+}
+
+const NAMED_HTML_ENTITIES: Record<string, string> = {
```
doistbot, line 92: [P2] The custom entity decoder only handles a small hard-coded subset of named HTML entities. Help Center snippets and bodies in non-English locales can legitimately contain entities like `&ntilde;`, `&eacute;`, or `&ouml;`, which will currently leak through literally in `search` and `view` output. Using a full HTML entity decoder here would avoid broken localized rendering.

doistbot (commented):
This PR introduces the new `td hc` command suite to seamlessly search and view Todoist Help Center articles directly from the terminal. It's a great enhancement that cleanly integrates external documentation into the CLI workflow. A few refinements are noted to handle edge cases more robustly, including gracefully parsing non-JSON HTTP error responses, enforcing stricter validation for limit flags, fully documenting the new commands in the agent skill content, and improving markdown accuracy by resolving relative links, expanding HTML entity decoding, and clarifying the locale option description.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F279&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Af8407a60-3cad-11f1-927d-3cec3495515e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Af8407a60-3cad-11f1-927d-3cec3495515e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/help-center.ts
```diff
@@ -0,0 +1,631 @@
...
+    let rendered = fragment
+        .replace(/<!--[\s\S]*?-->/g, '')
+        .replace(/<(script|style)[^>]*>[\s\S]*?<\/\1>/gi, '')
+        .replace(/<br\s*\/?>/gi, '\n')
+
+    rendered = rendered.replace(
+        /<a\b[^>]*href=(['"])(.*?)\1[^>]*>([\s\S]*?)<\/a>/gi,
+        (_match, _quote: string, href: string, inner: string) => {
+            const label = renderInlineHtml(inner).replace(/\n+/g, ' ').trim()
+            const cleanHref = decodeHtmlEntities(href)
```
doistbot, line 182: [P2] Relative links are emitted unchanged in the generated markdown. Zendesk article bodies commonly use `href="/hc/..."`; with the current code those become markdown links to `/hc/...`, which aren't actionable from terminal output. Resolving relative `href`s against the Help Center origin before formatting would keep intra-article links usable.

src/lib/skills/content.ts
```diff
@@ -209,6 +210,14 @@ td reminder location get id:456
 
 \`td attachment view\` prints text attachments directly and encodes binary content as base64. Use \`--json\` for metadata plus content.
 
+### Help Center
+\`\`\`bash
+td hc
+td hc --help
+\`\`\`
+
+\`td hc\` queries the Todoist online Help Center. Run \`td hc --help\` for locale discovery, article search, and article viewing details.
```
doistbot, line 219: [P2] This leaves the skill source-of-truth underspecified for the new command surface. The repo guidance requires `SKILL_CONTENT` to reflect new subcommands and options, but this section only tells agents to run `td hc --help` and omits the actual `locales` / `search` / `view` behaviors and flags like `--locale`, `--limit`, `--json`, `--html`, and `--browser`.

src/commands/hc/index.ts
```diff
@@ -0,0 +1,62 @@
+import { Command } from 'commander'
+import { listHelpCenterLocales } from './locales.js'
+import { searchHelpCenterArticles } from './search.js'
+import { viewHelpCenterArticle } from './view.js'
+
+const LOCALE_OPTION_DESCRIPTION = 'Help Center locale (default: en-us)'
```
doistbot, line 6: [P3] This shared option description is misleading for `td hc view`. When the ref is a Help Center URL, the command defaults to the URL's locale rather than `en-us`, so the current text is only accurate for `search`. A view-specific description like "Override Help Center locale" would better match the implementation.

src/lib/help-center.ts
```diff
@@ -0,0 +1,631 @@
...
+            htmlUrl: getArticleFallbackUrl(apiMatch[1].toLowerCase(), apiMatch[2]),
+        }
+    }
+
+    return null
+}
+
+export async function searchHelpCenter(
+    query: string,
+    options: { locale?: string; limit?: number | string } = {},
```
scottlovegrove, line 482: Should we have a default for the limit?

scottlovegrove (approved):
What about having something like `td hc locale --set-default es-es` so that then the locale could be omitted from any subsequent calls. Otherwise you have to remember to set the correct locale every time.

src/lib/skills/content.ts
gnapse: Intentional. This command is not that important to put all its sub-commands in an agent's context upfront. If the agent thinks it needs the help center, it can query `--help` for more info.

scottlovegrove: I almost left a similar note, then figured help centre wasn't _that_ important 😄

src/lib/help-center.ts
gnapse: Addressed partially, only resolving the absolute paths to resolve them to `https://get.todoist.help`.

src/lib/help-center.ts
gnapse: A bit nitpicky. We can fix it if the issue arises in practice.

PR #283, opened 2026-04-21
feat(apps): add update subcommand with --add/--remove-oauth-redirect

scottlovegrove:
## Summary
- Adds `td apps update <ref>` with mutually-exclusive `--add-oauth-redirect <url>` / `--remove-oauth-redirect <url>` flags so developers can manage an app's OAuth redirect URIs from the CLI.
- URI validation mirrors `todoist-web`'s app editor (`src/settings/integrations/app-management/app-editor/helpers.ts`): https, http(s)://localhost[:port][/path], or a custom scheme; rejects `javascript`/`data`/`file`/`vbscript`/`ftp`. Validation runs **before** any API call.
- Always fetches the app first via the existing `resolveAppRef`, parses the backend's three accepted shapes (JSON array, plain string, legacy comma-separated), and re-serializes using todoist-web's same rule (single → plain string, multiple → JSON array). Clears `oauthRedirectUri` to `null` when the last URI is removed.
- Add rejects already-present URIs with `ALREADY_EXISTS`. Remove with a not-present URI exits 0 and makes no API call. Remove with `--yes` mutates; without `--yes` prints a "Would remove… Use --yes to confirm." preview, matching `td task delete`. Both paths support `--dry-run` and `--json`.

## Test plan
- [x] `npm run type-check`
- [x] `npm test` (1463/1463 — 10 new tests covering conflicting flags, missing flags, invalid URI, add to empty/non-empty, dedupe, dry-run, remove not-present, remove without `--yes`, remove with `--yes` clearing to null, remove with `--yes` keeping remaining URIs)
- [x] `npm run sync:skill` — `skills/todoist-cli/SKILL.md` regenerated and committed
- [ ] Manual: run against a real app to confirm the backend round-trips the JSON-array form correctly


scottlovegrove:
@doistbot /review

src/commands/apps/helpers.ts
```diff
@@ -0,0 +1,70 @@
+// URL validation + serialization for app OAuth redirect URIs.
+//
+// Mirrored from todoist-web so the CLI accepts the same set of URIs the web
+// app editor does. Keep in sync with:
+// src/settings/integrations/app-management/app-editor/helpers.ts in todoist-web.
+
+const urlRegex = /^https:\/\/[^:/]+\.[a-zA-Z]{2,}(\/[^:]+)?$/
+const localhostRegex = /^https?:\/\/(localhost|127\.0\.0\.1)(:\d+)?(\/.*)?$/
+const customSchemeRegex = /^[a-zA-Z][a-zA-Z0-9+\-.]*:\/\/.+$/
+const dangerousSchemes = new Set(['javascript', 'data', 'file', 'vbscript', 'http', 'https', 'ftp'])
```
scottlovegrove: We should still allow `http://localhost`

src/commands/apps/helpers.ts
```diff
@@ -0,0 +1,70 @@
+// URL validation + serialization for app OAuth redirect URIs.
+//
+// Mirrored from todoist-web so the CLI accepts the same set of URIs the web
+// app editor does. Keep in sync with:
+// src/settings/integrations/app-management/app-editor/helpers.ts in todoist-web.
```
scottlovegrove: Do not link to private code

src/commands/apps/helpers.ts
scottlovegrove, line 1: We should add some tests for this file. Especially so that `http://example.com` fails, but `http://localhost` passes.

src/commands/apps/helpers.ts
scottlovegrove: Also, why is `https` listed as a dangerous scheme?

src/commands/apps/update.ts
```diff
@@ -0,0 +1,121 @@
...
+        if (!isQuiet()) {
+            console.log(`Added OAuth redirect URI to ${app.displayName} (id:${app.id}): ${add}`)
+        }
+        return
+    }
+
+    // --remove-oauth-redirect path
+    const toRemove = remove as string
+
+    if (!current.includes(toRemove)) {
```
doistbot, line 89: [P1] JSON output broken on no-op removal

When `--json` is passed but the URI to remove is not present on the app, the command exits `0` but outputs a human-readable string. This breaks machine parsing for scripts that depend on the `--json` output format. If no changes are needed, the command should output the existing `app` object as JSON.

```typescript
    if (!current.includes(toRemove)) {
        if (options.json) {
            console.log(JSON.stringify(app, null, 2))
            return
        }
        console.log(
            `"${toRemove}" is not an OAuth redirect URI for "${app.displayName}" — nothing to remove.`,
        )
        return
    }
```

src/commands/apps/update.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    if (options.dryRun) {
+        printDryRun('remove OAuth redirect URI', {
+            App: `${app.displayName} (id:${app.id})`,
+            URI: toRemove,
+        })
+        return
+    }
+
+    if (!options.yes) {
```
doistbot, line 110: [P2] Incompatible preview output with `--json` flag

When `td apps update` is called with `--remove-oauth-redirect` and `--json` but without `--yes`, the command outputs a human-readable preview text, breaking the expected JSON format. If `--json` is used (implying programmatic usage), the command should return a structured error instructing the caller to provide `--yes`.

```typescript
    if (!options.yes) {
        if (options.json) {
            throw new CliError(
                'CONFIRMATION_REQUIRED',
                'Confirmation required to remove OAuth redirect URI.',
                ['Pass --yes to confirm.'],
            )
        }
        console.log(
            `Would remove OAuth redirect URI from ${app.displayName} (id:${app.id}): ${toRemove}`,
        )
        console.log('Use --yes to confirm.')
        return
    }
```

src/commands/apps/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+                return parsed.filter(
+                    (item): item is string => typeof item === 'string' && item.length > 0,
+                )
+            }
+        } catch {
+            // Fall through to comma-separated parsing.
+        }
+    }
+
+    return trimmed
```
doistbot: [P2] Legacy comma-separated fallback splits valid single URLs containing commas

If the backend returns a single valid URL that happens to contain a query parameter with commas (e.g., `https://example.com/cb?data=1,2,3`), this logic will incorrectly split it into `['https://example.com/cb?data=1', '2', '3']` because it doesn't start with `[`. This breaks the user's valid URL when they later add or remove another URI.

Consider checking if the entire string is a valid single URL before falling back to comma splitting.

```typescript
    if (validateRedirectUri(trimmed)) {
        return [trimmed]
    }

    return trimmed
        .split(',')
        .map((u) => u.trim())
        .filter((u) => u.length > 0)
```

src/commands/apps/update.ts
```diff
@@ -0,0 +1,121 @@
...
+        )
+    }
+
+    if (!add && !remove) {
+        throw new CliError('NO_CHANGES', 'No changes specified.', [
+            'Use --add-oauth-redirect <url> or --remove-oauth-redirect <url>.',
+        ])
+    }
+
+    if (add !== undefined && !validateRedirectUri(add)) {
```
doistbot: [P2] `--remove-oauth-redirect` skips the preflight URI validation. This guard only checks `add`, so `td apps update ... --remove-oauth-redirect javascript://alert(1)` still goes on to `getApi()` / `resolveAppRef()` and may even exit as a successful no-op instead of returning `INVALID_URL`.

If the command is meant to validate redirect URIs before any API call, the remove path needs the same check before loading the app.

src/commands/apps/update.ts
```diff
@@ -0,0 +1,121 @@
...
+            })
+            return
+        }
+
+        const updated = await api.updateApp(app.id, {
+            oauthRedirectUri: serializeOAuthRedirectUris(next),
+        })
+
+        if (options.json) {
+            console.log(JSON.stringify(updated, null, 2))
```
doistbot: [P2] This new update subcommand is bypassing the shared JSON formatter. The rest of the add/create/update commands go through `formatJson(...)`, but this path is hand-rolling `JSON.stringify(updated)` instead, which makes apps update the odd one out and makes future JSON-shape changes easier to miss.

Please add app support to the shared formatter and use it here (and in the remove branch below) so `--json` stays consistent with the repository's existing mutating-command pattern.

doistbot (commented):
This PR introduces a helpful `td apps update` subcommand to manage OAuth redirect URIs directly from the CLI. The implementation nicely aligns with the web client's validation and serialization rules, providing a solid foundation for developer app management. There are just a few edge cases to smooth out regarding the `--json` flag's behavior during no-ops or unconfirmed removals, utilizing the shared `formatJson` helper, applying preflight validation to removals, and preventing the legacy fallback parser from splitting valid URLs that contain commas.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F283&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab81ce880-3da0-11f1-8fda-4caecb4f6fff%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab81ce880-3da0-11f1-8fda-4caecb4f6fff&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/apps/helpers.ts
scottlovegrove: is still allowed — `localhostRegex` is tested *before* the custom-scheme branch, so it matches first and the `dangerousSchemes` set never sees it. Covered by the new `validateRedirectUri` tests (`http://localhost`, `http://localhost:3000`, `http://localhost/callback` all pass; `http://example.com` fails).

src/commands/apps/helpers.ts
scottlovegrove: Dropped the cross-repo path reference in 9ad9f3b.

src/commands/apps/helpers.ts
scottlovegrove: Added `src/commands/apps/helpers.test.ts` in 9ad9f3b. Covers the `http://example.com`-fails / `http://localhost`-passes asymmetry you called out, plus the parse/serialize contract and the new comma-in-query-string preservation doistbot flagged.

src/commands/apps/helpers.ts
scottlovegrove: Good question — I first removed them as dead code, but the full-suite tests regressed immediately. `customSchemeRegex` is permissive enough that a malformed multi-URL string like `https://a.com/cb,https://b.com/cb` will match it (scheme `https`, body `//a.com/cb,https://b.com/cb`), so well-formed https URLs are only accepted earlier by `urlRegex` and anything reaching this branch with an http/https scheme is malformed. Kept them in and added a comment in 9ad9f3b explaining why they're a backstop rather than redundant.

src/commands/apps/update.ts
scottlovegrove: Fixed in 9ad9f3b — the no-op remove path now calls `formatJson(app, 'app')` when `--json` is set, so scripts get the unchanged app back in a parseable shape. Test: `remove no-op with --json outputs the unchanged app as JSON`.

src/commands/apps/update.ts
scottlovegrove: Fixed in 9ad9f3b. `--remove --json` without `--yes` now throws `CONFIRMATION_REQUIRED` (added to `ErrorCode`) with a hint pointing at `--yes`, so JSON callers get a structured error instead of a human preview they can't parse. Test: `remove without --yes and with --json throws CONFIRMATION_REQUIRED instead of printing a preview`.

src/commands/apps/helpers.ts
scottlovegrove: Fixed in 9ad9f3b. Before falling back to comma-splitting we now check whether the whole trimmed string is itself a valid redirect URI via `validateRedirectUri` — if so it's treated as a single entry and the query-string commas survive. Regression test: `preserves a single valid URL whose query string contains commas`.

src/commands/apps/update.ts
scottlovegrove: Fixed in 9ad9f3b. The preflight now validates both `add` and `remove`, and rejects `javascript://alert(1)` etc. before any API call on either path. Test: `rejects invalid redirect URIs on --remove before any API call`.

src/commands/apps/update.ts
scottlovegrove: Fixed in 9ad9f3b. Added `'app'` to `EntityType` in `src/lib/output.ts` with an essential-fields list (id, clientId, displayName, status, serviceUrl, oauthRedirectUri, appTokenScopes, userCount), and both the add and remove success paths now go through `formatJson(updated, 'app')`. Test: `add with --json outputs only the essential app fields via the shared formatter`.

src/commands/apps/update.ts
```diff
@@ -0,0 +1,142 @@
...
+        )
+    }
+
+    if (!add && !remove) {
+        throw new CliError('NO_CHANGES', 'No changes specified.', [
+            'Use --add-oauth-redirect <url> or --remove-oauth-redirect <url>.',
+        ])
+    }
+
+    // Both flags get the same preflight validation — keeps the API call out
```
doistbot: [P2] Remove validation for `--remove-oauth-redirect`.

While it makes sense to validate new URIs being added to prevent saving malformed data, validating the URI being removed prevents users from cleaning up malformed URIs that might already be present on their app (e.g. due to legacy data or past bugs). Since removing an invalid URI poses no risk, and the command already gives clear feedback if the string is not found (`"<url>" is not an OAuth redirect URI... nothing to remove`), we should allow any string to be passed to `--remove-oauth-redirect`.

*(Note: If you apply this suggestion, you'll also need to drop the corresponding `rejects invalid redirect URIs on --remove` test in `apps.test.ts`.)*

```suggestion
    // Validate added URIs to prevent saving malformed data. We intentionally do not
    // validate removed URIs so users can clean up legacy malformed URIs if any exist.
    if (add !== undefined && !validateRedirectUri(add)) throw invalidUri(add)
```

src/commands/apps/helpers.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (localhostRegex.test(url)) {
+        try {
+            new URL(url)
+            return true
+        } catch {
+            return false
+        }
+    }
+    if (validateIntegrationUrl(url)) return true
+    if (customSchemeRegex.test(url)) {
```
doistbot, line 35: [P1] Legacy comma-separated custom-scheme redirects get misparsed here. Because this branch accepts any `scheme://...` string, `validateRedirectUri('myapp://cb,otherapp://cb')` returns true; then `parseOAuthRedirectUris()` treats the whole legacy value as one URI instead of splitting it. For apps already stored in the old comma-separated format, `--remove-oauth-redirect otherapp://cb` will no-op and `--add` will write the malformed combined string back. Tighten custom-scheme validation here, or explicitly reject commas before using the single-URI fast path.

src/commands/apps/update.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (options.dryRun) {
+            printDryRun('add OAuth redirect URI', {
+                App: `${app.displayName} (id:${app.id})`,
+                URI: add,
+            })
+            return
+        }
+
+        const updated = await api.updateApp(app.id, {
+            oauthRedirectUri: serializeOAuthRedirectUris(next),
```
doistbot, line 73: [P2] This now writes the backend's JSON-array encoding for multi-URI apps, but the existing `td apps view` plain renderer still prints `oauthRedirectUri` verbatim. After the first `--add` on an app that already has a redirect, users will see a raw `[...]` blob rather than readable URIs. Since this subcommand is what makes multi-URI apps common, it would be better to update the human-facing render path in the same change to parse and format multiple redirects cleanly.

src/lib/skills/content.ts
```diff
@@ -260,6 +262,8 @@ The \`apps\` command surface manages the user's registered Todoist developer app
...
 
+\`td apps update <ref> --add-oauth-redirect <url>\` appends an OAuth redirect URI to the app, and \`--remove-oauth-redirect <url>\` takes one off (requires \`--yes\` to actually mutate, like \`td task delete\`). The two flags are mutually exclusive — pass one at a time. The URI is validated before any API call: \`https://<host>\`, \`http(s)://localhost[:port][/path]\`, or a custom-scheme URI (e.g. \`myapp://callback\`) are accepted; \`javascript\`, \`data\`, \`file\`, \`vbscript\`, and \`ftp\` schemes are rejected. Adding a URI already set on the app fails with \`ALREADY_EXISTS\`; removing a URI that isn't on the app exits 0 with a message and makes no API call. Supports \`--dry-run\` and \`--json\`.
```
doistbot: [P3] The docs here only mention `http(s)://localhost`, but the new validator and tests also accept `http(s)://127.0.0.1`. Because this file is the source of truth for agent help, that drift will steer users away from a supported local callback form. Either document the loopback-IP variant here (and in the `INVALID_URL` hint) or stop accepting it in validation.

doistbot (commented):
This PR introduces the `td apps update` subcommand, providing developers with a convenient way to manage OAuth redirect URIs directly from the CLI. The robust handling of different backend data shapes and the thoughtful inclusion of dry-run and confirmation prompts make this a high-quality addition. A few adjustments are needed to polish the implementation, specifically around removing validation for deleted URIs to allow cleaning up legacy data, tightening custom-scheme validation to properly parse comma-separated values, updating `td apps view` to cleanly render the new JSON-array format, and aligning the skill documentation with the supported loopback IP formats.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F283&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1d928390-3da7-11f1-81ba-3e7aa22ef80f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1d928390-3da7-11f1-81ba-3e7aa22ef80f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

PR #288, opened 2026-04-22
feat(comment): steer agents away from downloading image attachments

rmartins90:
## Overview

When an agent (e.g. Claude Code) fetches a comment and then `curl`s the image `fileUrl` to `Read` it, the Claude API can return `400 invalid_request_error: Could not process image`. Once that happens the image stays pinned in conversation context, and every retry hits the same error — the session gets stuck until you `/compact` or start over.

This PR nudges agents toward the safe path: use `td attachment view` (base64 text, doesn't go through the vision pipeline) instead of raw download + `Read`.

## Changes

- `td comment view`: when an attachment's `fileType` starts with `image/`, print a dim `Hint` line under the URL pointing at `td attachment view` and warning against direct download.
- `SKILL.md` / `content.ts`: new `Core Patterns` bullet explaining the failure mode, plus an expanded note on the `td attachment view` paragraph. Agents that load the skill now get the guidance upfront.
- Test covers both the new hint on image attachments and a regression check that PDF attachments still render without it.
- Regenerated `skills/todoist-cli/SKILL.md` via `npm run sync:skill` — `check:skill-sync` passes.

## Test plan

- [ ] `npm run check` passes (lint + format)
- [ ] `npm test -- src/commands/comment` passes
- [ ] `npm run check:skill-sync` passes
- [ ] Manually run `td comment view id:<comment-with-image>` and confirm the hint appears below the URL, in dim styling
- [ ] Manually run `td comment view id:<comment-with-pdf>` and confirm no hint appears

### DIFF
--- a/src/commands/comment/view.ts
+++ b/src/commands/comment/view.ts
@@ -40,5 +40,12 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
         if (att.fileSize) console.log(`  Size:  ${formatFileSize(att.fileSize)}`)
         if (att.fileType) console.log(`  Type:  ${att.fileType}`)
         if (att.fileUrl) console.log(`  URL:   ${att.fileUrl}`)
+        if (att.fileType?.startsWith('image/')) {
+            console.log(
+                chalk.dim(
+                    "  Hint:  image attachment — fetch via 'td attachment view <url>' if needed; do not download and Read directly",
+                ),
+            )
+        }
     }
 }
--- a/src/commands/comment/view.ts
+++ b/src/commands/comment/view.ts
@@ -6,6 +6,25 @@ import { formatFileSize, formatJson } from '../../lib/output.js'
 import { lenientIdRef } from '../../lib/refs.js'
 import { commentUrl, projectCommentUrl } from '../../lib/urls.js'
 
+const IMAGE_EXT_PATTERN = /\.(png|jpe?g|gif|webp|bmp|svg|heic|heif|avif|tiff?)(?:\?|#|$)/i
+
+type AttachmentLike = {
+    fileName?: string | null
+    fileType?: string | null
+    fileUrl?: string | null
+}
+
+function isImageAttachment(att: AttachmentLike | null | undefined): boolean {
+    if (!att) return false
+    if (att.fileType?.startsWith('image/')) return true
+    if (att.fileName && IMAGE_EXT_PATTERN.test(att.fileName)) return true
+    if (att.fileUrl && IMAGE_EXT_PATTERN.test(att.fileUrl)) return true
+    return false
+}
+
+const IMAGE_HINT =
+    "image attachment — fetch via 'td attachment view <url>' if needed; do not download and Read directly"
+
 export async function viewComment(commentId: string, options: ViewOptions): Promise<void> {
     const api = await getApi()
     const id = lenientIdRef(commentId, 'comment')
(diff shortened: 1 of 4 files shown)


src/commands/comment/view.ts
```diff
@@ -40,5 +40,12 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
         if (att.fileSize) console.log(`  Size:  ${formatFileSize(att.fileSize)}`)
         if (att.fileType) console.log(`  Type:  ${att.fileType}`)
         if (att.fileUrl) console.log(`  URL:   ${att.fileUrl}`)
+        if (att.fileType?.startsWith('image/')) {
```
doistbot: [P2] This guard relies only on `fileType`, even though the surrounding code already treats attachment metadata as partial. If `getComment()` returns an image without `fileType` (or with a generic type), the safety hint never appears; if `fileType` is present but `fileUrl` is missing, the output now suggests `td attachment view <url>` even though there is no URL to use. Gate the hint on `fileUrl` as well, and use a best-effort image check from `fileName`/`fileUrl` when `fileType` is absent.

doistbot (commented):
This PR introduces a helpful hint in `td comment view` and updates the skill documentation to guide AI agents toward using `td attachment view` for images, preventing a known API failure. These changes thoughtfully improve the reliability of agent sessions by steering them toward the safer base64 path. A small refinement is needed to ensure the hint only displays when a `fileUrl` is actually present, and to incorporate a fallback check for image file extensions in case the `fileType` metadata is missing.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F288&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9abb8220-3e58-11f1-83cf-a6208da53693%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9abb8220-3e58-11f1-83cf-a6208da53693&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/comment/view.ts
```diff
@@ -40,5 +40,12 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
...
+            console.log(
+                chalk.dim(
+                    "  Hint:  image attachment — fetch via 'td attachment view <url>' if needed; do not download and Read directly",
+                ),
+            )
+        }
```
scottlovegrove: I suppose there's no way we can have something like this if `--json` has been provided, otherwise they will just the raw attachment data with no other hint (other than the skill) that they should use the view attachment command.

scottlovegrove (approved):

src/commands/comment/view.ts
rmartins90: Fixed. Hint now also requires `fileUrl`, and falls back to `fileName`/`fileUrl` extension when `fileType` is missing.

src/commands/comment/view.ts
rmartins90: Yeah, good call. Writing the hint to stderr when `--json` is set - stdout stays clean for anything piping it, but the agent still sees it in the tool output. Bumped the skill note too.

PR #289, opened 2026-04-22
feat(task): add --no-due flag to clear due date

jameson-m:
Adds `--no-due` to `td task update` to clear a task's due date, mirroring the existing `--no-deadline` flag. Sends `dueString: null` (the SDK's documented clear mechanism).


scottlovegrove (approved):
Thanks for adding this

PR #294, opened 2026-04-23
feat(workspace): add `td workspace use <ref>` to persist a default workspace

scottlovegrove:
## Summary
- Adds `td workspace use <ref>` to store a default workspace id under `workspace.defaultWorkspace` in the user config, and `td workspace use --clear` to remove it.
- When a workspace ref is omitted from `td workspace view/update/delete/projects/users/user-tasks/activity/insights` and `td folder list/create`, the command falls back to the stored default. `resolveWorkspaceRef` absorbs the fallback behind its existing signature, widened to `string | undefined`.
- Commands where "no workspace" has distinct semantics (`td template create`, `td project move --to-workspace`, `td project archived-count --workspace`) are deliberately unchanged.
- `td config view` now shows the configured default under a Workspace section, and `td doctor`'s validator understands the new nested config block.

## Test plan
- [x] `npm run type-check`
- [x] `npm test` (7 new tests cover: set, `id:` refs, `--clear`, `--clear` no-op, default fallback, explicit ref overrides default, `WORKSPACE_REQUIRED` error when neither is set)
- [x] `npm run sync:skill` — committed regenerated `skills/todoist-cli/SKILL.md`
- [ ] Manual smoke: `td workspace use "My WS"` → `td workspace projects` (uses default) → `td workspace use --clear`

### DIFF
--- a/src/commands/folder/helpers.ts
+++ b/src/commands/folder/helpers.ts
@@ -1,19 +1,15 @@
 import type { Folder } from '@doist/todoist-sdk'
 import { getApi } from '../../lib/api/core.js'
 import { fetchWorkspaces, type Workspace } from '../../lib/api/workspaces.js'
-import { readConfig } from '../../lib/config.js'
 import { CliError } from '../../lib/errors.js'
-import { resolveFolderRef, resolveWorkspaceRef } from '../../lib/refs.js'
+import { readDefaultWorkspaceRef, resolveFolderRef, resolveWorkspaceRef } from '../../lib/refs.js'
 
 export async function resolveWorkspaceForFolder(options: {
     workspace?: string
 }): Promise<Workspace> {
-    if (options.workspace) {
-        return resolveWorkspaceRef(options.workspace)
-    }
-    const savedId = (await readConfig()).workspace?.defaultWorkspace
-    if (savedId) {
-        return resolveWorkspaceRef(`id:${savedId}`)
+    const ref = options.workspace ?? (await readDefaultWorkspaceRef())
+    if (ref) {
+        return resolveWorkspaceRef(ref)
     }
     const workspaces = await fetchWorkspaces()
     if (workspaces.length === 0) {
--- a/src/commands/folder/helpers.ts
+++ b/src/commands/folder/helpers.ts
@@ -1,6 +1,7 @@
 import type { Folder } from '@doist/todoist-sdk'
 import { getApi } from '../../lib/api/core.js'
 import { fetchWorkspaces, type Workspace } from '../../lib/api/workspaces.js'
+import { readConfig } from '../../lib/config.js'
 import { CliError } from '../../lib/errors.js'
 import { resolveFolderRef, resolveWorkspaceRef } from '../../lib/refs.js'
 
@@ -9,6 +10,10 @@ export async function resolveWorkspaceForFolder(options: {
 }): Promise<Workspace> {
     if (options.workspace) {
         return resolveWorkspaceRef(options.workspace)
+    }
+    const savedId = (await readConfig()).workspace?.defaultWorkspace
+    if (savedId) {
+        return resolveWorkspaceRef(`id:${savedId}`)
     }
     const workspaces = await fetchWorkspaces()
     if (workspaces.length === 0) {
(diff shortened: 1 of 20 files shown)


src/commands/workspace/index.ts
```diff
@@ -248,15 +215,22 @@ export function registerWorkspaceCommand(program: Command): void {
...
+                return showWorkspaceInsights(refArg || options.workspace, options)
             },
         )
 
+    const useCmd = workspace
+        .command('use [ref]')
+        .description('Set the default workspace used when [ref] is omitted from other commands')
+        .option('--clear', 'Remove the saved default workspace')
+        .action((ref: string | undefined, options: UseWorkspaceOptions) => {
+            if (!ref && !options.clear) {
```
doistbot: [P2] Overlapping validation: Because this check calls `useCmd.help()` and exits, the custom `CliError` inside `useWorkspace` (which contains helpful actionable hints) will never be reached when the user runs the command without arguments. Consider removing this validation block and directly calling `useWorkspace(ref, options)` so the handler's intended error message is shown.

src/commands/workspace/use.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export interface UseWorkspaceOptions {
+    clear?: boolean
+}
+
+export async function useWorkspace(
+    ref: string | undefined,
+    options: UseWorkspaceOptions = {},
+): Promise<void> {
+    if (options.clear) {
```
doistbot, line 22: [P3] Conflicting arguments: If a user explicitly provides both a reference and the clear flag (e.g., `td workspace use "Acme" --clear`), the command silently ignores the reference and clears the default workspace. It would be safer to throw a `CliError` explicitly if `options.clear && ref` are both present.

src/commands/folder/helpers.ts
```diff
@@ -10,6 +11,10 @@ export async function resolveWorkspaceForFolder(options: {
     if (options.workspace) {
         return resolveWorkspaceRef(options.workspace)
     }
+    const savedId = (await readConfig()).workspace?.defaultWorkspace
```
doistbot: [P2] This re-implements the new default-workspace lookup that `resolveWorkspaceRef()` now owns by reading `config.workspace.defaultWorkspace` and rebuilding the `id:` ref locally. Keeping that logic in two places is easy to let drift if the stored format or normalization changes; extracting a shared accessor and reusing it here would keep the folder-specific single-workspace fallback without duplicating config semantics.

doistbot (commented):
This PR introduces the ability to set and clear a default workspace via `td workspace use`, seamlessly integrating it as a fallback for workspace and folder commands while updating configuration tools. The implementation thoughtfully handles the fallback for most commands and preserves explicit "no workspace" semantics where needed, which will make the CLI experience much smoother for users. A few minor adjustments are noted regarding the command validation logic, specifically around handling conflicting arguments, ensuring custom error messages aren't masked by early exits, and deduplicating the configuration lookup for folder commands.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F294&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1b4c4ae0-3f59-11f1-906c-2c5fef66edf9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1b4c4ae0-3f59-11f1-906c-2c5fef66edf9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

rmartins90 (commented):
🚀

PR #295, opened 2026-04-24
feat(skill): prompt when install agent is omitted

gnapse:
This updates `td skill install` so running it without an agent opens an interactive checklist in TTY sessions, while non-interactive runs still show help.

The checklist keeps the first viable target preselected for a fast Enter-to-install flow, allows selecting multiple install targets in one run, and still ranks detected agent roots first. In global mode it only shows install targets that can actually succeed on that machine; in local mode it also shows the other supported agents.

The install flow now preflights the full selection before writing anything, so a bad target does not leave partial installs behind. It also keeps clean cancellation via `Ctrl+C`, updates the `--help` text and README guidance, and switches the prompt implementation to `@inquirer/checkbox`.

## Test

```shell
npm run build
npm run type-check
npm test
npx oxlint src
```

## Demo

https://github.com/user-attachments/assets/957890c6-ce2a-45c6-b196-1f0ead347d6f

### DIFF
--- a/src/commands/skill/install.ts
+++ b/src/commands/skill/install.ts
@@ -53,17 +53,18 @@ async function getInstallChoices(local: boolean): Promise<InstallChoice[]> {
     const undetectedUniversal = choices.filter(
         (choice) => !choice.detected && choice.agent === 'universal',
     )
+    const unavailableNonUniversal = local ? undetectedNonUniversal : []
 
     if (detectedNonUniversal.length > 0) {
         return [
             ...detectedNonUniversal,
             ...detectedUniversal,
-            ...undetectedNonUniversal,
             ...undetectedUniversal,
+            ...unavailableNonUniversal,
         ]
     }
 
-    return [...detectedUniversal, ...undetectedUniversal, ...undetectedNonUniversal]
+    return [...detectedUniversal, ...undetectedUniversal, ...unavailableNonUniversal]
 }
 
 async function promptForAgent(options: InstallOptions): Promise<string> {
@@ -116,8 +117,17 @@ async function promptForAgent(options: InstallOptions): Promise<string> {
                 return
             }
 
-            const selection = Number.parseInt(input, 10)
-            if (Number.isNaN(selection) || selection < 1 || selection > choices.length) {
+            if (!/^\d+$/.test(input)) {
+                reject(
+                    new CliError('INVALID_SELECTION', `Invalid selection: ${input}`, [
+                        `Enter a number from 1 to ${choices.length}, or press Enter for ${defaultChoice.agent}. Press Ctrl+C to cancel.`,
+                    ]),
+                )
+                return
+            }
+
+            const selection = Number(input)
+            if (!Number.isInteger(selection) || selection < 1 || selection > choices.length) {
                 reject(
                     new CliError('INVALID_SELECTION', `Invalid selection: ${input}`, [
                         `Enter a number from 1 to ${choices.length}, or press Enter for ${defaultChoice.agent}. Press Ctrl+C to cancel.`,
--- a/src/commands/skill/install.ts
+++ b/src/commands/skill/install.ts
@@ -1,5 +1,5 @@
 import { access } from 'node:fs/promises'
-import { createInterface } from 'node:readline'
+import checkbox, { Separator } from '@inquirer/checkbox'
 import chalk from 'chalk'
 import { CliError } from '../../lib/errors.js'
 import { getInstaller, listAgents, skillInstallers } from '../../lib/skills/index.js'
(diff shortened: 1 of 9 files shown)


doistbot (commented):
This PR introduces a helpful interactive chooser for `td skill install` when an agent is omitted, gracefully falling back to help text in non-interactive environments. The new fallback behavior and expanded test coverage are great additions that noticeably improve the CLI's developer experience. There are a couple of minor opportunities for refinement noted, specifically around filtering out undetected non-universal agents during global installs to avoid guaranteed failures, and using stricter integer validation to prevent typos from triggering accidental selections.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F295&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Acae859d0-4012-11f1-97ed-c0cc1d3d0ed0%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Acae859d0-4012-11f1-97ed-c0cc1d3d0ed0&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/skill/install.ts
```diff
@@ -1,12 +1,151 @@
...
+    )
+    const undetectedUniversal = choices.filter(
+        (choice) => !choice.detected && choice.agent === 'universal',
+    )
+
+    if (detectedNonUniversal.length > 0) {
+        return [
+            ...detectedNonUniversal,
+            ...detectedUniversal,
+            ...undetectedNonUniversal,
```
doistbot: [P2] In global mode, including `undetectedNonUniversal` here exposes choices that are guaranteed to fail later. `installer.install(false, ...)` rejects non-`universal` agents when their root dir does not exist, so on a machine without `~/.claude` or `~/.codex` the prompt still shows those as normal options even though selecting them can only end in `NOT_INSTALLED`. Filtering them out for global installs, or at least marking them unavailable, would make the chooser consistent with what can actually be installed.

src/commands/skill/install.ts
```diff
@@ -1,12 +1,151 @@
...
+        rl.question(`Enter a number [1]: `, (answer) => {
+            rl.close()
+            const input = answer.trim()
+
+            if (!input) {
+                resolve(defaultChoice.agent)
+                return
+            }
+
+            const selection = Number.parseInt(input, 10)
```
doistbot: [P3] `Number.parseInt()` will accept partial input such as `2abc` or `1.5` and treat it as a valid selection. That means a typo can install a skill instead of being rejected. Using `Number(input)` plus `Number.isInteger(...)`, or a digits-only check before parsing, would make invalid entries fail as intended.

scottlovegrove:
Not a blocker, but entering a number feels very _ancient_ 😄 Can we not have something where the user can interactively select the agent (even multiple ones)?

gnapse:
> Not a blocker, but entering a number feels very _ancient_ 😄 Can we not have something where the user can interactively select the agent (even multiple ones)?

Yes, but it is not trivial, and I wanted to get something out without having to bring in a lot of code or a dependency. But I was planning to add that later as a subsequent change.

But given your comment, I'll explore doing it here to see how it looks like in the code.

scottlovegrove:
Part of me wonders whether we should abandon our own skill install code and piggy back on the `skills` package. We could keep the updating part of it as we use it in `postinstall`.

gnapse:
> Part of me wonders whether we should abandon our own skill install code and piggy back on the `skills` package. We could keep the updating part of it as we use it in `postinstall`.

That could work. But I wished the command was shorter instead of:

```shell
npx skills add Doist/todoist-cli --skill todoist-cli
```

I needed a short command for our landing page. I wonder if we can rework our own `td skill install` command to call `npx skills add` itself 🤔

scottlovegrove:
I said get rid of the code, not the command. We could still keep our own command structure, but have it forwarded to skills (which I just noticed you said at the end of that comment)

gnapse:
> I said get rid of the code, not the command. We could still keep our own command structure, but have it forwarded to skills (which I just noticed you said at the end of that comment)

Will give it a try

PR #300, opened 2026-04-28
feat(auth): support multiple authenticated Todoist accounts

scottlovegrove:
## Summary

- Replace single-user auth state with a per-user store keyed by Todoist user id, selected via the new global `--user <id|email>` flag (parsed once and stripped from argv so every command honours it without per-command wiring) or via `user.defaultUser` in config.
- Move tokens into per-user keyring slots `user-<id>` and shift `auth_mode` / `auth_scope` / `auth_flags` onto per-user records. A best-effort postinstall migration (`src/lib/migrate-auth.ts`) upgrades v1 installs by hitting `/api/v1/user`; failure leaves v1 state untouched and the runtime resolver falls back to the legacy token until the next attempt.
- Update `td auth login` / `td auth token` to identify the account and upsert it (auto-default on first login). `td auth logout` errors when multiple accounts are stored without `--user`. `td auth status` shows the active account and lists the others.
- Add the minimum-viable `td user list` and `td user use <ref>` so the active account can be inspected and the default chosen. Other subcommands (`current`, `remove`, `default` alias) plus `td config view` / `td doctor` polish ship in a stacked follow-up.

## Test plan
- [x] `npm run type-check`
- [x] `npm test` — 1549 tests
- [x] `npm run check:skill-sync`
- [ ] Smoke: fresh install with no config → `td auth login` adds user, sets as default, `td task list` works
- [x] Smoke: second `td auth login` for a different account → both visible in `td user list`; default still the first
- [x] Smoke: `td --user <ref> task list` overrides per call; `td user use <ref>` switches the default
- [x] Smoke: install on a v1 box → postinstall migrates to v2, legacy keyring entry removed
- [ ] Smoke: install offline / with bad token → postinstall skips, runtime fallback keeps the v1 token usable

### DIFF
--- a/src/commands/auth/status.ts
+++ b/src/commands/auth/status.ts
@@ -47,11 +47,14 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
         return
     }
 
+    // env source wins over default: when running with TODOIST_API_TOKEN,
+    // showing `(default)` would hide the more important "this is an env
+    // override, not your stored credential" signal.
     const defaultMarker =
-        defaultUserId === user.id
-            ? ' (default)'
-            : metadata.source === 'env'
-              ? ' (TODOIST_API_TOKEN)'
+        metadata.source === 'env'
+            ? ' (TODOIST_API_TOKEN)'
+            : defaultUserId === user.id
+              ? ' (default)'
               : ''
     console.log(chalk.green('✓'), `Authenticated${defaultMarker}`)
     console.log(`  Email: ${user.email}`)
--- a/src/commands/auth/status.ts
+++ b/src/commands/auth/status.ts
@@ -1,6 +1,7 @@
 import chalk from 'chalk'
 import { getApi } from '../../lib/api/core.js'
-import { type AuthMode, getAuthMetadata } from '../../lib/auth.js'
+import { type AuthMode, getAuthMetadata, listStoredUsers, readConfig } from '../../lib/auth.js'
+import { getDefaultUserId } from '../../lib/users.js'
 
 function formatAuthMode(authMode: AuthMode, authScope?: string): string {
     if (authMode === 'read-only') {
@@ -16,6 +17,9 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
     const api = await getApi()
     const user = await api.getUser()
     const metadata = await getAuthMetadata()
+    const config = await readConfig()
+    const storedUsers = await listStoredUsers()
+    const defaultUserId = getDefaultUserId(config)
     const modeLabel = formatAuthMode(metadata.authMode, metadata.authScope)
 
     if (options.json) {
@@ -28,15 +32,44 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
                     authMode: metadata.authMode,
                     authScope: metadata.authScope,
                     authFlags: metadata.authFlags,
+                    source: metadata.source,
+                    isDefault: defaultUserId === user.id,
+                    storedUsers: storedUsers.map((u) => ({
+                        id: u.id,
+                        email: u.email,
+                        isDefault: defaultUserId === u.id,
+                    })),
                 },
                 null,
                 2,
             ),
         )
-    } else {
-        console.log(chalk.green('✓'), 'Authenticated')
-        console.log(`  Email: ${user.email}`)
-        console.log(`  Name:  ${user.fullName}`)
-        console.log(`  Mode:  ${modeLabel}`)
+        return
+    }
+
+    const defaultMarker =
+        defaultUserId === user.id
+            ? ' (default)'
+            : metadata.source === 'env'
+              ? ' (TODOIST_API_TOKEN)'
+              : ''
+    console.log(chalk.green('✓'), `Authenticated${defaultMarker}`)
+    console.log(`  Email: ${user.email}`)
+    console.log(`  Name:  ${user.fullName}`)
+    console.log(`  Mode:  ${modeLabel}`)
+
+    const others = storedUsers.filter((u) => u.id !== user.id)
+    if (others.length > 0) {
+        console.log()
+        console.log(chalk.dim(`Other stored accounts (${others.length}):`))
+        for (const other of others) {
+            const marker = other.id === defaultUserId ? chalk.dim(' (default)') : ''
+            console.log(`  ${other.email} ${chalk.dim(`(id:${other.id})`)}${marker}`)
+        }
+        console.log(
+            chalk.dim(
+                'Use `td user use <id|email>` to switch default, or `--user <ref>` per command.',
+            ),
+        )
     }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,14 +15,6 @@ import { preloadMarkdown } from './lib/markdown.js'
 import { formatError, formatErrorJson } from './lib/output.js'
 import { startEarlySpinner, stopEarlySpinner } from './lib/spinner.js'
 
-// Seed the global-args cache from the *original* argv (so `--user <ref>` is
-// captured) before stripping the flag from process.argv. Commander has no
-// global-option attachment, so leaving `--user` in argv would make every
-// subcommand error on it. Stripping here lets every command transparently
-// honor `--user` without per-command wiring.
-getRequestedUserRef()
-process.argv = [process.argv[0], process.argv[1], ...stripUserFlag(process.argv.slice(2))]
-
 program
     .name('td')
     .description('Todoist CLI')
@@ -182,6 +174,38 @@ const commands: Record<string, [string, () => Promise<(p: Command) => void>]> = 
 // Register placeholders so --help lists all commands
 for (const [name, [description]] of Object.entries(commands)) {
     program.command(name).description(description)
+}
+
+// Validate `--user` and strip it from argv before commander parses.
+//
+// Commander has no global-option attachment, so leaving `--user` in argv
+// would make every subcommand error on it. We capture the value via
+// `parseGlobalArgs` first and remove it here. Before stripping we surface
+// usage errors that the parser can detect now but commander never will
+// because it never sees the flag — bare `--user`, `--user=` (empty), and
+// `--user <known-subcommand>` (almost always a forgotten value).
+{
+    const originalArgs = process.argv.slice(2)
+    const sawUserFlag = originalArgs.some((a) => a === '--user' || a.startsWith('--user='))
+    if (sawUserFlag) {
+        const ref = getRequestedUserRef()
+        const reportUserFlagError = (message: string, hints: string[]): never => {
+            const err = new CliError('USER_FLAG_INVALID', message, hints)
+            console.error(isJsonMode() ? formatErrorJson(err) : formatError(err))
+            process.exit(1)
+        }
+        if (!ref) {
+            reportUserFlagError('--user requires a value: <id|email>.', [
+                'Example: td --user scott@doist.com task list',
+            ])
+        } else if (Object.hasOwn(commands, ref)) {
+            reportUserFlagError(
+                `--user requires a value: <id|email>. Got "${ref}", which looks like a subcommand — did you forget the value?`,
+                [`Example: td --user scott@doist.com ${ref}`],
+            )
+        }
+    }
+    process.argv = [process.argv[0], process.argv[1], ...stripUserFlag(originalArgs)]
 }
 
 // completion-server needs the command tree to walk for completions.
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,11 +3,25 @@
 import { type Command, program } from 'commander'
 import packageJson from '../package.json' with { type: 'json' }
 import { CliError } from './lib/errors.js'
-import { isJsonMode, isNdjsonMode, isRawMode } from './lib/global-args.js'
+import {
+    getRequestedUserRef,
+    isJsonMode,
+    isNdjsonMode,
+    isRawMode,
+    stripUserFlag,
+} from './lib/global-args.js'
 import { initializeLogger } from './lib/logger.js'
 import { preloadMarkdown } from './lib/markdown.js'
 import { formatError, formatErrorJson } from './lib/output.js'
 import { startEarlySpinner, stopEarlySpinner } from './lib/spinner.js'
+
+// Seed the global-args cache from the *original* argv (so `--user <ref>` is
+// captured) before stripping the flag from process.argv. Commander has no
+// global-option attachment, so leaving `--user` in argv would make every
+// subcommand error on it. Stripping here lets every command transparently
+// honor `--user` without per-command wiring.
+getRequestedUserRef()
+process.argv = [process.argv[0], process.argv[1], ...stripUserFlag(process.argv.slice(2))]
 
 program
     .name('td')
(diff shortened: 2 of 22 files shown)


src/lib/auth.ts
```diff
@@ -63,251 +90,392 @@ export interface TokenStorageResult {
...
+    }
+    if (!input.email) {
+        throw new CliError('INVALID_USER', 'Cannot store user record: missing email')
     }
 
+    const trimmedToken = input.token.trim()
     const config = await readConfig()
-    const configToken = getConfigToken(config)
+    const previouslyExisted = getStoredUsers(config).some((u) => u.id === input.id)
+    const shouldSetDefault = getStoredUsers(config).length === 0 && !config.user?.defaultUser
```
doistbot: [P2] Orphaned `defaultUser` prevents auto-defaulting

If the config has 0 stored users but `config.user?.defaultUser` is set to an orphaned ID (e.g., from manual edits or unexpected state), `shouldSetDefault` will evaluate to `false`. The first added user won't become the default, and adding a second user will throw `NoUserSelectedError` because the default remains invalid. It's safer to always set the first user as the default if the store is empty.

```suggestion
    const shouldSetDefault = getStoredUsers(config).length === 0
```

src/commands/auth/status.ts
```diff
@@ -28,15 +32,44 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
-    } else {
-        console.log(chalk.green('✓'), 'Authenticated')
-        console.log(`  Email: ${user.email}`)
-        console.log(`  Name:  ${user.fullName}`)
-        console.log(`  Mode:  ${modeLabel}`)
+        return
+    }
+
+    const defaultMarker =
+        defaultUserId === user.id
```
doistbot: [P2] Environment token indicator overshadowed by default marker

If the CLI is run with `TODOIST_API_TOKEN` and the token happens to belong to the `defaultUser`, this condition prints `(default)` instead of `(TODOIST_API_TOKEN)`. The environment variable override is critical context for debugging and should take precedence in the output.

```suggestion
        metadata.source === 'env'
            ? ' (TODOIST_API_TOKEN)'
            : defaultUserId === user.id
              ? ' (default)'
```

src/lib/auth.ts
```diff
@@ -63,251 +90,392 @@ export interface TokenStorageResult {
...
     const config = await readConfig()
-    const configToken = getConfigToken(config)
-    const secureStore = createSecureStore()
+    const users = getStoredUsers(config)
+    const requestedRef = opts.ref ?? getRequestedUserRef()
+
+    // If the user is on a v1 (legacy) config and we have no v2 users yet, fall
+    // through to the legacy-token path so existing installs keep working until
+    // postinstall (or a later `td auth login`) migrates them.
+    if (users.length === 0) {
```
doistbot, line 128: [P1] Empty v2 configs still fall back to the legacy token path. `users.length === 0` is true both for a real v1 install and for a migrated/logged-out v2 config with `users: []`, so if postinstall failed to delete the old `api-token` keyring entry, `td auth logout` can remove the user-scoped token and the next command will silently authenticate again via the stale legacy credential. This fallback needs to be gated on the legacy shape (`config.users` absent), not on the array being empty.

src/lib/auth.ts
```diff
@@ -63,251 +90,392 @@ export interface TokenStorageResult {
...
-                token: storedToken.trim(),
-                metadata: {
-                    authMode: config.auth_mode ?? 'unknown',
-                    authScope: config.auth_scope,
-                    authFlags: config.auth_flags,
-                    source: 'secure-store',
-                },
-            }
-        }
+        await writeConfig(next)
```
doistbot, line 269: [P1] A `writeConfig(next)` failure is no longer just metadata loss in the multi-user model. If the keyring write succeeds but the config update fails, the new account has no `config.users` record (and possibly no default), so `td auth login` / `td auth token` can print success and exit 0 even though later commands cannot resolve that account at all. This should fail hard or roll back the secure-store write instead of being downgraded to a warning.

src/lib/auth.ts
```diff
@@ -63,251 +90,392 @@ export interface TokenStorageResult {
...
     }
 
-    await writeConfig(withPendingSecureStoreClear(withoutAuthMetadata(config)))
-    return {
-        storage: 'config-file',
-        warning: buildFallbackWarning('local auth state cleared in'),
+    const next = stripLegacyAuthFields(removeStoredUser(ensureV2Shape(config), id))
+
+    try {
+        await writeConfig(next)
```
doistbot: [P1] This leaves a broken stored account behind when the config write fails after the secret has already been deleted. The user remains in `config.users`, so `td user list` and default-user resolution still see the account, but any later command against it fails with `NO_TOKEN` because the credential is already gone. Logout should either restore the secret on config-write failure or surface the operation as failed instead of returning a warning.

src/index.ts
```diff
@@ -3,12 +3,26 @@
...
 import { formatError, formatErrorJson } from './lib/output.js'
 import { startEarlySpinner, stopEarlySpinner } from './lib/spinner.js'
 
+// Seed the global-args cache from the *original* argv (so `--user <ref>` is
+// captured) before stripping the flag from process.argv. Commander has no
+// global-option attachment, so leaving `--user` in argv would make every
+// subcommand error on it. Stripping here lets every command transparently
+// honor `--user` without per-command wiring.
+getRequestedUserRef()
+process.argv = [process.argv[0], process.argv[1], ...stripUserFlag(process.argv.slice(2))]
```
doistbot: [P2] Stripping `--user` before Commander parses argv also strips away its normal missing-value validation. For example, `td --user --json auth status` now consumes `--json` as the user ref, removes the real `--json` flag, and ends up as `USER_NOT_FOUND("--json")` instead of a usage error. Please validate that `--user` is followed by a non-flag value before mutating `process.argv`.

doistbot (commented):
This PR introduces a comprehensive multi-user authentication system by transitioning to per-user config and keyring stores, alongside a new global `--user` flag and dedicated account management commands. The architectural shift is well-thought-out and provides a clean foundation for managing multi-account workflows with smooth legacy migrations. There are just a few edge cases to address regarding state synchronization and configuration, specifically ensuring that config write failures don't leave the secure store out of sync, refining the legacy fallback logic to prevent stale authentications, and hardening the manual argument parsing for the new global flag.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F300&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad6cb41c0-42fd-11f1-8436-6afff3346a44%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad6cb41c0-42fd-11f1-8436-6afff3346a44&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
scottlovegrove: Fixed in 992795c — the legacy fallback in `resolveActiveUser` (and `clearApiToken`) is now gated on `!Array.isArray(config.users)` so a logged-out v2 config doesn't peek at the legacy `api-token` keyring slot. New test in `src/lib/auth.test.ts` covers this.

src/lib/auth.ts
scottlovegrove: Fixed in 992795c — `upsertUser` now rolls back the keyring secret and throws `CONFIG_WRITE_FAILED` when the config write fails, rather than warning. Test added.

src/lib/auth.ts
scottlovegrove: Fixed in 992795c — `removeUserById` now writes the config first and only deletes the secret afterward. A config-write failure throws `CONFIG_WRITE_FAILED` with the keyring untouched, so a retry is safe. Test added.

src/lib/auth.ts
scottlovegrove: Fixed in 992795c — `shouldSetDefault` is now `getStoredUsers(config).length === 0`. An orphaned `defaultUser` no longer blocks first-login from claiming the slot. Test added.

src/commands/auth/status.ts
scottlovegrove: Fixed in 992795c — env source now wins the marker. `(TODOIST_API_TOKEN)` takes precedence over `(default)` in `auth status`.

src/index.ts
scottlovegrove: Fixed in 992795c — both `parseGlobalArgs` and `stripUserFlag` now refuse to consume the next argv element as a `--user` value when it starts with `-`. `td --user --json ...` now leaves `--json` intact for commander to surface a usage error. Tests added.

scottlovegrove:
@doistbot /review

src/lib/auth.ts
```diff
@@ -63,251 +90,427 @@ export interface TokenStorageResult {
...
+}
 
-        throw new NoTokenError()
+/**
+ * Like `resolveActiveUser` but returns whichever credentials are at hand
+ * without mutating storage. Useful for `td doctor` / `td config view` where
+ * we want to inspect (and report on) what would be used.
+ */
+export async function probeApiToken(): Promise<{ token: string; metadata: AuthMetadata }> {
+    const resolved = await resolveActiveUser()
```
doistbot, line 190: [P2] `probeApiToken()` now inherits `resolveActiveUser()`'s strict "must have a selected user" behavior. After a normal flow like A(default) → add B/C → `td auth logout --user A`, both `td doctor` and `td config view` will start failing with `NO_USER_SELECTED` instead of helping the user recover, because they both go through this probe helper. Returning an explicit "ambiguous user" probe result here would keep those diagnostic commands usable in the very state they need to diagnose.

src/lib/auth.ts
```diff
@@ -63,251 +90,427 @@ export interface TokenStorageResult {
...
+        return { token: user.api_token.trim(), source: 'config-file' }
+    }
+    const secureStore = createSecureStore(accountForUser(user.id))
+    try {
+        const stored = await secureStore.getSecret()
+        if (stored?.trim()) {
+            return { token: stored.trim(), source: 'secure-store' }
+        }
+    } catch (error) {
+        if (!(error instanceof SecureStoreUnavailableError)) throw error
```
doistbot: [P2] This path currently turns a `SecureStoreUnavailableError` into `NoTokenError`. With a v2 user stored only in the keyring, `td config view` / `td doctor` will report "no credentials" instead of surfacing that the credential manager is unavailable, even though those callers already have dedicated handling for the unavailable-store case. Re-throwing the secure-store error here preserves that distinction and keeps the diagnostics accurate.

src/index.ts
doistbot: [P2] Rewriting `process.argv` before Commander parses options also removes Commander's required-argument validation for `--user`. For example, `td --user auth status` (missing ref) is interpreted as `user="auth"`, the real subcommand disappears, and the user gets an unrelated unknown-command error. This needs an explicit validation step before stripping, otherwise malformed `--user` invocations become very confusing to recover from.

doistbot (commented):
This PR introduces excellent support for managing multiple Todoist accounts, including a per-user keyring store, global `--user` flag parsing, and new user management commands. The migration path and overall architecture cleanly separate user context from command logic, greatly enhancing the flexibility of the CLI. There are a few edge cases to address around diagnostic tools failing when a default user is unset or the secure store is unavailable, as well as ensuring the global `--user` flag is properly validated before it gets stripped from the arguments.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F300&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4deba340-4302-11f1-88b5-1e87e137b74e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4deba340-4302-11f1-88b5-1e87e137b74e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
scottlovegrove: Taken care of in #301

scottlovegrove: Already covered by #301 — `view.ts` there catches `NoUserSelectedError` in `probeTokenPresence` and renders an ambiguous-state `Active:` line. `td doctor` already catches it in #300 (the `NoUserSelectedError` branch in `checkAuthentication`). Once both PRs land the diagnostic flow stays usable; rather than reshape the `probeApiToken` API I'd prefer to keep the throw and rely on the consumer-side handling that #301 introduces. Happy to revisit if you'd rather see the producer-side change.

src/lib/auth.ts
scottlovegrove: Fixed in e21a0b4 — `loadTokenForStoredUser` now re-throws `SecureStoreUnavailableError` rather than turning it into `NoTokenError`. `getAuthMetadata` still catches both errors and returns the "unknown" default so permission/scope checks don't hard-fail when the keyring is offline; only `probeApiToken` (the diagnostic helper) lets it propagate. New test in `auth.test.ts` covers both paths.

src/index.ts
scottlovegrove: Fixed in e21a0b4 — `index.ts` now validates `--user` before stripping. Bare `--user` / `--user=` and `--user <known-subcommand>` (the `auth` case in your example) now error with `USER_FLAG_INVALID` and a clear hint (`got "auth", which looks like a subcommand — did you forget the value?`). Other malformed cases (`--user --json …`) were already covered by the previous round of fixes.

nvignola:
✅

PR #301, opened 2026-04-28
feat(user,config,doctor): polish multi-user surface

scottlovegrove:
Stacked on top of #300 — review/merge that one first.

## Summary

- Add `td user current`, `td user remove <ref>`, and a `td user default <ref>` alias for `use`.
- Teach `td config view` about multi-user state: shows the active account, lists all stored accounts (mode/scope/flags/storage) with a default marker, surfaces the ambiguous-resolution case, and masks per-user \`api_token\` entries in JSON output.
- Add a \`users\` check to \`td doctor\` (count, default presence, plaintext fallback) and recognise the \"multiple stored, no default\" warning state in the auth check.

## Test plan
- [x] \`npm run type-check\`
- [x] \`npm test\` — 1557 tests
- [x] \`npm run check\` (lint + format)
- [x] \`npm run check:skill-sync\`
- [ ] Smoke: \`td user current\` shows the right account in default / \`--user\` / env / legacy modes
- [ ] Smoke: \`td user remove <ref>\` deletes the keyring slot + config entry; clears default if it was the target
- [ ] Smoke: \`td config view\` with two accounts shows the active + others block; \`--json\` masks plaintext tokens
- [ ] Smoke: \`td doctor\` warns when multiple accounts are stored without a default

### DIFF
--- a/src/commands/config/view.ts
+++ b/src/commands/config/view.ts
@@ -1,7 +1,15 @@
 import chalk from 'chalk'
-import { type AuthMetadata, NoTokenError, probeApiToken, TOKEN_ENV_VAR } from '../../lib/auth.js'
+import {
+    type AuthMetadata,
+    listStoredUsers,
+    NoTokenError,
+    probeApiToken,
+    type StoredUser,
+    TOKEN_ENV_VAR,
+} from '../../lib/auth.js'
 import { type Config, CONFIG_PATH, readConfigStrict } from '../../lib/config.js'
 import { SECURE_STORE_DESCRIPTION, SecureStoreUnavailableError } from '../../lib/secure-store.js'
+import { getDefaultUserId, NoUserSelectedError } from '../../lib/users.js'
 
 export interface ViewConfigOptions {
     json?: boolean
@@ -11,6 +19,7 @@ export interface ViewConfigOptions {
 type TokenStatus =
     | { state: 'present'; token: string; metadata: AuthMetadata }
     | { state: 'missing' }
+    | { state: 'ambiguous' }
     | { state: 'unavailable'; reason: string }
 
 function maskToken(token: string): string {
@@ -35,6 +44,7 @@ async function probeTokenPresence(): Promise<TokenStatus> {
         return { state: 'present', token, metadata }
     } catch (error) {
         if (error instanceof NoTokenError) return { state: 'missing' }
+        if (error instanceof NoUserSelectedError) return { state: 'ambiguous' }
         if (error instanceof SecureStoreUnavailableError) {
             return {
                 state: 'unavailable',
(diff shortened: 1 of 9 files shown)


src/commands/config/view.ts
```diff
@@ -62,32 +72,85 @@ function renderTokenLine(token: TokenStatus, showToken: boolean): string {
...
+    if (token.state === 'present' && token.metadata.email) {
+        const isDefault = token.metadata.userId === defaultUserId
+        const marker =
+            token.metadata.source === 'env' ? ' (TODOIST_API_TOKEN)' : isDefault ? ' (default)' : ''
+        lines.push(
+            `  Active:        ${token.metadata.email} ${chalk.dim(`(id:${token.metadata.userId ?? '—'})`)}${marker}`,
+        )
+    } else if (token.state === 'present' && token.metadata.source === 'env') {
+        lines.push(`  Active:        ${chalk.dim('(TODOIST_API_TOKEN)')}`)
+    } else if (token.state === 'ambiguous') {
```
doistbot, line 121: [P2] Missing fallback for `present` state without email

If a token is present but lacks an email (e.g., legacy single-user credentials) and is not sourced from `env`, the `Active:` line will be entirely omitted from the output because it falls through the `if`/`else` chain. Adding a catch-all branch ensures the UI remains consistent.

```suggestion
    } else if (token.state === 'present') {
        lines.push(`  Active:        ${chalk.dim('(legacy single-user credentials)')}`)
    } else if (token.state === 'ambiguous') {
```

src/commands/doctor.ts
```diff
@@ -269,10 +286,54 @@ async function checkForUpdates(offline: boolean): Promise<DoctorCheck> {
...
+    const config = await readConfig()
+    const defaultId = getDefaultUserId(config)
+    const plaintext = users.filter((u) => u.api_token).map((u) => u.email)
+    const details: Record<string, unknown> = {
+        count: users.length,
+        defaultUserId: defaultId,
+        plaintextFallbackEmails: plaintext,
+    }
+
+    if (users.length > 1 && !defaultId) {
```
doistbot: [P2] This warning does not match the CLI’s real account-resolution rules. With `TODOIST_API_TOKEN` set, commands without `--user` will not error even when there is no default, and a stale `defaultUserId` currently skips this warning even though auth resolution still fails. It would be better to derive this state from the same logic as `resolveActiveUser` (or at least validate that the configured default exists before treating it as selected).

doistbot (commented):
This PR nicely polishes the multi-user surface by introducing new user management commands, updating the configuration view, and adding helpful doctor checks. These additions significantly improve visibility into stored accounts and make credential management much smoother. A few minor edge cases remain to be ironed out, specifically regarding missing output fallbacks for legacy single-user credentials, aligning doctor warnings with actual account-resolution rules, and ensuring partial cleanup warnings aren't incorrectly suppressed by the quiet flag during account removal.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F301&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aef6ec850-42fd-11f1-9b97-b36d28bdccc4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aef6ec850-42fd-11f1-9b97-b36d28bdccc4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/user/remove.ts
```diff
@@ -0,0 +1,32 @@
...
+        )
+    }
+
+    const config = await readConfig()
+    const { user } = requireUserByRef(config, ref)
+    const wasDefault = getDefaultUserId(config) === user.id
+
+    const result = await removeUserById(user.id)
+
+    if (!isQuiet()) {
```
doistbot, line 21: [P2] Wrapping the warning path in `!isQuiet()` means `td --quiet user remove ...` also suppresses cleanup warnings from `removeUserById()`. `--quiet` is supposed to hide success chatter, but warnings like partial keyring/config cleanup still need to reach stderr so the user knows removal was incomplete.

src/commands/config/view.ts
scottlovegrove: Fixed in e37491b — added a fallback branch for legacy single-user credentials (token present, no email, source ≠ env) plus a generic catch-all so the `Active:` line is never omitted.

src/commands/doctor.ts
scottlovegrove: Fixed in e37491b — `checkStoredUsers` now skips the multi-user warning when `TODOIST_API_TOKEN` is set (env wins before resolver runs) and treats `defaultUser` as selected only when it actually resolves to a stored user. Orphaned defaults now surface as their own warning message and detail field.

src/commands/user/remove.ts
scottlovegrove: Fixed in e37491b — split the success-chatter block (still gated by `--quiet`) from the warning emission, so partial-cleanup warnings always reach stderr.

scottlovegrove:
@doistbot /review

src/commands/config/view.ts
```diff
@@ -62,32 +72,92 @@ function renderTokenLine(token: TokenStatus, showToken: boolean): string {
...
+            `  Active:        ${token.metadata.email} ${chalk.dim(`(id:${token.metadata.userId ?? '—'})`)}${marker}`,
+        )
+    } else if (token.state === 'present' && token.metadata.source === 'env') {
+        lines.push(`  Active:        ${chalk.dim('(TODOIST_API_TOKEN)')}`)
+    } else if (token.state === 'present') {
+        // Token is resolvable but we don't know which Todoist account it
+        // belongs to — i.e. legacy v1 single-user credentials before
+        // postinstall migration runs.
+        lines.push(`  Active:        ${chalk.dim('(legacy single-user credentials)')}`)
+    } else if (token.state === 'ambiguous') {
```
doistbot, line 121: [P2] This branch collapses two different states into the same message. `NoUserSelectedError` is also thrown when `user.defaultUser` points at a deleted account, so `td config view` currently says "no default" even when there is a configured default and it is stale. Since `defaultUserId` and `users` are already available here, it would be more useful to report the orphaned-pointer case explicitly.

src/commands/doctor.ts
```diff
@@ -269,10 +286,69 @@ async function checkForUpdates(offline: boolean): Promise<DoctorCheck> {
...
+    const plaintext = users.filter((u) => u.api_token).map((u) => u.email)
+    const details: Record<string, unknown> = {
+        count: users.length,
+        defaultUserId: defaultId,
+        defaultUserResolved: hasUsableDefault,
+        envTokenSet,
+        plaintextFallbackEmails: plaintext,
+    }
+
+    if (users.length > 1 && !hasUsableDefault && !envTokenSet) {
```
doistbot, line 333: [P2] The early return here suppresses the plaintext-fallback warning whenever the install is also in the "multiple users, no usable default" state. In that scenario `td doctor` only reports the missing default, so the new `users` check no longer surfaces that some accounts are stored plaintext unless the caller inspects `details`. Consider folding the plaintext info into this warning or emitting a second check so both conditions are visible.

doistbot (commented):
This PR nicely rounds out the multi-user experience by introducing new user lifecycle commands alongside expanded diagnostic output in `td config view` and `td doctor`. These updates significantly enhance visibility into complex account states and make managing multiple profiles much more intuitive. A couple of edge cases in the diagnostic reporting could be refined, specifically distinguishing between an unset default and an orphaned default pointer in the config view, as well as ensuring the plaintext-fallback warning isn't inadvertently suppressed when an environment lacks a usable default user.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F301&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab7ef8060-4305-11f1-8121-fbedcfd0507e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab7ef8060-4305-11f1-8121-fbedcfd0507e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/config/view.ts
scottlovegrove: Fixed in 0b6e1dd — the ambiguous-state `Active:` line now distinguishes "no default selected" from "default points at unknown user <id>" using the `defaultUserId` and `users` we already have on hand.

src/commands/doctor.ts
scottlovegrove: Fixed in 0b6e1dd — `checkStoredUsers` now returns up to three checks rather than early-returning. A single install can hit both the missing/orphaned-default warning and the plaintext-fallback warning, and they now surface independently. `runDoctorChecks` flattens via `...await checkStoredUsers()`.

PR #302, opened 2026-04-29
feat: track cli usage in request headers

gnapse:
## Summary

This PR makes Todoist CLI identify itself consistently to Todoist backend services.

Every Todoist-bound request now carries a shared header contract so backend logs can reliably recognize CLI traffic, attribute it to a low-cardinality command label, and correlate related requests from the same CLI process.

That coverage applies across the main SDK-backed API client as well as direct Todoist fetch paths such as OAuth and legacy auth migration, so the backend sees the same shape regardless of how the request was made.

## Why

Before this change, the CLI did not consistently describe itself across outbound requests to Todoist services.

This PR standardizes that request-level contract so backend tracking can happen from normal API logs instead of requiring a separate client-side telemetry system.

## What changed

- Introduced one shared request-metadata layer for Todoist-bound requests instead of piecemeal per-call headers.
- Standardized command attribution so backend logs receive a stable low-cardinality command label, including routed `td view ...` flows and internal non-Commander flows.
- Added tests around header generation and auth-related request paths, and documented the shared contract in the codebase.

## What every request now sends

All requests from the CLI to Todoist backend services now send the following headers:

- `User-Agent: todoist-cli/<version>`
  Identifies the binary and its version in the conventional user-agent slot.

- `doist-platform: cli`
  Explicit machine-readable platform marker for backend classification.

- `doist-version: <version>`
  Dedicated version field so backend consumers do not need to parse `User-Agent`.

- `doist-os: <normalized os>`
  Normalized operating-system label for environment-level usage analysis.

- `X-TD-Request-Id: <uuid>`
  A new UUID for every outbound request. This is request-scoped, not process-scoped.

- `X-TD-Session-Id: <uuid>`
  A stable UUID for the lifetime of the current CLI process. Multiple requests from the same run share this value.

- `X-TD-CLI-Command: <command label>`
  A low-cardinality, lower-case, colon-separated label describing the CLI action responsible for the request.
  Examples include `today`, `task:add`, `project:list`, `task:view`, and `postinstall:auth-migrate`.

A few details about `X-TD-CLI-Command`:
- Normal Commander-driven commands derive it automatically from the active command path.
- Routed `td view ...` flows override the top-level `view` label with the routed target command, so backend logs see the real action rather than just `view`.
- Internal flows that do not naturally run under Commander can provide an explicit low-cardinality override.
- If no specific command context is available, the fallback value is `unknown`, so the header is still present.

### DIFF
--- a/src/commands/view.ts
+++ b/src/commands/view.ts
@@ -1,6 +1,7 @@
 import { Command } from 'commander'
 import { CliError } from '../lib/errors.js'
 import { classifyTodoistUrl } from '../lib/refs.js'
+import { setActiveCommandPath } from '../lib/usage-tracking.js'
 
 function looksLikeTodoistAppUrl(token: string): boolean {
     return /^https?:\/\/app\.todoist\.com\/app\/\S+/.test(token)
@@ -36,11 +37,16 @@ function extractViewInvocation(
 async function runRoutedCommand(
     loadRegister: () => Promise<(program: Command) => void>,
     argv: string[],
+    commandPath: string,
 ): Promise<void> {
     const proxy = new Command()
     proxy.exitOverride()
     const register = await loadRegister()
     register(proxy)
+    // `src/index.ts` already recorded the top-level `view` command on the main
+    // program. Override it here so downstream API requests are attributed to
+    // the routed target command (`task.view`, `today`, etc.) instead.
+    setActiveCommandPath(commandPath)
     await proxy.parseAsync(['node', 'td', ...argv])
 }
 
@@ -78,11 +84,13 @@ Examples:
                         return runRoutedCommand(
                             async () => (await import('./today.js')).registerTodayCommand,
                             ['today', ...invocation.passthroughArgs],
+                            'today',
                         )
                     case 'upcoming':
                         return runRoutedCommand(
                             async () => (await import('./upcoming.js')).registerUpcomingCommand,
                             ['upcoming', ...invocation.passthroughArgs],
+                            'upcoming',
                         )
                 }
             }
@@ -93,21 +101,25 @@ Examples:
                     return runRoutedCommand(
                         async () => (await import('./task/index.js')).registerTaskCommand,
                         ['task', 'view', ref, ...invocation.passthroughArgs],
+                        'task.view',
                     )
                 case 'project':
                     return runRoutedCommand(
                         async () => (await import('./project/index.js')).registerProjectCommand,
                         ['project', 'view', ref, ...invocation.passthroughArgs],
+                        'project.view',
                     )
                 case 'label':
                     return runRoutedCommand(
                         async () => (await import('./label/index.js')).registerLabelCommand,
                         ['label', 'view', ref, ...invocation.passthroughArgs],
+                        'label.view',
                     )
                 case 'filter':
                     return runRoutedCommand(
                         async () => (await import('./filter/index.js')).registerFilterCommand,
                         ['filter', 'show', ref, ...invocation.passthroughArgs],
+                        'filter.show',
                     )
             }
         })
--- a/src/commands/view.ts
+++ b/src/commands/view.ts
@@ -45,7 +45,7 @@ async function runRoutedCommand(
     register(proxy)
     // `src/index.ts` already recorded the top-level `view` command on the main
     // program. Override it here so downstream API requests are attributed to
-    // the routed target command (`task.view`, `today`, etc.) instead.
+    // the routed target command (`task:view`, `today`, etc.) instead.
     setActiveCommandPath(commandPath)
     await proxy.parseAsync(['node', 'td', ...argv])
 }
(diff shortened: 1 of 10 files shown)


src/lib/usage-tracking.ts
```diff
@@ -0,0 +1,125 @@
...
+        statusText: response.statusText,
+        headers: Object.fromEntries(response.headers.entries()),
+        text: () => response.text(),
+        json: () => response.json(),
+    }
+}
+
+export function createTrackedFetch(baseFetch: typeof fetch = globalThis.fetch): CustomFetch {
+    return async (url, options = {}) => {
+        const { timeout: _timeout, headers, ...rest } = options
```
doistbot: [P1] Lost request timeouts due to stripped `timeout` property

By destructuring and explicitly ignoring the `timeout` property, any timeouts configured by the `@doist/todoist-sdk` are lost. This means network requests can hang indefinitely if the connection stalls, which is especially problematic for a CLI.

Node.js native `fetch` uses `AbortSignal` for timeouts instead of a `timeout` option. You can map the SDK's `timeout` property to an `AbortSignal.timeout()` to preserve the expected behavior. Node 20.18+ natively supports both `AbortSignal.timeout` and `AbortSignal.any`.

```suggestion
        const { timeout: timeoutMs, headers, signal, ...rest } = options

        let abortSignal = signal
        if (timeoutMs) {
            const timeoutSignal = AbortSignal.timeout(timeoutMs)
            abortSignal = signal ? AbortSignal.any([signal, timeoutSignal]) : timeoutSignal
        }

        const response = await baseFetch(url, {
            ...rest,
            signal: abortSignal,
            headers: mergeTodoistHeaders(headers),
        })
```

doistbot (commented):
This PR establishes a consistent mechanism for tracking CLI usage by injecting standard client identity headers and command paths into all backend requests. This is a great addition that will significantly improve observability and our ability to correlate backend logs with specific CLI interactions. There is just one flagged issue to address regarding the custom fetch wrapper inadvertently stripping the `timeout` property, which could lead to indefinitely hanging network requests if not mapped to an `AbortSignal`.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F302&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A316f0f40-4415-11f1-83c8-d6738c7ae4e4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A316f0f40-4415-11f1-83c8-d6738c7ae4e4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/usage-tracking.ts
```diff
@@ -0,0 +1,133 @@
...
+        : activeCommandPath
+
+    const headers: Record<string, string> = {
+        'User-Agent': getUserAgent(),
+        'doist-platform': 'cli',
+        'doist-version': CLI_VERSION,
+        'doist-os': getDoistOs(),
+        'X-TD-Request-Id': randomUUID(),
+        'X-TD-Session-Id': SESSION_ID,
+        'X-Todoist-CLI-Command': normalizedCommandPath ?? 'unknown',
```
henningmu: nit (and unsure) but shouldn't it be `TD-CLI` to match the other `*-TD-*` headers
```suggestion
        'X-TD-CLI-Command': normalizedCommandPath ?? 'unknown',
```

src/commands/view.ts
```diff
@@ -36,11 +37,16 @@ function extractViewInvocation(
 async function runRoutedCommand(
     loadRegister: () => Promise<(program: Command) => void>,
     argv: string[],
+    commandPath: string,
```
henningmu, line 40: Why do we need to pass this in instead of inferring it from `argv`?

henningmu (approved):
No objections, this is privacy-friendly tracking. Please double-check that backend actually forwards the relevant headers into Datadog (might be selective)

src/lib/usage-tracking.ts
```diff
@@ -0,0 +1,133 @@
...
+    statusText: string
+    headers: Record<string, string>
+    text(): Promise<string>
+    json(): Promise<unknown>
+}
+
+type CustomFetch = (
+    url: string,
+    options?: RequestInit & { timeout?: number },
+) => Promise<CustomFetchResponse>
```
scottlovegrove: These are redundant and could be brought in from the SDK.

src/commands/view.ts
gnapse: Inferring it requires some heuristics about how many of the initial argv values to pick to build the command path. This simplifies the work needed to be done.

gnapse:
> double-check that backend actually forwards the relevant headers into Datadog

Mostly yes, but not completely. `User-Agent`, `doist-platform`, `doist-version`, `doist-os`, `X-TD-Request-Id`, and `X-TD-Session-Id` already make it into Datadog logs via the existing backend log context. `X-Todoist-CLI-Command` does not currently get promoted into Datadog, so command-level attribution still needs backend work if we want that visible there. I knew this beforehand. I added that tracking value just in case we do the change in the future.

PR #305, opened 2026-05-01
fix: honour HTTP_PROXY on the customFetch path

### LINKED ISSUE #304: HTTP_PROXY/HTTPS_PROXY no longer honored after 1.60.0
## Summary

  Starting in 1.60.0, the CLI ignores the standard `HTTP_PROXY`,
  `HTTPS_PROXY`, and `NO_PROXY` environment variables. Requests go
  direct, which breaks usage in environments where outbound traffic
  must transit a corporate proxy.

  ## Affected versions

  - Broken: 1.60.0
  - Last working: 1.59.0

  ## Reproduction

  1. Set `HTTPS_PROXY=http://localhost:8888` (or any proxy of your choice).
  2. Run any CLI command that hits the API, e.g. `td today`.
  3. Observe that the request goes direct to `api.todoist.com` rather
     than through the configured proxy.

  Under 1.59.0, the same setup routes the request through the proxy.

  ## Root cause

_This section is all LLM generated and unverified_

  `@doist/todoist-sdk` honors `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`
  via undici's `EnvHttpProxyAgent`, but only on its **native-fetch**
  code path. In `transport/fetch-with-retry.js`:

  ```js
  if (customFetch) {
      fetchResponse = await customFetch(url, ...)   // no dispatcher
  } else {
      const dispatcher = await getDefaultDispatcher() // EnvHttpProxyAgent
      nativeFetchOptions.dispatcher = dispatcher
      ...
  }
```

  In #302 (commit 5c692f8), `src/lib/api/core.ts` started passing a
  `customFetch` to the SDK to attach usage-tracking headers:

```js
  const rawApi = new TodoistApi(token, { customFetch: createTrackedFetch() })
```

  `createTrackedFetch` in `src/lib/usage-tracking.ts` wraps
  `globalThis.fetch` directly, with no dispatcher. That silently
  bypasses the SDK's `EnvHttpProxyAgent`, so the proxy env vars are
  no longer applied to any SDK-routed request.

  The same shape applies to `fetchTodoist` in `oauth.ts` and
  `migrate-auth.ts` — those also call `globalThis.fetch` without a
  dispatcher, though in those flows the pre-1.60.0 code was already
  calling bare fetch, so it's only a regression on the SDK path.

scottlovegrove:
## Summary

- Bumps `@doist/todoist-sdk` to `10.1.0` (which now publicly exports `getDefaultDispatcher`).
- In `createTrackedFetch` and `fetchTodoist`, attach the SDK's shared `EnvHttpProxyAgent` dispatcher whenever the underlying fetch is the real native one — restoring `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` support on the `customFetch` path.
- Test stubs continue to skip the dispatcher so their captured options stay clean.

Fixes #304.

## Why

#302 (1.60.0) routed every Todoist-bound request through a `customFetch` so we could attach the usage-tracking headers. The SDK's `transport/fetch-with-retry.ts` only attaches the proxy dispatcher on its **native-fetch** branch — once a `customFetch` is supplied, that branch is skipped, so the proxy env vars were silently dropped. This regressed every corporate-proxy user from 1.60.0 onwards.

The SDK fix landed in [Doist/todoist-sdk-typescript#594](https://github.com/Doist/todoist-sdk-typescript/pull/594), shipped as 10.1.0. This PR consumes that and wires it into our two outbound paths.

## Test plan

- [x] `npm run check` clean (lint + format)
- [x] `npm test` — 1577 passing (4 new dispatcher cases added)
- [ ] Manual proxy check: `HTTPS_PROXY=http://localhost:8888 td today` transits the proxy
- [ ] Manual NO_PROXY check: `HTTPS_PROXY=http://localhost:8888 NO_PROXY=api.todoist.com td today` bypasses the proxy
- [ ] Manual OAuth check: `HTTPS_PROXY=http://localhost:8888 td auth login` token exchange transits the proxy

### DIFF
--- a/src/lib/usage-tracking.test.ts
+++ b/src/lib/usage-tracking.test.ts
@@ -1,4 +1,5 @@
-import { describe, expect, it } from 'vitest'
+import { getDefaultDispatcher } from '@doist/todoist-sdk'
+import { afterEach, describe, expect, it, vi } from 'vitest'
 import {
     buildUsageTrackingHeaders,
     createTrackedFetch,
@@ -7,6 +8,16 @@ import {
     resetUsageTrackingForTests,
     setActiveCommandPath,
 } from './usage-tracking.js'
+
+vi.mock('@doist/todoist-sdk', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('@doist/todoist-sdk')>()
+    return {
+        ...actual,
+        getDefaultDispatcher: vi.fn(() => Promise.resolve(undefined)),
+    }
+})
+
+const getDefaultDispatcherMock = vi.mocked(getDefaultDispatcher)
 
 describe('usage tracking', () => {
     it('normalizes commander command paths into header-friendly values', () => {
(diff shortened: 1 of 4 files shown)


doistbot (commented):
This PR effectively restores proxy support for corporate users by integrating the newly exposed `getDefaultDispatcher` from the SDK into the outbound `customFetch` paths. It is a crucial fix that cleanly bridges the proxy regression while keeping test stubs properly isolated. A few refinements are noted to ensure explicitly provided dispatchers aren't inadvertently overwritten, to extract the duplicated proxy-attachment logic into a shared helper, and to streamline both the test mock setups and TypeScript assertions for better maintainability.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F305&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Af7e6d140-4577-11f1-8cd1-a0c581f7e32f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Af7e6d140-4577-11f1-8cd1-a0c581f7e32f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/usage-tracking.ts
```diff
@@ -72,6 +76,14 @@ function mergeTodoistHeaders(
     return Object.fromEntries(mergedHeaders.entries())
 }
 
+async function attachDispatcher(options: RequestInit): Promise<void> {
+    const dispatcher = await getDefaultDispatcher()
+    if (dispatcher !== undefined) {
+        // @ts-expect-error - dispatcher is a valid option for Node's fetch but not in the TS types
```
doistbot, line 91: [P2] This assignment unconditionally replaces any `dispatcher` already present on `fetchOptions`. Because both `fetchTodoist` and `createTrackedFetch` spread arbitrary request options first, a caller that explicitly passes a custom Undici dispatcher will now have it silently clobbered by the env proxy agent. Please preserve an existing dispatcher and only call `getDefaultDispatcher()` when one was not supplied.

src/lib/usage-tracking.test.ts
```diff
@@ -113,6 +124,111 @@ describe('usage tracking', () => {
...
+        })
+
+        it('attaches the env proxy dispatcher when createTrackedFetch uses native fetch', async () => {
+            const fakeDispatcher = { kind: 'env-http-proxy-agent' } as unknown as NonNullable<
+                Awaited<ReturnType<typeof getDefaultDispatcher>>
+            >
+            getDefaultDispatcherMock.mockResolvedValue(fakeDispatcher)
+
+            let captured: RequestInit | undefined
+            const originalFetch = globalThis.fetch
```
doistbot: [P3] This manual `globalThis.fetch` override and the associated `try/finally` block are duplicated across multiple tests. Consider using `vi.spyOn(globalThis, 'fetch').mockImplementation(...)` which automatically handles restoration and avoids the duplicated boilerplate. You could also extract `fakeDispatcher` to the outer `describe` block.

src/lib/usage-tracking.ts
```diff
@@ -93,26 +109,35 @@ export function createTrackedFetch(baseFetch: typeof fetch = globalThis.fetch):
...
     commandPath?: string,
 ): Promise<Response> {
     const { headers, ...rest } = options
-    return fetchImpl(url, {
+    const fetchOptions: RequestInit = {
         ...rest,
         headers: mergeTodoistHeaders(headers, commandPath),
-    })
+    }
+    if (fetchImpl === globalThis.fetch) {
```
doistbot: [P2] This repeats the same native-fetch check and `attachDispatcher(fetchOptions)` behavior that `createTrackedFetch` now has. Since the goal here is to keep both Todoist outbound paths proxy-aware in exactly the same way, please extract a shared helper (for example `maybeAttachDispatcher(fetchImpl, options)` or a shared fetch-options builder) and call it from both places so these paths do not drift again.

src/lib/usage-tracking.ts
doistbot, line 91: [P3] Using `@ts-expect-error` here will cause a compiler error if `@types/node` or TypeScript later updates `RequestInit` to include undici's `dispatcher` property. Consider removing the comment and using a type assertion instead to future-proof the build: `(options as RequestInit & { dispatcher: unknown }).dispatcher = dispatcher`.

scottlovegrove: I'd push back on this one — I think the framing is inverted.

`@ts-expect-error` is *more* future-proof than a cast here, not less: when `@types/node` eventually adds `dispatcher` to `RequestInit`, the directive will start erroring (because the line no longer needs suppression) and tell us to delete the workaround. A `(options as RequestInit & { dispatcher: unknown }).dispatcher = ...` cast would silently persist forever, even after the typing is fixed upstream.

This is also exactly the pattern the SDK itself uses at the equivalent call site in `transport/fetch-with-retry.ts`:

```js
// @ts-expect-error - dispatcher is a valid option for Node.js fetch but not in the TS types
nativeFetchOptions.dispatcher = dispatcher
```

So the directive is doing exactly its intended job — flagging "this is a known type-system gap" — and I think it's the right call to keep it.

Resolving as not addressed.

PR #311, opened 2026-05-02
feat(auth): add `td auth token view` for script and agent use

scottlovegrove:
## Summary

- New `td auth token view` subcommand prints the stored API token for the active user (or `--user <ref>`) to stdout, so scripts can capture it via `TOKEN=$(td auth token view)` without ever surfacing the token in an agent's conversation context.
- Refuses when `TODOIST_API_TOKEN` is set in the environment — the token is already available there, and silently echoing it would mask multi-user `--user` requests when env happens to be set. New error code `TOKEN_FROM_ENV` with actionable hints.
- Wired as a nested subcommand of the existing `auth token` command. Commander matches subcommand names before falling through to the parent's `[token]` positional, so `view` is dispatched as a subcommand and never treated as a literal token value. `td auth token <real-token>` keeps working as before, since real Todoist tokens are 40-char hex strings.

## Test plan

- [x] `npm test -- src/commands/auth/auth.test.ts` — 24/24 pass (4 new: happy path, env refusal, `NoTokenError` propagation, `UserNotFoundError` propagation with `--user`)
- [x] `npm run check:skill-sync` — `skills/todoist-cli/SKILL.md` regenerated and in sync
- [ ] Manual: `TOKEN=$(td auth token view); echo "len=${#TOKEN}"`
- [ ] Manual: `td auth token view --user <other-email>` resolves the right account
- [ ] Manual: `TODOIST_API_TOKEN=fake td auth token view` exits non-zero with `TOKEN_FROM_ENV`
- [ ] Manual: `td auth token <some-token>` still saves (no regression on the parent action)

### DIFF
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -76,6 +76,7 @@ import {
     resolveActiveUser,
     upsertUser,
 } from '../../lib/auth.js'
+import { resetGlobalArgs } from '../../lib/global-args.js'
 import { startCallbackServer } from '../../lib/oauth-server.js'
 import { buildAuthorizationUrl, exchangeCodeForToken } from '../../lib/oauth.js'
 import { UserNotFoundError } from '../../lib/users.js'
@@ -562,9 +563,21 @@ describe('auth command', () => {
             const program = createProgram()
             mockResolveActiveUser.mockRejectedValue(new UserNotFoundError('missing@example.com'))
 
-            await expect(
-                program.parseAsync(['node', 'td', 'auth', 'print-token']),
-            ).rejects.toHaveProperty('code', 'USER_NOT_FOUND')
+            // `--user <ref>` is parsed from process.argv by the global-args
+            // layer (not commander) and stripped before commander sees the
+            // argv. Stub process.argv to mirror production wiring so the
+            // test exercises the same code path as a real invocation.
+            const originalArgv = process.argv
+            process.argv = ['node', 'td', 'auth', 'print-token', '--user', 'missing@example.com']
+            resetGlobalArgs()
+            try {
+                await expect(
+                    program.parseAsync(['node', 'td', 'auth', 'print-token']),
+                ).rejects.toHaveProperty('code', 'USER_NOT_FOUND')
+            } finally {
+                process.argv = originalArgv
+                resetGlobalArgs()
+            }
             expect(consoleSpy).not.toHaveBeenCalled()
         })
     })
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -11,6 +11,7 @@ vi.mock('../../lib/auth.js', async (importOriginal) => {
         getAuthMetadata: vi.fn(),
         listStoredUsers: vi.fn(),
         readConfig: vi.fn(),
+        resolveActiveUser: vi.fn(),
     }
 })
 
@@ -67,14 +68,17 @@ import open from 'open'
 import { createApiForToken, getApi } from '../../lib/api/core.js'
 import {
     NoTokenError,
+    TOKEN_ENV_VAR,
     clearApiToken,
     getAuthMetadata,
     listStoredUsers,
     readConfig,
+    resolveActiveUser,
     upsertUser,
 } from '../../lib/auth.js'
 import { startCallbackServer } from '../../lib/oauth-server.js'
 import { buildAuthorizationUrl, exchangeCodeForToken } from '../../lib/oauth.js'
+import { UserNotFoundError } from '../../lib/users.js'
 import { createMockApi } from '../../test-support/mock-api.js'
 import { registerAuthCommand } from './index.js'
 
@@ -85,6 +89,7 @@ const mockClearApiToken = vi.mocked(clearApiToken)
 const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
 const mockListStoredUsers = vi.mocked(listStoredUsers)
 const mockReadConfig = vi.mocked(readConfig)
+const mockResolveActiveUser = vi.mocked(resolveActiveUser)
 const mockGetApi = vi.mocked(getApi)
 const mockCreateApiForToken = vi.mocked(createApiForToken)
 const mockStartCallbackServer = vi.mocked(startCallbackServer)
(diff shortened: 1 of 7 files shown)


src/lib/skills/content.ts
```diff
@@ -38,6 +38,8 @@ td auth login --read-only --additional-scopes=backups
 td auth login --additional-scopes=app-management,backups
 td auth token
 td auth status
+td auth print-token
```
doistbot: [P2] This agent-facing quick reference now shows a bare `td auth print-token` invocation, which would print the secret directly into an agent/tool transcript if followed literally. Since the intended safe usage is shell capture, please replace the bare examples with captured usage such as `TOKEN=$(td auth print-token)` / `TOKEN=$(td auth print-token --user ...)`.

src/commands/auth/auth.test.ts
```diff
@@ -498,4 +503,69 @@ describe('auth command', () => {
...
+            ).rejects.toHaveProperty('code', 'NO_TOKEN')
+            expect(consoleSpy).not.toHaveBeenCalled()
+        })
+
+        it('propagates UserNotFoundError when --user ref does not match', async () => {
+            const program = createProgram()
+            mockResolveActiveUser.mockRejectedValue(new UserNotFoundError('missing@example.com'))
+
+            await expect(
+                program.parseAsync(['node', 'td', 'auth', 'print-token']),
```
doistbot: [P3] The test description explicitly mentions testing "when --user ref does not match", but the `--user` flag is omitted from the simulated CLI arguments. Consider adding `'--user', 'missing@example.com'` to the `parseAsync` array so the test execution accurately reflects its intent.

doistbot (commented):
This PR introduces the `td auth print-token` subcommand to securely retrieve stored API tokens for scripts and agents, complete with thoughtful environment variable handling. The addition elegantly supports automation workflows without complicating the existing authentication commands. A few minor adjustments would polish this up, specifically updating the agent quick reference examples to use shell capture rather than bare invocations to prevent transcript token leaks, and ensuring the missing `--user` flag is included in the user mismatch test to align with its description.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F311&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A969ec2f0-4624-11f1-9593-cf539ed91f48%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A969ec2f0-4624-11f1-9593-cf539ed91f48&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/skills/content.ts
scottlovegrove: Fixed in 04c88f4 — replaced both bare invocations in the auth quick reference with the `TOKEN=$(...)` capture form, and reinforced the surrounding prose so an agent following the docs literally won't leak the secret into its transcript.

src/commands/auth/auth.test.ts
scottlovegrove: Fixed in 04c88f4 — `--user` is consumed by the global-args layer from `process.argv` and stripped before commander sees the argv (commander would otherwise reject it as an unknown option), so the test now stubs `process.argv` and resets the cached parse, mirroring how production wires the flag.

henningmu:
> so scripts can capture it via TOKEN=$(td auth print-token) without ever surfacing the token in an agent's conversation context.

That's nice, but it does not prevent an agent from running `td auth print-token` and accessing it that way. Is that a vector we want to defend against?

> Naming: auth print-token (flat sibling) rather than auth token view to avoid colliding with the existing auth token [token] save command.

If we only match `auth token view`, there are no clashes – unless someone wants to set their token to `view`, which is not supported anyway 😁

scottlovegrove:
> That's nice, but it does not prevent an agent from running `td auth print-token` and accessing it that way. Is that a vector we want to defend against?

It crossed my mind, but I don't know how we can. There's a note in the skill that says agents should not call it directly themselves 🤷🏻‍♂️ 

> If we only match `auth token view`, there are no clashes – unless someone wants to set their token to `view`, which is not supported anyway 😁

This is what I originally wanted, but was convinced by the argument that `view` could then be construed as a token. I really would prefer `token view`... 😄

gnapse:
> That's nice, but it does not prevent an agent from running `td auth print-token` and accessing it that way. Is that a vector we want to defend against?

I was also thinking about this, alongside the "why". Why we need this? I know you said the following:

> * so scripts can capture it via `TOKEN=$(td auth print-token)`

But why do we need scripts to capture the token?

scottlovegrove:
> But why do we need scripts to capture the token?

- I was doing some work with an agent and getting it to do some direct API calls, and I noticed my token appearing in the Claude chat a _lot_. 
- I also based this on a PR comment from @henningmu in another repo about not wanting to have an environment variable being used and should use the authenticated token from the CLI (for which there was no extraction method)
- It's just about _us_. I've seen claude want to make direct calls to github, so it gets the token via the gh cli, so there could be other places, where users may wish to setup some kind of local automation that requires a Todoist token. 
- I read an article the other day (wish I'd saved it), that said having tokens as an environment variable (like in bash/zsh/fish/etc) means that your agent has access to it as soon as you start it, whether you asked it to or not. So having this guardrails against that scenario as well.

gnapse (approved):
> having tokens as an environment variable

In any case, I was not arguing for that either. That is clearly worse. That's why we initially had it in a config file (still not great) or in the OS secure storage that we use now.

But anyway, I see that `gh` CLI also has something like this. Maybe it's ok.

PR #313, opened 2026-05-06
refactor: integrate @doist/cli-core 0.2.0

scottlovegrove:
## Summary

First consumer-side integration of [@doist/cli-core](https://www.npmjs.com/package/@doist/cli-core. Replaces locally-duplicated implementations with the shared package and proves the extraction surface end-to-end.

What's wired through cli-core now:

| Module | Was | Now |
| --- | --- | --- |
| `src/lib/errors.ts` | full local `CliError` class | local subclass over `BaseCliError<ErrorCode>`, keeps the historical positional `(code, message, hints?, type?)` signature so the ~200 existing call sites stay untouched |
| `src/lib/config.ts` | local path / read / strict-read / write impls | wraps `getConfigPath` / `readConfig` / `readConfigStrict` / `writeConfig`; schema (`Config`, `validateConfigForDoctor`, all the `KNOWN_*_KEYS`) stays |
| `src/lib/spinner.ts` | `process.stdout.isTTY` direct read | `isStdoutTTY()` |
| `src/lib/global-args.ts` | `process.env.CI` direct read | `isCI()` |
| `src/lib/output.ts` | inline `JSON.stringify(_, null, 2)` / `items.map(JSON.stringify).join('\n')` | delegates serialization to `formatJson` / `formatNdjson`; entity-aware processing (`processItem`, `_meta` cursor sentinel) stays |

`formatErrorJson` deliberately keeps its inline compact `JSON.stringify` — it emits single-line JSON, not pretty-print, so it doesn't fit cli-core's `formatJson` primitive. Same for `shouldDisableSpinner`, which keeps the `TD_SPINNER` env-var override and global-args fan-out (CLI-specific flags).

## vitest config

Added `server.deps.inline: ['@doist/cli-core']` to `vitest.config.ts`. Without it, vitest treats the real-installed cli-core as external and `vi.doMock('node:fs/promises', …)` in `auth.test.ts` doesn't reach its compiled imports, which fails 23 auth/migrate-auth tests.

## Net effect

8 files changed, +92/-107. The tests are unchanged.

## Test plan

- [x] `npm run type-check`
- [x] `npm run check`
- [x] `npm test` — 1589 / 1589 pass
- [ ] Manually exercise `td auth login` / `td auth status` / `td doctor` against a real config to confirm parity at the wrapper boundary

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1,5 +1,5 @@
 import {
-    getConfigPath,
+    getConfigPath as getConfigPathCore,
     readConfig as readConfigCore,
     readConfigStrict as readConfigStrictCore,
     writeConfig as writeConfigCore,
@@ -7,7 +7,17 @@ import {
 import { CliError } from './errors.js'
 import { normalizeHelpCenterLocale } from './help-center.js'
 
-export const CONFIG_PATH = getConfigPath('todoist-cli')
+const APP_NAME = 'todoist-cli'
+
+/**
+ * Resolve the canonical config path lazily. Computing on each call (instead of
+ * caching at module load) keeps the path responsive to vitest's `vi.doMock`
+ * for `node:os` — which only reliably reaches cli-core's compiled `homedir()`
+ * call after the mock has been set up by the test, not at import time.
+ */
+export function getConfigPath(): string {
+    return getConfigPathCore(APP_NAME)
+}
 
 export type AuthMode = 'read-only' | 'read-write' | 'unknown'
 export type UpdateChannel = 'stable' | 'pre-release'
@@ -101,7 +111,9 @@ export const AUTH_FLAGS: ReadonlySet<AuthFlag> = new Set(AUTH_FLAG_ORDER)
 const UPDATE_CHANNELS: ReadonlySet<UpdateChannel> = new Set(['stable', 'pre-release'])
 
 export async function readConfig(): Promise<Config> {
-    return (await readConfigCore<Config>(CONFIG_PATH)) as Config
+    // cli-core's readConfig returns Partial<T> via {} on any failure, so the
+    // result is always an object — Config's all-optional fields make the cast safe.
+    return (await readConfigCore<Config>(getConfigPath())) as Config
 }
 
 export type StrictReadResult = { state: 'missing' } | { state: 'present'; config: Config }
@@ -112,7 +124,8 @@ export type StrictReadResult = { state: 'missing' } | { state: 'present'; config
  * swallows errors for runtime code paths; this one surfaces them.
  */
 export async function readConfigStrict(): Promise<StrictReadResult> {
-    const result = await readConfigStrictCore(CONFIG_PATH)
+    const path = getConfigPath()
+    const result = await readConfigStrictCore(path)
     switch (result.state) {
         case 'missing':
             return { state: 'missing' }
@@ -121,13 +134,13 @@ export async function readConfigStrict(): Promise<StrictReadResult> {
         case 'read-failed':
             throw new CliError(
                 'CONFIG_READ_FAILED',
-                `Could not read config file ${CONFIG_PATH}: ${result.error.message}`,
+                `Could not read config file ${path}: ${result.error.message}`,
                 ['Check file permissions, or run `td doctor` to diagnose'],
             )
         case 'invalid-json':
             throw new CliError(
                 'CONFIG_INVALID_JSON',
-                `Config file at ${CONFIG_PATH} is not valid JSON: ${result.error.message}`,
+                `Config file at ${path} is not valid JSON: ${result.error.message}`,
                 [
                     'Fix the JSON by hand, or delete the file and re-authenticate with `td auth login`',
                 ],
@@ -135,7 +148,7 @@ export async function readConfigStrict(): Promise<StrictReadResult> {
         case 'invalid-shape':
             throw new CliError(
                 'CONFIG_INVALID_SHAPE',
-                `Config file at ${CONFIG_PATH} must contain a JSON object (got ${result.actual})`,
+                `Config file at ${path} must contain a JSON object (got ${result.actual})`,
                 [
                     'Fix the JSON by hand, or delete the file and re-authenticate with `td auth login`',
                 ],
@@ -144,7 +157,7 @@ export async function readConfigStrict(): Promise<StrictReadResult> {
 }
 
 export async function writeConfig(config: Config): Promise<void> {
-    await writeConfigCore(CONFIG_PATH, config, { deleteWhenEmpty: true })
+    await writeConfigCore(getConfigPath(), config, { deleteWhenEmpty: true })
 }
 
 // Keep this validator ad-hoc for now: it is only used by `td doctor`, and the
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1,10 +1,13 @@
-import { chmod, mkdir, readFile, unlink, writeFile } from 'node:fs/promises'
-import { homedir } from 'node:os'
-import { dirname, join } from 'node:path'
+import {
+    getConfigPath,
+    readConfig as readConfigCore,
+    readConfigStrict as readConfigStrictCore,
+    writeConfig as writeConfigCore,
+} from '@doist/cli-core'
 import { CliError } from './errors.js'
 import { normalizeHelpCenterLocale } from './help-center.js'
 
-export const CONFIG_PATH = join(homedir(), '.config', 'todoist-cli', 'config.json')
+export const CONFIG_PATH = getConfigPath('todoist-cli')
 
 export type AuthMode = 'read-only' | 'read-write' | 'unknown'
 export type UpdateChannel = 'stable' | 'pre-release'
@@ -98,13 +101,7 @@ export const AUTH_FLAGS: ReadonlySet<AuthFlag> = new Set(AUTH_FLAG_ORDER)
 const UPDATE_CHANNELS: ReadonlySet<UpdateChannel> = new Set(['stable', 'pre-release'])
 
 export async function readConfig(): Promise<Config> {
-    try {
-        const content = await readFile(CONFIG_PATH, 'utf-8')
-        const parsed = JSON.parse(content)
-        return isObject(parsed) ? (parsed as Config) : {}
-    } catch {
-        return {}
-    }
+    return readConfigCore<Config>(CONFIG_PATH)
 }
 
 export type StrictReadResult = { state: 'missing' } | { state: 'present'; config: Config }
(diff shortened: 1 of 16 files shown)


src/lib/output.ts
```diff
@@ -441,11 +434,9 @@ export function formatPaginatedJson<T extends object>(
     full = false,
     showUrl = false,
 ): string {
-    if (!type) {
-        return JSON.stringify(data, null, 2)
-    }
+    if (!type) return formatJsonCore(data)
```
doistbot: [P2] `formatPaginatedJson`/`formatPaginatedNdjson` now call `formatJsonCore`/`formatNdjsonCore` directly even though `formatJson` and `formatNdjson` in this file are the CLI’s serialization wrappers. After pre-processing the items, reuse those local helpers (`formatJson({ results, nextCursor })`, `formatNdjson(items as object[])`) so future CLI-specific serialization tweaks stay centralized instead of drifting between paginated and non-paginated output.

src/lib/config.ts
```diff
@@ -1,10 +1,13 @@
...
+    getConfigPath,
+    readConfig as readConfigCore,
+    readConfigStrict as readConfigStrictCore,
+    writeConfig as writeConfigCore,
+} from '@doist/cli-core'
 import { CliError } from './errors.js'
 import { normalizeHelpCenterLocale } from './help-center.js'
 
-export const CONFIG_PATH = join(homedir(), '.config', 'todoist-cli', 'config.json')
+export const CONFIG_PATH = getConfigPath('todoist-cli')
```
doistbot: [P2] `CONFIG_PATH` is now delegated to cli-core, but this repo still has multiple hard-coded `~/.config/todoist-cli/...` strings in user-facing hints/docs. That splits the source of truth for the config location and will drift the next time cli-core changes path resolution. Please centralize a display path/dir next to `CONFIG_PATH` and use that everywhere instead of duplicating the literal path.

src/lib/config.ts
```diff
@@ -98,13 +101,7 @@ export const AUTH_FLAGS: ReadonlySet<AuthFlag> = new Set(AUTH_FLAG_ORDER)
...
 
 export async function readConfig(): Promise<Config> {
-    try {
-        const content = await readFile(CONFIG_PATH, 'utf-8')
-        const parsed = JSON.parse(content)
-        return isObject(parsed) ? (parsed as Config) : {}
-    } catch {
-        return {}
-    }
+    return (await readConfigCore<Config>(CONFIG_PATH)) as Config
```
doistbot: [P2] This wrapper now asserts `Config` instead of enforcing its contract. The old implementation normalized non-object JSON to `{}` before returning; with the blind cast, callers can no longer rely on `readConfig()` always returning an object if cli-core ever returns a scalar/array. Keep the object-shape guard in this layer (or require a cli-core API that guarantees object output) so the local contract stays explicit.

src/lib/output.ts
```diff
@@ -376,10 +371,8 @@ export function formatNdjson<T extends object>(
     full = false,
     showUrl = false,
 ): string {
-    if (!type) {
-        return items.map((item) => JSON.stringify(item)).join('\n')
-    }
-    return items.map((item) => JSON.stringify(processItem(item, type, full, showUrl))).join('\n')
+    if (!type) return formatNdjsonCore(items)
+    return formatNdjsonCore(items.map((item) => processItem(item, type, full, showUrl)))
```
doistbot, line 375: [P2] This typed NDJSON path now materializes a full `processItem(...)` array before serialization, whereas the removed implementation transformed and `JSON.stringify`’d each item in the same pass. For larger `--ndjson` outputs that adds avoidable allocation and a second walk over the data. Consider keeping the transform+stringify fused here, or adding a cli-core helper that accepts a projector.

src/lib/output.ts
```diff
@@ -454,18 +445,13 @@ export function formatPaginatedNdjson<T extends object>(
...
     showUrl = false,
 ): string {
-    const lines = data.results.map((item) => {
-        if (!type) {
-            return JSON.stringify(item)
-        }
-        return JSON.stringify(processItem(item, type, full, showUrl))
-    })
-
+    const items: unknown[] = type
```
doistbot: [P2] `formatPaginatedNdjson` now does extra O(n) work on every page: the untyped branch clones `data.results` with `[...data.results]`, and the typed branch builds a full processed-item array, before `formatNdjsonCore` serializes everything. This helper is used by most `list --ndjson` commands, so the extra copy/pass shows up in a hot path. You can avoid it by passing `data.results` straight through when there’s no cursor, and only allocating a new array when you actually need to append the `_meta` record.

src/lib/config.ts
```diff
@@ -115,61 +112,39 @@ export type StrictReadResult = { state: 'missing' } | { state: 'present'; config
...
-    }
-
-    if (!isObject(parsed)) {
-        const actual = Array.isArray(parsed) ? 'array' : typeof parsed
-        throw new CliError(
-            'CONFIG_INVALID_SHAPE',
-            `Config file at ${CONFIG_PATH} must contain a JSON object (got ${actual})`,
-            ['Fix the JSON by hand, or delete the file and re-authenticate with `td auth login`'],
-        )
+    const result = await readConfigStrictCore(CONFIG_PATH)
```
doistbot: [P2] This switch adds new cli-core adapter branches, but the test suite still only exercises the malformed-JSON path indirectly via `config view`. There’s no focused coverage that `read-failed` and `invalid-shape` are translated back into this CLI’s historical `CONFIG_READ_FAILED` / `CONFIG_INVALID_SHAPE` errors, which makes the wrapper boundary easy to regress if `@doist/cli-core` changes. Add a small `src/lib/config.test.ts` that mocks `readConfigStrictCore()` and asserts each branch maps to the expected `CliError`.

vitest.config.ts
```diff
@@ -7,5 +7,14 @@ export default defineConfig({
         globals: true,
         root: 'src',
         include: ['**/*.test.ts'],
+        // Inline @doist/cli-core so vitest's module mocks (e.g. vi.doMock for
```
doistbot, line 10: [P3] This PR changes test tooling behavior by requiring `@doist/cli-core` to be inlined under Vitest. Repo guidance says `CODEBASE.md` should be updated when build/test/release tooling changes, so please add a short note there so the repo map stays accurate.

doistbot (commented):
This PR successfully integrates the new `@doist/cli-core` package, replacing local implementations for error handling, configuration reading, and output formatting. This is a great milestone that effectively reduces duplication while preserving existing CLI contracts and behaviors. There are a few refinements to consider, specifically optimizing memory efficiency in NDJSON output serialization by avoiding unnecessary array allocations, standardizing the use of local serialization wrappers, centralizing the configuration display path to avoid hardcoded strings, restoring the object-shape guard for config reads, adding test coverage for config error mappings, and noting the Vitest tooling change in CODEBASE.md.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F313&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A973d3080-497f-11f1-8a94-db8c392dab7e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A973d3080-497f-11f1-8a94-db8c392dab7e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/output.ts
scottlovegrove: Fixed in db32089 — `formatPaginatedJson` / `formatPaginatedNdjson` now route through the local `formatJson` / `formatNdjson` wrappers (which themselves delegate to cli-core's primitives). Future CLI-specific serialization tweaks now live in one place for both paginated and non-paginated paths.

src/lib/config.ts
scottlovegrove: Noted but deferred to a follow-up — agreed the hard-coded `~/.config/todoist-cli/...` strings in user-facing hints should reference a single source. They live across hint arrays in errors and CLI copy in commands; pulling them through a shared `describeConfigDir()` (or similar) is a wider sweep than this PR's scope. I'll open a follow-up issue / PR rather than bolt it on here.

src/lib/config.ts
scottlovegrove: Acknowledged. The cast looks blind but isn't lossy: cli-core's `readConfig<T>` returns `Partial<T>` and is implemented to fall back to `{}` on any failure mode (missing / unreadable / invalid-JSON / non-object), so the runtime value is always an object. `Config`'s fields are all optional, which makes `Partial<Config>` and `Config` structurally equivalent. Added a short comment explaining the invariant inline. Happy to add a runtime `isObject` guard if you'd prefer belt-and-braces; my read is the cli-core API contract is already enough.

src/lib/output.ts
scottlovegrove: Noted. The fused transform+stringify version saved one walk per item; the new code separates them. At todoist page sizes (default 200 items, max 500) the extra allocation isn't measurable in our local timings. If/when this shows up as a hot path I'd add a `formatNdjsonWith(items, projector)` to cli-core rather than re-inline the loop here, so the optimisation is shared. Leaving as-is for now.

src/lib/output.ts
scottlovegrove: Fixed in db32089 — the no-cursor path now passes `data.results` straight through to the local `formatNdjson` wrapper (no `[...data.results]` clone, no extra walk). The cursor path still has to materialize because we need to append the `_meta` record before serialization; that's unavoidable without the projector helper from the previous comment.

src/lib/config.ts
scottlovegrove: Added in db32089 — `src/lib/config.test.ts` covers every branch of the wrapper translation. Each test mocks cli-core's `readConfigStrict` to return one of the discriminated states and asserts the wrapper's behaviour: `missing` and `present` pass through, `read-failed` / `invalid-json` / `invalid-shape` throw a `CliError` with the expected todoist code and hint copy.

vitest.config.ts
scottlovegrove: Added in db32089 — `CODEBASE.md > Testing` now documents the `server.deps.inline` requirement and the reason for it.

engfragui:
✅ Seems like the majority of the changes are to replace `CONFIG_PATH` with `getConfigPath` (which are easy to review). Although I might not understand all the nuances, the rest of the changes seem sane too 👌

PR #314, opened 2026-05-06
refactor(spinner): delegate to @doist/cli-core 0.3.0

scottlovegrove:
## Summary

Replaces the local spinner module with `createSpinner` from `@doist/cli-core` 0.3.0.

```ts
const spinner = createSpinner({ isDisabled: shouldDisableSpinner })

export const {
    LoadingSpinner,
    withSpinner,
    startEarlySpinner,
    stopEarlySpinner,
} = spinner
```

cli-core owns the whole module — early-spinner singleton with stdout interception, adopt/release semantics for chained API calls, the `LoadingSpinner` class, the `withSpinner` wrapper, the colour palette. The only CLI-specific bit is `shouldDisableSpinner` (the `TD_SPINNER` env var + global-args fan-out for `--json` / `--ndjson` / `--no-spinner` / etc.), which we wire in as `isDisabled`.

## Behaviour change worth flagging

`yocto-spinner.success()` and `.error()` already prepend their own `✔` / `✖` glyphs. The old code added a manual `✓ ` / `✗ ` on top, which produced doubled symbols in real terminals (e.g. `✔ ✓ Done`). cli-core 0.3 drops the manual prefixes and tests confirm the bare text is forwarded — so the success/fail line now renders as a single `✔ Done` instead of `✔ ✓ Done`. Surfaced during the cli-core review.

## Other changes

- Bumps `@doist/cli-core` 0.2.0 → 0.3.0.
- Drops `yocto-spinner` from direct dependencies; it's now reached transitively through cli-core.
- Trims `spinner.test.ts` to 4 smoke tests covering the wrapper's public API plus `withSpinner`'s happy + error paths. Deep behaviour (adoption, glyph rendering, stdout-interceptor lifecycle, `--json`-style gating) is exhaustively covered in cli-core's own `spinner.test.ts`.

## Test plan

- [x] `npm run type-check`
- [x] `npm run check`
- [x] `npm test` — 1570 / 1570 pass against the published `@doist/cli-core@0.3.0`
- [ ] Manually run a long command (`td today`, `td auth login`) to confirm spinner UX still feels right and the glyph isn't doubled

### DIFF
--- a/src/lib/spinner.test.ts
+++ b/src/lib/spinner.test.ts
@@ -40,4 +40,30 @@ describe('spinner wrapper', () => {
         const s = new LoadingSpinner()
         expect(s.start({ text: 'noop', noSpinner: true })).toBe(s)
     })
+
+    it('forwards shouldDisableSpinner to cli-core (TD_SPINNER=false suppresses the spinner)', () => {
+        // shouldDisableSpinner is the wrapper's only CLI-specific hook.
+        // Force an interactive, non-CI context so the *only* thing left
+        // gating the spinner is shouldDisableSpinner reading TD_SPINNER.
+        const stdout = process.stdout as unknown as { isTTY?: boolean }
+        const originalIsTTY = stdout.isTTY
+        stdout.isTTY = true
+        const originalCI = process.env.CI
+        delete process.env.CI
+        process.env.TD_SPINNER = 'false'
+
+        const writeBefore = process.stdout.write
+        try {
+            startEarlySpinner()
+            // If the wiring is intact, cli-core saw isDisabled() === true
+            // and never installed its stdout interceptor.
+            expect(process.stdout.write).toBe(writeBefore)
+        } finally {
+            stdout.isTTY = originalIsTTY
+            if (originalCI === undefined) delete process.env.CI
+            else process.env.CI = originalCI
+            delete process.env.TD_SPINNER
+            stopEarlySpinner()
+        }
+    })
 })
--- a/src/lib/spinner.test.ts
+++ b/src/lib/spinner.test.ts
@@ -1,6 +1,5 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
-import yoctoSpinnerFactory from 'yocto-spinner'
-import { resetGlobalArgs } from './global-args.js'
+import { describe, expect, it } from 'vitest'
+
 import {
     LoadingSpinner,
     resetEarlySpinner,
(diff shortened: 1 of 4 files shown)


src/lib/spinner.test.ts
```diff
@@ -9,346 +8,36 @@ import {
...
-    text: '',
-}
-
-vi.mock('yocto-spinner', () => ({
-    default: vi.fn(() => mockSpinnerInstance),
-}))
-
-// Mock chalk to avoid colors in tests
-vi.mock('chalk')
+// Smoke tests for the local wrapper. The detailed spinner behaviour
```
doistbot, line 11: [P2] `@doist/cli-core` cannot cover the `shouldDisableSpinner` callback that this wrapper injects from `global-args.ts`, so dropping all `--json` / `--no-spinner` / `TD_SPINNER=false` cases leaves the only repo-specific behavior unverified. Keep at least one integration test here that exercises the disable path through `withSpinner` or `startEarlySpinner`.

doistbot (commented):
This PR cleanly delegates the spinner implementation to `@doist/cli-core`, replacing the local module and removing redundant dependencies. It is a great cleanup that deletes significant boilerplate while fixing the doubled success and error glyphs. Regarding the test trims, since the `shouldDisableSpinner` logic remains specific to this repository, retaining an integration test for the disable paths (such as `--json` or `--no-spinner`) will ensure this local configuration continues to be verified.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F314&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aef7c20c0-498a-11f1-97ff-9f3c73e72ca9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aef7c20c0-498a-11f1-97ff-9f3c73e72ca9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/spinner.test.ts
scottlovegrove: Fixed in 3c9531f — added an integration test that sets `TD_SPINNER=false` in an otherwise interactive (`isTTY=true`, `CI` cleared) context and asserts `startEarlySpinner` doesn't install its stdout interceptor. That's the only observable side-effect that proves todoist's `shouldDisableSpinner` was called and its `true` reached cli-core's `isDisabled` gate. Picked `TD_SPINNER` deliberately because cli-core can't disable on it — only our wiring can.

PR #319, opened 2026-05-09
refactor: integrate @doist/cli-core 0.6.0 (markdown + changelog)

scottlovegrove:
## Summary

- **Markdown rendering** — `src/lib/markdown.ts` now delegates `preloadMarkdown` / `renderMarkdown` to `@doist/cli-core/markdown`, keeping the local leading-`* ` bullet escape (uncompletable task prefix) on top. Drops the local `marked-terminal-renderer` ambient `.d.ts` shim now that cli-core owns the typing.
- **Changelog command** — `src/commands/changelog.ts` is now a thin wrapper over `registerChangelogCommand` from `@doist/cli-core/commands`, passing `path`, `repoUrl`, and `version`. cli-core defaults match prior behavior (heading 2, `*` bullets, no continuation indent, no empty-version filtering). Local test deleted — cli-core covers the behavior.
- **Top-level error handler** — `src/index.ts` now matches `instanceof BaseCliError` (cli-core class) so errors thrown from cli-core commands route through the existing `formatError` path. The local `CliError` extends base, so subclass throws still match. `formatError` / `formatErrorJson` widened accordingly.
- Bumps `@doist/cli-core` from `0.5.0` to `0.6.0`.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm test` (1578 tests, 6 changelog tests removed alongside the local impl — covered by cli-core)
- [x] Smoke `td changelog -n 1` — output unchanged
- [x] Smoke `td changelog -n abc` — error routes via cli-core `CliError(INVALID_TYPE)` through todoist's top-level handler


src/lib/markdown.ts
```diff
@@ -1,22 +1,19 @@
...
-    instance.use(createTerminalRenderer(darkTheme()))
-    markedInstance = instance
+    if (preloaded) return
+    await corePreloadMarkdown()
+    preloaded = true
 }
 
 export async function renderMarkdown(text: string): Promise<string> {
-    if (!markedInstance) return text
+    if (!preloaded) return text
```
doistbot: [P2] This short-circuit adds a second source of truth for markdown readiness on top of `@doist/cli-core/markdown`, so the wrapper is no longer a thin delegate. Reuse the core module’s own preload state (or cache the shared preload promise here and always call `coreRenderMarkdown`) so local and core readiness logic can’t drift.

src/index.ts
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+import { CliError as BaseCliError } from '@doist/cli-core'
```
doistbot: [P2] This new direct `@doist/cli-core` import duplicates cli-core-specific error knowledge that’s also being added in `src/lib/output.ts`, even though `src/lib/errors.ts` is already the project’s error abstraction. Re-export the base error type/guard from `lib/errors.ts` and have both callers reuse that single local entry point.

src/commands/changelog.test.ts
```diff
@@ -1,155 +0,0 @@
-import { Command } from 'commander'
```
doistbot, line 1: [P2] Removing this file drops the only repo-level coverage for the Todoist-specific changelog integration: the local wrapper’s path/version/repo wiring and its interaction with this CLI’s top-level error handling. cli-core’s own tests won’t catch a broken local integration, so keep a slim integration test here instead of deleting coverage entirely.

src/commands/changelog.ts
```diff
@@ -1,109 +1,15 @@
...
-    }
-}
 
 export function registerChangelogCommand(program: Command): void {
-    program
-        .command('changelog')
-        .description('Show recent changelog entries')
-        .option('-n, --count <number>', 'Number of versions to show', '5')
-        .action(changelogAction)
+    registerCoreChangelogCommand(program, {
```
doistbot, line 10: [P2] This now routes `td changelog` through cli-core’s markdown-based rendering, but `src/index.ts` still treats `changelog` as a `noMarkdownCommands` case. As a result the renderer no longer preloads in parallel with the command import, so every `td changelog` run pays that setup cost serially. Remove `changelog` from the skip set (or explicitly preload here) to keep the previous fast path.

src/lib/output.ts
```diff
@@ -391,10 +397,10 @@ function resolveErrorParts(
     }
 }
 
-export function formatError(error: CliError): string
+export function formatError(error: AnyCliError): string
```
doistbot, line 396: [P2] This adds a new formatter path for `@doist/cli-core` error objects, but `output.test.ts` still only exercises the string overload. Add a case with a real cli-core `CliError` instance (ideally covering both `formatError` and `formatErrorJson`) so this integration does not regress silently.

src/index.ts
```diff
@@ -289,7 +290,7 @@ initializeLogger()
 program
     .parseAsync()
     .catch((err: Error) => {
-        if (err instanceof CliError) {
+        if (err instanceof BaseCliError) {
```
doistbot, line 292: [P2] This broadens the CLI’s user-facing error contract from the repo-local `src/lib/errors.ts` wrapper to any `@doist/cli-core` `CliError`, but the checked-in repository guidance still documents the local wrapper as the required path. Please either translate shared-package errors back to the local `CliError` or update the guidance/docs in this PR so the standard matches the implementation.

src/types/marked-terminal-renderer.d.ts
```diff
@@ -1,16 +0,0 @@
-// The marked-terminal-renderer package ships type declarations in
```
doistbot, line 1: [P3] Removing this shim also removes the only file under `src/types`, but `CODEBASE.md` still lists `src/types/marked-terminal-renderer.d.ts` in the repo map. Please update the map in this PR so the documented structure stays accurate.

doistbot (commented):
This PR successfully integrates @doist/cli-core 0.6.0 to standardize markdown rendering and the changelog command while thoughtfully reducing local boilerplate. Transitioning these features to the shared package improves long-term maintainability by letting the core module handle the heavy lifting. A few adjustments are needed to fully align the integration, such as centralizing the new core error types within our local abstraction, preserving slim integration coverage for the changelog wiring, synchronizing markdown preload behavior, and updating both test coverage and repository documentation to reflect the structural changes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F319&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9d9daf10-4b80-11f1-92a5-fabb3fbc47ca%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9d9daf10-4b80-11f1-92a5-fabb3fbc47ca&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/markdown.ts
scottlovegrove: Fixed in 4dd6e1a — replaced the dual flag with a cached preload promise (`preloadPromise`). Single source of truth for readiness; `renderMarkdown` short-circuits when no preload was ever requested (preserves the original semantics) and otherwise awaits the cached promise before delegating to `coreRenderMarkdown`.

src/index.ts
scottlovegrove: Fixed in 4dd6e1a — `src/lib/errors.ts` now re-exports `BaseCliError` from `@doist/cli-core` as a value export, and both `src/index.ts` and `src/lib/output.ts` import it from there. Single local entry point for the cli-core error class.

src/commands/changelog.test.ts
scottlovegrove: Fixed in 4dd6e1a — re-added a slim `src/commands/changelog.test.ts` that mocks `node:fs/promises`, registers the wrapper, and asserts the wrapper passes the local `CHANGELOG.md` path through and that the footer link uses `Doist/todoist-cli` + `packageJson.version`. cli-core covers the parsing/formatting behavior; this covers our wiring.

src/commands/changelog.ts
scottlovegrove: cli-core's `registerChangelogCommand` doesn't render via `marked` / `marked-terminal-renderer` — its `formatForTerminal` styles output with `chalk` directly (bold green for version headings, dimmed bullets, inline bold/code via `chalk.bold`/`chalk.cyan`). So the `noMarkdownCommands` skip set entry is still correct: `td changelog` doesn't pay any markdown-renderer setup cost. No change needed.

src/lib/output.ts
scottlovegrove: Fixed in 4dd6e1a — added two cases in `src/lib/output.test.ts`: `formatError` against a real `BaseCliError` instance (asserting code + message + hints render) and `formatErrorJson` against another instance (asserting the JSON shape with `code`/`message`/`hints`).

src/index.ts
scottlovegrove: Fixed in 4dd6e1a. Both legs:\n- `BaseCliError` is now re-exported from `src/lib/errors.ts`, so the local abstraction owns the cli-core type/class — no consumer reaches into `@doist/cli-core` directly.\n- `AGENTS.md` and `CODEBASE.md` updated to note that the global `parseAsync().catch` also matches the re-exported `BaseCliError`, so errors thrown by `@doist/cli-core` helpers route through the same formatter as the local `CliError` subclass.

src/types/marked-terminal-renderer.d.ts
scottlovegrove: Fixed in 4dd6e1a — dropped the `src/types/` entry from the repo map in `CODEBASE.md` (the directory is gone now).

rfgamaral (commented):
✅

PR #320, opened 2026-05-09
fix(deps): bump @doist/todoist-sdk to 10.1.3 for Node 26 gzip fix

### LINKED ISSUE #318: Node 26 / undici 7 / gziped responses bug
I installed todoist-cli today and could not get it to work with `td auth login` failing after showing success in the browser and other unsuccessful API response parsing. I let claude code diagnose the issue and below is what it came up with.

You can probably take it from here with your own claude.  Let me know if you need anything else from me (the human) or if such bug reports are not suitable here.

On Node 26 / undici 7, `@doist/todoist-cli@1.62.0` cannot parse any gzip-encoded
response from the Todoist API. Every command that calls a gzipped endpoint fails
with a Zod validation error against the response shape, even though the API
returned `200 OK` with valid JSON in the body.

The cause is in `@doist/todoist-sdk` (currently `10.1.2`): it installs an
`EnvHttpProxyAgent` from undici as the dispatcher for every `fetch()` call, and
on Node 26 that agent does not include the gzip-decompression interceptor that
Node's default `fetch` dispatcher uses. Compressed bytes reach `JSON.parse`,
parsing fails silently, and the SDK's `try/catch` falls back to assigning the
raw text to `response.data`. Subsequent destructuring (`data.results`,
`data.fullName`, …) produces `undefined`, which Zod then rejects at the root.

scottlovegrove:
## Summary

Bumps `@doist/todoist-sdk` from `10.1.1` to `10.1.3`, picking up [Doist/todoist-sdk-typescript#599](https://github.com/Doist/todoist-sdk-typescript/pull/599).

That SDK release composes undici's `interceptors.decompress()` onto its `EnvHttpProxyAgent` so gzipped response bodies are decoded before reaching the CLI. On Node 24+, attaching any custom dispatcher to global `fetch` strips `content-encoding` from the response but doesn't actually decompress the body — every `td` command that hits a gzipped endpoint was failing with a Zod root-level `expected: object, received: string` error. The CLI's `createTrackedFetch` reuses the SDK's dispatcher transparently, so no CLI code change is needed; only the version bump.

Closes #318.

## Test plan

- [x] `npm install && td auth status` on Node 26.1.0 — prints user (was failing)
- [x] `td inbox` — lists tasks (was failing)
- [x] `td auth status 2>&1 | grep -i experimental` — empty (the SDK suppresses undici's one-time `ExperimentalWarning`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   ],
   "dependencies": {
     "@doist/cli-core": "0.6.0",
-    "@doist/todoist-sdk": "10.1.1",
+    "@doist/todoist-sdk": "10.1.3",
     "@napi-rs/keyring": "1.3.0",
     "@pnpm/tabtab": "0.5.4",
     "chalk": "5.6.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@doist/cli-core": "0.6.0",
-        "@doist/todoist-sdk": "10.1.1",
+        "@doist/todoist-sdk": "10.1.3",
         "@napi-rs/keyring": "1.3.0",
         "@pnpm/tabtab": "0.5.4",
         "chalk": "5.6.2",
@@ -169,9 +169,9 @@
       }
     },
     "node_modules/@doist/todoist-sdk": {
-      "version": "10.1.1",
-      "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-10.1.1.tgz",
-      "integrity": "sha512-V60AjoJG5QjiJ2iYB8IIC4hLA4B5PLUybYysc1GU2qC3xjN7Zy3juRukKo8PO+O1VMXeb8EACMri5U/0GPnLkA==",
+      "version": "10.1.3",
+      "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-10.1.3.tgz",
+      "integrity": "sha512-yFnNS5C7sXS+yWVmWF1zx+XZ0bMxyM8LMXWH6069Kjxnu/Nj2x4vrMCs2qSV/vfNThyCJ6Gw+4lc8r0CLsN/1Q==",
       "license": "MIT",
       "dependencies": {
         "camelcase": "6.3.0",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -52,7 +52,7 @@
   ],
   "dependencies": {
     "@doist/cli-core": "0.6.0",
-    "@doist/todoist-sdk": "10.1.1",
+    "@doist/todoist-sdk": "10.1.3",
```
doistbot, line 55: [P2] This bump is specifically fixing a Node 24+/26-only regression, but the automated test workflows still execute on `.nvmrc` (currently Node 20) and this PR adds no regression coverage for the affected runtime. Please add CI coverage for at least one supported Node 24+/26 job so future SDK bumps can catch this gzip/dispatcher failure automatically.

doistbot (commented):
This PR bumps the @doist/todoist-sdk dependency to 10.1.3, successfully resolving a gzip decompression bug affecting users on Node 24 and 26. This is a clean update that seamlessly restores CLI functionality for modern runtimes without requiring custom workarounds. To prevent similar runtime-specific regressions from slipping through in the future, expanding the CI matrix to include a Node 24 or 26 test job would be a valuable addition.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F320&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad91e1570-4b8f-11f1-8e2d-f7a4e434830a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad91e1570-4b8f-11f1-8e2d-f7a4e434830a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

package.json
scottlovegrove: Since it's an SDK issue, I think that's fine.

PR #327, opened 2026-05-11
fix(task): add --no-labels flag to clear labels on update

### LINKED ISSUE #326: td task update --labels "" always returns HTTP 400 (CLI sends empty body)
## Summary

`td task update <ref> --labels ""` always fails with `HTTP 400`. The bug is **client-side**, not server-side: the CLI's truthy guard at `src/commands/task/update.ts:47` treats the empty string as "labels flag absent" and never sets `args.labels`. With no other field flags passed, the SDK then sends an empty body `{}`, which the server (correctly) rejects.

This is structurally similar to `#316` (now fixed), but unrelated in cause. The server already accepts `{"labels": []}` on the task update endpoint just fine. Only the CLI mishandles `--labels ""`.

## Reproduce (CLI, td 1.62.0)

```
$ td task add "zzz-labels-test" --labels "deep,quick"
Created: zzz-labels-test
ID: 6gcg5X49F6XJWjV4

$ td task update 6gcg5X49F6XJWjV4 --labels ""
Error: API_ERROR
HTTP 400: Bad Request
$ echo $?
1

$ td task view 6gcg5X49F6XJWjV4 --json | jq '.labels'
[
  "deep",
  "quick"
]
```

Labels are unchanged after the failed call. Same result against any task with one or more labels.

## Reproduce (instrumented wire payload + response body)

Wrapped `globalThis.fetch` to log both POST request body and the server's response:

```js
// /tmp/td-fetch-trace.mjs
const orig = globalThis.fetch;
globalThis.fetch = async (url, init = {}) => {
    const method = init.method ?? 'GET';
    const isTaskPost = method === 'POST' && url.toString().includes('/api/v1/tasks/');
    if (isTaskPost) {
        process.stderr.write(`[FETCH-TRACE] ${method} ${url}\n`);
        process.stderr.write(`[FETCH-TRACE] req-body=${init.body ?? '<none>'}\n`);
    }
    const res = await orig(url, init);
    if (isTaskPost) {
        const text = await res.clone().text();
        process.stderr.write(`[FETCH-TRACE] resp-status=${res.status}\n`);
        process.stderr.write(`[FETCH-TRACE] resp-body=${text}\n`);
    }
    return res;
};
```

Run:

```
$ NODE_OPTIONS="--import file:///tmp/td-fetch-trace.mjs" td task update 6gcgghgpWjGgGWr4 --labels ""

[FETCH-TRACE] POST https://api.todoist.com/api/v1/tasks/6gcgghgpWjGgGWr4
[FETCH-TRACE] req-body={}
[FETCH-TRACE] resp-status=400
[FETCH-TRACE] resp-body={"error":"At least one of supported fields should be set and non-empty","error_code":42,"error_extra":{"event_id":"397d22e72c424083898861299532961f","retry_after":2},"error_tag":"BAD_REQUEST","http_code":400}

Error: API_ERROR
HTTP 400: Bad Request
```

Two facts captured directly:

1. The literal wire payload is `{}` — not `{"labels": []}`, not `{"labels": null}`.
2. The server's response is `"At least one of supported fields should be set and non-empty"` — same string returned to a hand-rolled curl `POST` with body `{}` (probe F below).

The CLI swallows the response body and only surfaces `HTTP 400: Bad Request`, which is why this took some digging.

## Reproduce (raw REST)

Same OAuth token, same task endpoint, four probes:

```
--- Probe B: empty array accepted ---
POST https://api.todoist.com/api/v1/tasks/6gcg5X49F6XJWjV4
body: {"labels": []}

HTTP 200 — task returned with "labels": []

--- Probe E: empty-string array accepted (server normalizes) ---
body: {"labels": [""]}
HTTP 200 — response shows "labels": []

--- Probe F: empty body rejected ---
body: {}
HTTP 400
{"error":"At least one of supported fields should be set and non-empty","error_code":42,"error_extra":{"event_id":"5081e1d2f0444022a4d4b70797280c3c","retry_after":3},"error_tag":"BAD_REQUEST","http_code":400}

--- Probe G: null rejected ---
body: {"labels": null}
HTTP 400
{"error":"labels must be an array of labels","error_code":42,...}
```

Probe F's error string matches the CLI failure exactly — confirming the CLI sends `{}`.

## Root cause

`src/commands/task/update.ts:47` (and the compiled `dist/commands/task/update.js:30`):

```ts
if (options.labels) args.labels = options.labels.split(',').map((l) => l.trim())
```

`options.labels === ""` is falsy in JS, so `args.labels` is never set. When `--labels ""` is the only flag, the surrounding `args` object stays `{}` and the SDK's `updateTask(id, {})` POSTs an empty body.

Note the same file already handles "clear this field" correctly for siblings — `src/commands/task/update.ts` lines 19–28:

```ts
if (options.due === false) {
    args.dueString = null
} else if (options.due) {
    args.dueString = options.due
}

if (options.deadline === false) {
    args.deadlineDate = null
} else if (options.deadline) {
    args.deadlineDate = options.deadline
}
```

backed by `--no-due` / `--no-deadline` declarations in `src/commands/task/index.ts:183,185`. Labels has no equivalent `--no-labels` declaration and no clearing path.

## Suggested fix

Two reasonable shapes; preference for **(a)**:

**(a) Add `--no-labels` flag (mirrors `--no-due` / `--no-deadline`)** — cleanest UX, consistent with sibling clear-flags:

```ts
// src/commands/task/index.ts (option block around line 192)
.option('--labels <a,b>', 'New labels (replaces existing)')
.option('--no-labels', 'Remove all labels')

// src/commands/task/update.ts (around line 47)
if (options.labels === false) {
    args.labels = []
} else if (options.labels) {
    args.labels = options.labels.split(',').map((l) => l.trim())
}
```

**(b) Minimal patch** — change the truthy guard so `""` passes through:

```ts
if (options.labels !== undefined) args.labels = options.labels.split(',').map((l) => l.trim())
```

`"".split(",").map(trim)` is `[""]`, and probe E shows the server normalizes `[""]` to `[]`. Works, but relies on server-side normalization and is less discoverable than `--no-labels`.

## Workaround (current users)

Direct REST clears labels reliably:

```bash
TOKEN=$(security find-generic-password -s "todoist-cli" -w)   # macOS keychain example
curl -sS -X POST "https://api.todoist.com/api/v1/tasks/<task-id>" \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"labels": []}'
```

## Versions

- `td` 1.62.0 (npm `@doist/todoist-cli`, dist/index.js)
- Endpoint: `POST https://api.todoist.com/api/v1/tasks/{id}`
- Upstream `main` `src/commands/task/update.ts:47` still has the same guard at the time of filing.
- macOS, Node 20+

## Related

- `#316` (closed, server-side fix shipped) — same shape of bug for `--no-favorite` on labels, but rooted in the *server* validator. This task-update bug is the mirror image: it lives in the *CLI*, and the server was never the problem.

scottlovegrove:
## Summary

- Fixes #326 — `td task update <ref> --labels ""` returned HTTP 400 because the truthy guard at `src/commands/task/update.ts:47` swallowed the empty string and the SDK posted an empty body `{}`.
- Adds `--no-labels` flag mirroring the existing `--no-due` / `--no-deadline` pattern; clearing now sends `{"labels": []}`, which the server already accepts.
- Updates the dry-run preview to show `(remove all)` when `--no-labels` is used.

## Test plan

- [x] `npm test` — 1548 tests pass, including new `task update --no-labels` test
- [ ] Manual smoke against real API:
  ```
  td task add "zzz-labels-test" --labels "deep,quick"
  td task update <id> --no-labels
  td task view <id> --json | jq '.labels'   # expect []
  ```
- [ ] `td task update <id> --no-labels --dry-run` shows `Labels: (remove all)`

## Audit

Scanned every `src/commands/*/update.ts` for the same truthy-guard-on-optional-string shape. Only `task --labels` was an actually-fixable CLI-side bug. `description` and `duration` use a similar guard but the SDK type does not currently accept `null` / clearing, so they are out of scope here.

### DIFF
--- a/src/commands/task/index.ts
+++ b/src/commands/task/index.ts
@@ -190,6 +190,7 @@ Examples:
             ),
         )
         .option('--labels <a,b>', 'New labels (replaces existing)')
+        .option('--no-labels', 'Remove all labels')
         .option('--description <text>', 'New description')
         .option('--stdin', 'Read task description from stdin')
         .option('--assignee <ref>', 'Assign to user (name, email, id:xxx, or "me")')
--- a/src/commands/task/task.test.ts
+++ b/src/commands/task/task.test.ts
@@ -1671,6 +1671,35 @@ describe('task update --no-due', () => {
     })
 })
 
+describe('task update --no-labels', () => {
+    let mockApi: MockApi
+
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockApi = createMockApi()
+        mockGetApi.mockResolvedValue(mockApi)
+    })
+
+    it('removes all labels with --no-labels', async () => {
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+
+        mockApi.getTask.mockResolvedValue({
+            id: 'task-1',
+            content: 'Task',
+            labels: ['deep', 'quick'],
+        })
+        mockApi.updateTask.mockResolvedValue({ id: 'task-1', content: 'Task' })
+
+        await program.parseAsync(['node', 'td', 'task', 'update', 'id:task-1', '--no-labels'])
+
+        expect(mockApi.updateTask).toHaveBeenCalledWith('task-1', {
+            labels: [],
+        })
+        consoleSpy.mockRestore()
+    })
+})
+
 describe('task add/update --uncompletable', () => {
     let mockApi: MockApi
 
--- a/src/commands/task/task.test.ts
+++ b/src/commands/task/task.test.ts
@@ -1698,6 +1698,31 @@ describe('task update --no-labels', () => {
         })
         consoleSpy.mockRestore()
     })
+
+    it('previews --no-labels with --dry-run without calling API', async () => {
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+
+        mockApi.getTask.mockResolvedValue({
+            id: 'task-1',
+            content: 'Task',
+            labels: ['deep', 'quick'],
+        })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'task',
+            'update',
+            'id:task-1',
+            '--no-labels',
+            '--dry-run',
+        ])
+
+        expect(mockApi.updateTask).not.toHaveBeenCalled()
+        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining('(remove all)'))
+        consoleSpy.mockRestore()
+    })
 })
 
 describe('task add/update --uncompletable', () => {
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -143,7 +143,7 @@ Choosing between `task add` and `task quickadd`:
 Useful task flags:
 - `--stdin` on `task add` reads the task description from stdin; on `task quickadd` (and the top-level `td add`) it reads the full natural-language text from stdin.
 - `--parent`, `--section`, `--project`, `--workspace`, `--assignee`, `--labels`, `--due`, `--deadline`, `--duration`, and `--priority` cover most task workflows.
-- `td task complete --forever` stops recurrence; `td task update --no-due` clears the due date and `--no-deadline` clears deadlines; `td task move --no-parent` and `--no-section` detach from hierarchy.
+- `td task complete --forever` stops recurrence; `td task update --no-due` clears the due date, `--no-deadline` clears deadlines, and `--no-labels` removes all labels; `td task move --no-parent` and `--no-section` detach from hierarchy.
 
 ### Projects And Workspaces
 ```bash
--- a/src/commands/task/update.ts
+++ b/src/commands/task/update.ts
@@ -13,7 +13,7 @@ export interface UpdateOptions {
     due?: string | false
     deadline?: string | false
     priority?: string
-    labels?: string
+    labels?: string | false
     description?: string
     stdin?: boolean
     assignee?: string
@@ -44,7 +44,11 @@ export async function updateTask(ref: string, options: UpdateOptions): Promise<v
         args.deadlineDate = options.deadline
     }
     if (options.priority) args.priority = parsePriority(options.priority)
-    if (options.labels) args.labels = options.labels.split(',').map((l) => l.trim())
+    if (options.labels === false) {
+        args.labels = []
+    } else if (options.labels) {
+        args.labels = options.labels.split(',').map((l) => l.trim())
+    }
 
     if (options.stdin && options.description !== undefined) {
         throw new CliError('CONFLICTING_OPTIONS', 'Cannot use both --description and --stdin')
@@ -89,7 +93,7 @@ export async function updateTask(ref: string, options: UpdateOptions): Promise<v
             Due: args.dueString ?? undefined,
             Deadline: args.deadlineDate ?? undefined,
             Priority: options.priority,
-            Labels: options.labels,
+            Labels: options.labels === false ? '(remove all)' : options.labels,
             Assignee: options.assignee,
             Unassign: options.unassign ? 'yes' : undefined,
             Duration: options.duration,
--- a/src/lib/skills/content.ts
+++ b/src/lib/skills/content.ts
@@ -139,7 +139,7 @@ Choosing between \`task add\` and \`task quickadd\`:
 Useful task flags:
 - \`--stdin\` on \`task add\` reads the task description from stdin; on \`task quickadd\` (and the top-level \`td add\`) it reads the full natural-language text from stdin.
 - \`--parent\`, \`--section\`, \`--project\`, \`--workspace\`, \`--assignee\`, \`--labels\`, \`--due\`, \`--deadline\`, \`--duration\`, and \`--priority\` cover most task workflows.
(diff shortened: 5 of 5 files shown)


doistbot (commented):
This PR introduces a helpful `--no-labels` flag to cleanly remove labels during task updates, along with an updated dry-run preview. This correctly resolves the HTTP 400 errors users encountered when trying to clear labels with an empty string. A few minor adjustments have been flagged, specifically documenting the new flag in the skill content, adding test coverage for the dry-run output, and updating the test mocks to rely on shared repository fixtures.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F327&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9e9352c0-4d7a-11f1-89b7-5e73f060b815%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9e9352c0-4d7a-11f1-89b7-5e73f060b815&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/task/index.ts
```diff
@@ -190,6 +190,7 @@ Examples:
             ),
         )
         .option('--labels <a,b>', 'New labels (replaces existing)')
+        .option('--no-labels', 'Remove all labels')
```
doistbot, line 193: [P0] Per `AGENTS.md`, whenever flags or options are added in `src/commands/`, the `SKILL_CONTENT` in `src/lib/skills/content.ts` must be updated to match. Please document the new `--no-labels` flag there (similar to `--no-due`) and run `npm run sync:skill`.

src/commands/task/task.test.ts
```diff
@@ -1671,6 +1671,35 @@ describe('task update --no-due', () => {
...
+describe('task update --no-labels', () => {
+    let mockApi: MockApi
+
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockApi = createMockApi()
+        mockGetApi.mockResolvedValue(mockApi)
+    })
+
+    it('removes all labels with --no-labels', async () => {
```
doistbot, line 1683: [P2] This only exercises the mutating path. The diff also adds a new dry-run branch for `--no-labels` (`Labels: (remove all)`), but there’s no assertion for that output anywhere, so the user-visible behavior described in the PR can regress unnoticed. Please add a `task update ... --no-labels --dry-run` case that verifies `updateTask` is not called and the preview includes `(remove all)`.

src/commands/task/task.test.ts
```diff
@@ -1671,6 +1671,35 @@ describe('task update --no-due', () => {
...
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+
+        mockApi.getTask.mockResolvedValue({
```
doistbot, line 1687: [P3] This new test is hand-building SDK task payloads, but the repo's test conventions say command tests should use the shared fixtures from `src/test-support/fixtures.ts` instead of ad-hoc entities. Please switch these mocks to fixture-derived data so they stay aligned with the canonical test shapes.

src/commands/task/index.ts
scottlovegrove: Done in 9c15846 — extended the clear-flags hint in `SKILL_CONTENT` to mention `--no-labels` alongside `--no-due` / `--no-deadline`, and resynced `skills/todoist-cli/SKILL.md` via `npm run sync:skill`. `npm run check:skill-sync` is now green.

src/commands/task/task.test.ts
scottlovegrove: Done in 9c15846 — added a `previews --no-labels with --dry-run without calling API` test that asserts `updateTask` is not called and the console output contains `(remove all)`.

src/commands/task/task.test.ts
scottlovegrove: Holding off on this one. `src/commands/task/task.test.ts` doesn't import from `src/test-support/fixtures.ts` anywhere — the immediate sibling tests for `--no-due` (line 1654) and `--no-deadline` (line 1625) hand-build the same minimal `{ id, content, labels|due|deadline }` shape that this new test uses. Switching only the `--no-labels` test to fixtures would diverge from the local pattern; if we want to adopt fixtures here it's probably better as a broader file-wide cleanup PR. Happy to do that separately if you'd like.

PR #330, opened 2026-05-12
feat(project): add `--parent` on update and new `reorder` command

scottlovegrove:
## Summary
Brings parity with the MCP server's project re-parenting and reordering capabilities flagged at https://github.com/Doist/todoist-ai/issues/475#issuecomment-4426034538.

- `td project update --parent <ref>` / `--no-parent` re-nests a personal project under another personal parent (or moves to top level) via the sync `project_move` command. The SDK's `updateProject` does not accept `parentId`, so this uses the sync API directly — mirroring how the MCP `reorder-objects` tool does it.
- `td project reorder <ref>` is a new command with `--before <ref>` / `--after <ref>` / `--position <n>` (relative flags, one at a time) that reorders a personal project among its siblings via the sync `project_reorder` command. Sends a contiguous renumber of the full sibling list for predictability.

Workspace projects are rejected for both flows — they don't support sub-projects, and their sidebar order is workspace-scoped (different semantics, out of scope here).

Also takes the opportunity to extract the existing `project update` and new `project reorder` describes out of the ~2900-line `project.test.ts` into `update.test.ts` and `reorder.test.ts` siblings.

## Test plan
- [ ] `npm run type-check` clean
- [ ] `npm run check` clean
- [ ] `npm test` passes (28 new tests across update --parent and reorder)
- [ ] Manual smoke against a real account:
  - `td project create --name "Plan-Parent-A"`, `"Plan-Parent-B"`, `"Plan-Child"`
  - `td project update "Plan-Child" --parent "Plan-Parent-A"` → verify nesting in app
  - `td project update "Plan-Child" --parent "Plan-Parent-B"` → re-parent
  - `td project update "Plan-Child" --no-parent` → top level
  - `td project reorder "Plan-Child" --position 0` → verify
  - `td project reorder "Plan-Child" --before "Plan-Parent-B"` → verify
  - `td project reorder "Plan-Child" --after "Plan-Parent-A"` → verify
  - Cleanup with `td project delete`


src/commands/project/update.ts
```diff
@@ -50,7 +52,36 @@ export async function updateProject(ref: string, options: UpdateOptions): Promis
...
+            throw new CliError(
+                'WORKSPACE_NO_SUBPROJECTS',
+                'Workspace projects do not support sub-projects.',
+                ['Sub-projects are only supported for personal projects.'],
+            )
+        }
+        if (options.parent === false) {
+            newParentId = null
+        } else {
+            const parentProject = await resolveProjectRef(api, options.parent)
```
doistbot: [P2] This only rejects self-parenting; it still allows choosing any descendant as the new parent. In a tree like `A > B`, `td project update A --parent B` gets through to `project_move`, which will be invalid, and if other update fields were included they have already been persisted by `api.updateProject()` first. Please walk the proposed parent's ancestry (or the target's subtree) and reject descendant parents before setting `newParentId`.

src/commands/project/reorder.ts
```diff
@@ -0,0 +1,130 @@
...
+        })
+
+    const oldIndex = siblings.findIndex((p) => p.id === target.id)
+    if (oldIndex === -1) {
+        throw new CliError('NOT_FOUND', 'Target project not found among siblings.')
+    }
+
+    let newIndex: number
+    if (options.position !== undefined) {
+        const parsed = Number.parseInt(options.position, 10)
```
doistbot: [P2] `Number.parseInt()` is too permissive for a flag documented as a non-negative integer. Inputs like `1.5`, `2abc`, or `0x10` will pass this check and silently reorder to an unintended position. Validate the entire string before converting (for example with `/^\d+$/` or `Number.isInteger(Number(...))`).

src/commands/project/update.ts
```diff
@@ -50,7 +52,36 @@ export async function updateProject(ref: string, options: UpdateOptions): Promis
         }
     }
 
-    if (Object.keys(args).length === 0) {
+    let newParentId: string | null | undefined
+    let newParentName: string | undefined
+    if (options.parent !== undefined) {
```
doistbot, line 58: [P2] This parent-handling block reimplements the same parent-project resolution/validation that `project/create.ts` already has (`resolveProjectRef`, reject workspace parents, raise `WORKSPACE_NO_SUBPROJECTS`). The two paths are already starting to drift in wording, so future hierarchy-rule changes will need to be fixed twice. Please extract a shared helper for resolving a personal parent project and reuse it from both create and update.

doistbot (commented):
This PR nicely brings the CLI into parity with the MCP server by introducing project re-parenting and sibling reordering capabilities. These additions provide powerful new ways to manage project hierarchies, though the implementations present a few opportunities to optimize API usage and tighten data handling. Some adjustments are needed to correct the REST API mapping for sibling ordering, strengthen validation against descendant parent cycles and invalid position inputs, consolidate redundant network requests and duplicate helper logic, and ensure the new JSON outputs and skill content updates are fully covered.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F330&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae1080040-4dea-11f1-9c7a-4bcd9957325f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae1080040-4dea-11f1-9c7a-4bcd9957325f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/project/index.ts
```diff
@@ -194,6 +197,30 @@ Examples:
             return moveProject(ref, options)
         })
 
+    const reorderCmd = project
+        .command('reorder [ref]')
+        .description('Reorder a personal project among its siblings')
+        .option('--before <ref>', 'Place before this sibling (name or id:xxx)')
+        .option('--after <ref>', 'Place after this sibling (name or id:xxx)')
+        .option('--position <n>', 'Move to a 0-indexed position among siblings')
```
doistbot: [P2] `--position` is another non-negative order flag, but this command skips the existing order-parsing pattern we already use for folder order options and instead re-parses/validates the string later in `reorder.ts`. That has already drifted in behavior (`INVALID_ORDER` + hint there vs `INVALID_OPTIONS` here). Please extract/reuse a shared order parser and wire it into this option so all order-style flags validate consistently.

src/commands/project/reorder.ts
```diff
@@ -0,0 +1,130 @@
...
+        (cursor, limit) => api.getProjects({ cursor: cursor ?? undefined, limit }),
+        { limit: Number.MAX_SAFE_INTEGER, startCursor: undefined },
+    )
+
+    const targetParentId = (target as { parentId?: string | null }).parentId ?? null
+    const siblings: Project[] = allProjects
+        .filter((p) => !isWorkspaceProject(p))
+        .filter((p) => ((p as { parentId?: string | null }).parentId ?? null) === targetParentId)
+        .sort((a, b) => {
+            const ao = (a as { childOrder?: number }).childOrder ?? 0
```
doistbot: [P1] The REST API (`getProjects`) exposes the sibling order as `order`, not `childOrder` (which is the key used in the Sync API payload). At runtime, `childOrder` will be undefined, causing all projects to evaluate to `0` and fallback to their array index order, which breaks the sort predictability. Use `a.order ?? 0` (and `b.order`) instead, and update the mock objects in `reorder.test.ts` to match.

src/commands/project/update.ts
```diff
@@ -50,7 +52,36 @@ export async function updateProject(ref: string, options: UpdateOptions): Promis
...
+        if (isWorkspaceProject(project)) {
+            throw new CliError(
+                'WORKSPACE_NO_SUBPROJECTS',
+                'Workspace projects do not support sub-projects.',
```
doistbot: [P3] The error message "Workspace projects do not support sub-projects" is slightly confusing when the user is trying to set the `--parent` of a workspace project (i.e. trying to make *it* a sub-project). Consider rewording to something like `"Workspace projects cannot be sub-projects."` or `"Workspace projects cannot be nested."` to clarify.

src/commands/project/reorder.ts
```diff
@@ -0,0 +1,130 @@
...
+    const target = await resolveProjectRef(api, ref)
+
+    if (isWorkspaceProject(target)) {
+        throw new CliError(
+            'WORKSPACE_REORDER_UNSUPPORTED',
+            'Reorder is only supported for personal projects.',
+        )
+    }
+
+    const { results: allProjects } = await paginate(
```
doistbot: [P2] This command already paginates the full project list here, but the `resolveProjectRef()` calls above/below will do their own `getProjects()` lookup whenever the target or sibling is passed by name. In the common `reorder "A" --before "B"` flow that means multiple full-list scans for one operation. Please resolve both refs from `allProjects` (or extract a list-based resolver) so this command has a single source of truth.

src/commands/project/reorder.ts
```diff
@@ -0,0 +1,130 @@
...
+        (cursor, limit) => api.getProjects({ cursor: cursor ?? undefined, limit }),
+        { limit: Number.MAX_SAFE_INTEGER, startCursor: undefined },
+    )
+
+    const targetParentId = (target as { parentId?: string | null }).parentId ?? null
+    const siblings: Project[] = allProjects
```
doistbot: [P2] Typing this back to `Project[]` throws away the `!isWorkspaceProject()` narrowing and is why the rest of the function has to reach for structural casts like `{ parentId?: ... }` and `{ childOrder?: ... }`. Keep this collection narrowed to personal projects (or add a local type guard/helper) so the reorder logic stays type-checked against the SDK model.

src/commands/project/update.ts
```diff
@@ -66,17 +97,44 @@ export async function updateProject(ref: string, options: UpdateOptions): Promis
...
+    if (Object.keys(args).length > 0) {
+        const updated = await api.updateProject(project.id, args)
+        updatedName = updated.name
+        if (options.json && newParentId === undefined) {
+            console.log(formatJson(updated, 'project'))
+            return
+        }
+    }
+
+    if (newParentId !== undefined) {
```
doistbot, line 125: [P2] This always sends a `project_move` sync whenever `--parent`/`--no-parent` is present, even if the project is already under that parent (or already top level). That turns a no-op into an unnecessary write request and extra sync queue load. Compare against `project.parentId ?? null` first and skip the move when nothing actually changes.

src/commands/project/reorder.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    it('no-op when target already at requested position', async () => {
+        const program = createProgram()
+        mockApi.getProject.mockResolvedValueOnce(siblings[1])
+        mockApi.getProjects.mockResolvedValue({ results: siblings, nextCursor: null })
+
+        await program.parseAsync(['node', 'td', 'project', 'reorder', 'id:p-2', '--position', '1'])
+
+        expect(mockReorderProjects).not.toHaveBeenCalled()
+    })
+})
```
doistbot, line 304: [P2] The `reorder` command implements a `--json` flag to output the new sibling ordering, but there are no tests covering this behavior. Consider adding tests to verify that `--json` correctly outputs the expected array format and suppresses the standard log output (both for successful reorders and no-ops).

src/commands/project/update.test.ts
```diff
@@ -0,0 +1,396 @@
...
+            'id:p-1',
+            '--parent',
+            'id:p-2',
+            '--dry-run',
+        ])
+
+        expect(mockMoveProjectParent).not.toHaveBeenCalled()
+        expect(mockApi.updateProject).not.toHaveBeenCalled()
+    })
+})
```
doistbot, line 459: [P2] The PR introduces a specific code path to handle `--json` output when re-parenting (requiring an extra `resolveProjectRef` fetch since the sync API does not return the updated entity). This branch is currently untested. Consider adding a test case using both `--parent` and `--json` to verify the entity is fetched and the expected JSON payload is logged.

src/lib/skills/content.ts
```diff
@@ -152,6 +152,10 @@ td project create --name "New Project" --color blue
 td project update "Roadmap" --favorite
 td project update "Roadmap" --folder "Engineering"
 td project update "Roadmap" --no-folder
+td project update "Roadmap" --parent "Engineering"
+td project update "Roadmap" --no-parent
+td project reorder "Roadmap" --before "Marketing"
```
doistbot, line 159: [P2] The repo guidance requires `SKILL_CONTENT` to stay in sync with every added command/flag, but this update only documents `project reorder` via `--before`/`--position` examples. The shipped surface here also adds `--after`, `--json`, and `--dry-run` (plus the new `project update` parenting behavior), so the skill content is still incomplete. Please expand the project section in `content.ts` to cover the full new command/option surface, then regenerate `skills/todoist-cli/SKILL.md`.

src/commands/project/reorder.ts
scottlovegrove: Done in f635b80 — `--position` now uses the shared `parseOrderArg` from `src/lib/order.ts`, which validates with `Number.isInteger` and rejects `1.5`, `2abc`, `0x10`, negatives, etc. with `INVALID_ORDER`. Added a regression test in `reorder.test.ts`.

src/commands/project/update.ts
scottlovegrove: Done in f635b80 — extracted `resolvePersonalParent` to `src/commands/project/helpers.ts`. Both `create` and `update` now share that helper, so the workspace-parent guard and the wording live in one place.

src/commands/project/index.ts
scottlovegrove: Done in f635b80 — `parseOrderArg` is now in `src/lib/order.ts` and both `folder/index.ts` and `project/index.ts` import it. `--position` uses it as commander's `coerce`, so validation is consistent across all order-style flags (`INVALID_ORDER` with the same hint).

src/commands/project/update.ts
scottlovegrove: Done in f635b80 — `update --parent` now loads the personal project list once and walks the proposed parent's ancestry via the new `isDescendantOf` helper. If the proposed parent is a descendant of the target, we throw `INVALID_OPTIONS` *before* either `updateProject` or the sync `project_move` runs, so no partial metadata write.

src/commands/project/reorder.ts
scottlovegrove: Going to push back on this one. The `@doist/todoist-sdk` `PersonalProject` / `WorkspaceProject` TS types both expose `childOrder: number` (see `dist/types/types/projects/types.d.ts` — the Zod transform output has `childOrder`, not `order`). At runtime the REST `getProjects` response is parsed through that schema, so `p.childOrder` is the populated field — `p.order` would actually be `undefined`. The reorder tests confirm the sort comes out right with `childOrder`. I think this comment was reasoning from the Sync API command shape — which does use `childOrder` — but I'm reading the same name on the REST side too.

If I'm missing something here please flag back, but for now leaving as `childOrder`.

src/commands/project/update.ts
scottlovegrove: Done in f635b80 — the error users hit when trying to set `--parent` on a workspace project now reads "Workspace projects cannot be nested under another project." The parent-side message (when the target *parent* is a workspace project) is now "Workspace projects cannot be used as a parent." Both centralised in the new `resolvePersonalParent` helper / `update.ts` guard.

src/commands/project/reorder.ts
scottlovegrove: Done in f635b80 — pulled the project pagination into `loadPersonalProjects()` in `helpers.ts`. Reorder now calls it once, and the sibling `--before`/`--after` ref is resolved against that cached list via `resolvePersonalFromList()` — no extra `getProjects()` round trip. The target ref still goes through `resolveProjectRef` because we need it to surface the proper workspace-rejection error before we touch the personal cache.

src/commands/project/reorder.ts
scottlovegrove: Done in f635b80 — `siblings` is now typed as `PersonalProject[]` (and `loadPersonalProjects()` returns the narrowed type), so the `.parentId` / `.childOrder` accesses no longer need structural casts.

src/commands/project/update.ts
scottlovegrove: Done in f635b80 — now compare `project.parentId ?? null` against the resolved new parent before scheduling the sync. If they match, `newParentId` stays `undefined` and we hit the `NO_CHANGES` guard (when there are no other field changes) or just skip the `project_move`. Two regression tests in `update.test.ts`: one for `--parent` matching current parent, one for `--no-parent` on an already-top-level project.

src/commands/project/reorder.test.ts
scottlovegrove: Done in f635b80 — added two `--json` tests in `reorder.test.ts`: one that verifies the new ordering JSON payload (and asserts the standard log line is suppressed), and one that confirms the no-op path still emits the current ordering.

src/commands/project/update.test.ts
scottlovegrove: Done in f635b80 — added `"--parent --json refetches and emits the refreshed project as JSON"` in `update.test.ts`. It primes `getProject` for the initial target → parent → refreshed-target sequence, asserts `moveProjectParent` ran with the right args, and parses the emitted JSON to verify `parentId` reflects the move.

src/lib/skills/content.ts
scottlovegrove: Done in f635b80 — `src/lib/skills/content.ts` now covers `--after`, `--json`, and `--dry-run` for both `reorder` and `update --parent` (in addition to the previously documented `--before` / `--position` / `--parent` / `--no-parent`). Ran `npm run sync:skill` so `skills/todoist-cli/SKILL.md` matches.

gnapse (commented):
✅

PR #334, opened 2026-05-12
refactor(auth): integrate @doist/cli-core 0.10.0 (logout + status)

scottlovegrove:
## Summary

Bumps `@doist/cli-core` to **0.10.0** and migrates `td auth logout` + `td auth status` onto the new `attachLogoutCommand` / `attachStatusCommand` registrars that ship alongside the existing `attachLoginCommand` ([cli-core#16](https://github.com/Doist/cli-core/pull/16)).

- `TodoistTokenStore` gains `getLastClearResult()` so the logout `onCleared` callback can surface keyring-fallback warnings — cli-core's `TokenStore.clear: void` contract can't carry the `TokenStorageResult` directly.
- `auth/index.ts` now creates a single shared `TodoistTokenStore` instance used by login, logout, and status (login's wrapper takes the store as a parameter instead of constructing its own).
- `attachTodoistStatusCommand` routes both the active-snapshot path (`fetchLive`) and the unauthenticated path (`onNotAuthenticated`) through one `gatherStatusData` helper so env-token mode and `--user <ref>` (both return `null` from `TokenStore.active()` by the adapter's documented contract) render byte-for-byte identically to the persisted-default-user path.
- `auth logout` and `auth status` now accept `--ndjson` for free (framework-registered) on top of `--json`. `auth status` previously only had `--json`.
- `td auth token view` stays hand-rolled — its `--user <ref>` selector depends on multi-user state the `TokenStore` adapter intentionally drops from `active()`. cli-core PR #16 explicitly carves this out as a scope note pending a multi-user store contract.

## Test plan

- [x] `npm run check` (oxlint + oxfmt) clean
- [x] `npm test` — 1574/1574 across 60 files
- [x] `npm run build` clean
- [x] Smoke: `HOME=$(mktemp -d) td auth status` prints the legacy "No API token found." CliError envelope
- [x] Smoke: `td auth {login,logout,status} --help` show framework-registered `--json` / `--ndjson` flags
- [ ] Manual: full OAuth login → `td auth status` → `td auth status --json` → `td auth logout` round-trip against a real account
- [ ] Manual: `TODOIST_API_TOKEN=… td auth status` shows `✓ Authenticated (TODOIST_API_TOKEN)` env-mode branch
- [ ] Manual: multi-user setup shows the "Other stored accounts" enumeration

### DIFF
--- a/src/commands/auth/index.ts
+++ b/src/commands/auth/index.ts
@@ -1,14 +1,23 @@
 import type { Command } from 'commander'
+import { createTodoistTokenStore } from '../../lib/auth-store.js'
 import { attachTodoistLoginCommand } from './login.js'
-import { logout } from './logout.js'
-import { showStatus } from './status.js'
+import { attachTodoistLogoutCommand } from './logout.js'
+import { attachTodoistStatusCommand } from './status.js'
 import { viewToken } from './token-view.js'
 import { loginWithToken } from './token.js'
 
 export function registerAuthCommand(program: Command): void {
     const auth = program.command('auth').description('Manage authentication')
 
-    attachTodoistLoginCommand(auth)
+    // Shared store instance: login stashes the post-`set` storage result for
+    // its success handler, logout reads the post-`clear` result for the same
+    // keyring-fallback warning surface. Status uses `active()` as the
+    // authenticated-snapshot gate.
+    const store = createTodoistTokenStore()
+
+    attachTodoistLoginCommand(auth, store)
+    attachTodoistLogoutCommand(auth, store)
+    attachTodoistStatusCommand(auth, store)
 
     // `token` is a hybrid: it accepts a positional `[token]` (save) and also
     // exposes subcommands (`view`). Commander matches subcommand names before
@@ -16,6 +25,10 @@ export function registerAuthCommand(program: Command): void {
     // dispatches to the `view` subcommand — `view` is never treated as a
     // literal token value. Real Todoist tokens are 40-char hex strings, so
     // this disambiguation is safe in practice.
+    //
+    // `token view` stays hand-rolled (not migrated to `attachTokenViewCommand`)
+    // because it depends on the `--user <ref>` selector + env precedence rules
+    // that the `TokenStore` adapter intentionally drops from `active()`.
     const tokenCmd = auth
         .command('token [token]')
         .description('Save API token for CLI authentication (or use a subcommand: `view`)')
@@ -27,11 +40,4 @@ export function registerAuthCommand(program: Command): void {
             'Print the stored API token for the active user (or --user <ref>) to stdout for use in scripts',
         )
         .action(viewToken)
-
-    auth.command('status')
-        .description('Show current authentication status')
-        .option('--json', 'Output as JSON')
-        .action(showStatus)
-
-    auth.command('logout').description('Remove saved authentication token').action(logout)
 }
--- a/src/commands/auth/login.test.ts
+++ b/src/commands/auth/login.test.ts
@@ -37,6 +37,7 @@ vi.mock('@doist/cli-core/auth', async (importOriginal) => {
     }
 })
 
+import { createTodoistTokenStore } from '../../lib/auth-store.js'
 import { attachTodoistLoginCommand } from './login.js'
 
 type AttachOptions = {
@@ -53,22 +54,7 @@ function attachAndCapture(): AttachOptions {
     capturedAttachOptions.length = 0
     const program = new Command()
     program.exitOverride()
-    // Stub `TodoistTokenStore` — the login attacher just forwards the store to
-    // cli-core, which is mocked here. The tests only exercise the local
-    // resolveScopes / onSuccess callbacks.
-    const stubStore = {
-        async active() {
-            return null
-        },
-        async set() {},
-        async clear() {},
-        getLastStorageResult: () => undefined,
-        getLastClearResult: () => undefined,
-    }
-    attachTodoistLoginCommand(
-        program,
-        stubStore as unknown as Parameters<typeof attachTodoistLoginCommand>[1],
-    )
+    attachTodoistLoginCommand(program, createTodoistTokenStore())
     return capturedAttachOptions[capturedAttachOptions.length - 1].options as AttachOptions
 }
 
--- a/src/commands/auth/login.test.ts
+++ b/src/commands/auth/login.test.ts
@@ -53,7 +53,22 @@ function attachAndCapture(): AttachOptions {
     capturedAttachOptions.length = 0
     const program = new Command()
     program.exitOverride()
-    attachTodoistLoginCommand(program)
+    // Stub `TodoistTokenStore` — the login attacher just forwards the store to
+    // cli-core, which is mocked here. The tests only exercise the local
+    // resolveScopes / onSuccess callbacks.
+    const stubStore = {
+        async active() {
+            return null
+        },
+        async set() {},
+        async clear() {},
+        getLastStorageResult: () => undefined,
+        getLastClearResult: () => undefined,
+    }
+    attachTodoistLoginCommand(
+        program,
+        stubStore as unknown as Parameters<typeof attachTodoistLoginCommand>[1],
+    )
     return capturedAttachOptions[capturedAttachOptions.length - 1].options as AttachOptions
 }
 
--- a/src/commands/auth/logout.ts
+++ b/src/commands/auth/logout.ts
@@ -1,9 +1,30 @@
+import { attachLogoutCommand } from '@doist/cli-core/auth'
 import chalk from 'chalk'
-import { clearApiToken } from '../../lib/auth.js'
+import type { Command } from 'commander'
+import type { TodoistAccount, TodoistTokenStore } from '../../lib/auth-store.js'
 import { logTokenStorageResult } from './helpers.js'
 
-export async function logout(): Promise<void> {
-    const result = await clearApiToken()
-    console.log(chalk.green('✓'), 'Logged out')
-    logTokenStorageResult(result, 'Stored token removed from the system credential manager')
+/**
+ * Attach `td auth logout` via cli-core's generic `attachLogoutCommand`. The
+ * registrar emits the success line (`✓ Logged out` / `{ok:true}` / silent
+ * ndjson); `onCleared` only surfaces the keyring-fallback warning carried by
+ * `TokenStorageResult` — cli-core's `TokenStore.clear: void` contract can't
+ * expose it directly, so we stash it on the adapter (`getLastClearResult`).
+ */
+export function attachTodoistLogoutCommand(auth: Command, store: TodoistTokenStore): Command {
+    return attachLogoutCommand<TodoistAccount>(auth, {
+        store,
+        onCleared: ({ view }) => {
+            const result = store.getLastClearResult()
+            if (!result) return
+            if (view.json || view.ndjson) {
+                if (result.warning) console.error(chalk.yellow('Warning:'), result.warning)
+            } else {
+                logTokenStorageResult(
+                    result,
+                    'Stored token removed from the system credential manager',
+                )
+            }
+        },
+    })
 }
--- a/src/commands/auth/logout.ts
+++ b/src/commands/auth/logout.ts
@@ -1,5 +1,4 @@
 import { attachLogoutCommand } from '@doist/cli-core/auth'
-import chalk from 'chalk'
 import type { Command } from 'commander'
 import type { TodoistAccount, TodoistTokenStore } from '../../lib/auth-store.js'
 import { logTokenStorageResult } from './helpers.js'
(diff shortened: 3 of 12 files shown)


doistbot (commented):
This PR cleanly migrates the `auth logout` and `auth status` commands to the new `@doist/cli-core` registrars while unifying the underlying token store implementation. Leveraging the upstream framework brings welcome consistency to our authentication handling and machine-readable output formats. A few remaining details require attention, specifically resolving state-consistency edge cases with the `--user` flag, optimizing redundant credential lookups, deduplicating account mapping and keyring warning logic, and ensuring the new paths are fully covered by tests and documented in the agent skill content.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F334&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A07bc8b70-4e3b-11f1-9aed-908514313e24%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A07bc8b70-4e3b-11f1-9aed-908514313e24&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/index.ts
```diff
@@ -1,21 +1,34 @@
...
 
-    attachTodoistLoginCommand(auth)
+    // Shared store instance: login stashes the post-`set` storage result for
+    // its success handler, logout reads the post-`clear` result for the same
+    // keyring-fallback warning surface. Status uses `active()` as the
+    // authenticated-snapshot gate.
+    const store = createTodoistTokenStore()
+
+    attachTodoistLoginCommand(auth, store)
+    attachTodoistLogoutCommand(auth, store)
```
doistbot, line 19: [P3] The migration to `cli-core` implicitly adds `--ndjson` support to `td auth status` and `td auth logout` (along with `--json` for logout). According to `AGENTS.md`, any newly added flags must be reflected in `SKILL_CONTENT` (`src/lib/skills/content.ts`). Consider updating the Authentication reference there to mention the availability of these flags for these commands.

src/commands/auth/status.ts
```diff
@@ -56,23 +55,101 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+export function attachTodoistStatusCommand(auth: Command, store: TodoistTokenStore): Command {
+    let data: StatusData | null = null
+
+    return attachStatusCommand<TodoistAccount>(auth, {
+        store,
+        description: 'Show current authentication status',
+        fetchLive: async ({ account }) => {
+            data = await gatherStatusData()
+            return {
+                ...account,
```
doistbot: [P2] This reuses the account returned by `store.active()`, but `active()` still ignores `--user` and returns the default/sole stored account. For `td --user <other> auth status` when a different default user exists, `fetchLive()` can return a mixed snapshot (default user's `id`/`label`, selected user's email/auth metadata). Either make `active()` return `null` when `--user` is set, or rebuild the returned account from `data.user` instead of spreading `account`.

src/commands/auth/logout.ts
```diff
@@ -1,9 +1,30 @@
...
+ * `TokenStorageResult` — cli-core's `TokenStore.clear: void` contract can't
+ * expose it directly, so we stash it on the adapter (`getLastClearResult`).
+ */
+export function attachTodoistLogoutCommand(auth: Command, store: TodoistTokenStore): Command {
+    return attachLogoutCommand<TodoistAccount>(auth, {
+        store,
+        onCleared: ({ view }) => {
+            const result = store.getLastClearResult()
+            if (!result) return
+            if (view.json || view.ndjson) {
```
doistbot: [P2] The `view.json || view.ndjson` branch duplicates the keyring-fallback warning logic from `login.ts`. Instead of copying this pattern inline, consider updating `logTokenStorageResult` in `helpers.ts` to accept an `isMachineOutput` flag. This would allow both `login` and `logout` to rely purely on the helper while safely keeping the human-readable success message out of stdout.

src/commands/auth/status.ts
```diff
@@ -56,23 +55,101 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+ */
+export function attachTodoistStatusCommand(auth: Command, store: TodoistTokenStore): Command {
+    let data: StatusData | null = null
+
+    return attachStatusCommand<TodoistAccount>(auth, {
+        store,
+        description: 'Show current authentication status',
+        fetchLive: async ({ account }) => {
+            data = await gatherStatusData()
+            return {
```
doistbot: [P2] `auth-store.ts` already exposes `toTodoistAccount()` as the single source of truth for this shape, but `fetchLive` rebuilds the same mapping inline here. That duplicates the account projection and can drift — this version already refreshes `email` while leaving `label` from the stale snapshot. Reuse `toTodoistAccount(...)` here instead of hand-assembling the fields.

src/commands/auth/status.ts
```diff
@@ -56,23 +55,101 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+ * full status fetch through `onNotAuthenticated`; when `active()` does return
+ * a snapshot, `fetchLive` covers the same gather so renderText/renderJson can
+ * read from a single closure-captured `StatusData` regardless of which path
+ * we took. The same gather (`gatherStatusData`) is reused so output is
+ * byte-for-byte identical between paths.
+ */
+export function attachTodoistStatusCommand(auth: Command, store: TodoistTokenStore): Command {
+    let data: StatusData | null = null
+
+    return attachStatusCommand<TodoistAccount>(auth, {
```
doistbot, line 130: [P2] Wiring `status` through `attachStatusCommand` makes cli-core call `store.active()` before either branch here, and both branches then immediately run `gatherStatusData()` (`getApi()` + `getAuthMetadata()` + config reads). That means every `td auth status` now pays for an extra config/keyring lookup compared to the old direct handler, and the persisted-user path resolves credentials three times in one command. Consider keeping `status` hand-rolled, or reusing/caching the `active()` snapshot so this command only resolves auth once.

src/commands/auth/login.test.ts
```diff
@@ -53,7 +53,22 @@ function attachAndCapture(): AttachOptions {
...
+            return null
+        },
+        async set() {},
+        async clear() {},
+        getLastStorageResult: () => undefined,
+        getLastClearResult: () => undefined,
+    }
+    attachTodoistLoginCommand(
+        program,
+        stubStore as unknown as Parameters<typeof attachTodoistLoginCommand>[1],
```
doistbot: [P2] The `as unknown as Parameters<...>[1]` cast bypasses type safety for the mock store. Since `createTodoistTokenStore()` is a pure factory function, you can import and pass it directly here to avoid maintaining a manual stub and a brittle type cast.

src/commands/auth/status.ts
```diff
@@ -56,23 +55,101 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+        store,
+        description: 'Show current authentication status',
+        fetchLive: async ({ account }) => {
```
doistbot: [P2] This refactor adds a separate `fetchLive` path, but there is still no deterministic test that drives `store.active()` to a real account and exercises `fetchLive`/`renderText`/`renderJson`. The existing auth tests don't stub the store, so the account-to-cli-core field mapping here can regress unnoticed. Please add a focused test for the persisted-account path.

src/commands/auth/logout.ts
```diff
@@ -1,9 +1,30 @@
...
+ * Attach `td auth logout` via cli-core's generic `attachLogoutCommand`. The
+ * registrar emits the success line (`✓ Logged out` / `{ok:true}` / silent
+ * ndjson); `onCleared` only surfaces the keyring-fallback warning carried by
+ * `TokenStorageResult` — cli-core's `TokenStore.clear: void` contract can't
+ * expose it directly, so we stash it on the adapter (`getLastClearResult`).
+ */
+export function attachTodoistLogoutCommand(auth: Command, store: TodoistTokenStore): Command {
+    return attachLogoutCommand<TodoistAccount>(auth, {
+        store,
+        onCleared: ({ view }) => {
```
doistbot, line 16: [P2] The only logout coverage still checks the plain `✓ Logged out` path, but the Todoist-local behavior in this migration lives in `onCleared`: surfacing `TokenStorageResult.warning`, especially keeping JSON/NDJSON stdout clean by sending warnings to stderr. Please add a test for a warning result so this new branch is covered.

src/commands/auth/index.ts
scottlovegrove: Done in 03bb4d7 — `SKILL_CONTENT` now lists `td auth status --json` / `--ndjson` and `td auth logout --json` (plus a note that `--ndjson` is silent on logout), and the narrative paragraph below covers the three-attacher flag parity. `npm run check:skill-sync` is green.

src/commands/auth/status.ts
scottlovegrove: Fixed in 03bb4d7 — `fetchLive` no longer spreads the snapshot account. It rebuilds via `toTodoistAccount(...)` from `data.user`, and when `getRequestedUserRef()` is set it also bypasses the snapshot token shortcut and re-resolves through `getApi()` so the displayed account matches the `--user` selector instead of the snapshot default. New `falls back to getApi() when --user is set` test covers this branch.

src/commands/auth/logout.ts
scottlovegrove: Done in 03bb4d7 — `logTokenStorageResult` now takes an `isMachineOutput` flag and gates the human "stored/removed" confirmation behind it; warnings always go to stderr. Both `login.ts` and `logout.ts` collapse to a single call: `logTokenStorageResult(result, msg, view.json || view.ndjson)`.

src/commands/auth/status.ts
scottlovegrove: Done in 03bb4d7 — `fetchLive` now returns `toTodoistAccount({ id, email, authMode, authScope, authFlags })` built from the freshly-fetched `data.user` + `data.metadata`. No more hand-spread; the `label` field is populated consistently by the shared helper, and the `--user` mismatch flagged in the sibling comment is gone as a side benefit.

src/commands/auth/status.ts
scottlovegrove: Partially addressed in 03bb4d7. `gatherStatusData(token?)` now takes an optional pre-resolved token: when the snapshot path runs and no `--user <ref>` is set, `fetchLive` passes `snapshot.token` through and we skip the `resolveActiveUser` → keyring round-trip via `createApiForToken(token)`. The `--user` branch and `onNotAuthenticated` still go through `getApi()` because they need the selector / env precedence rules.

That removes the duplicate keyring read on the persisted-default-user path (the common case). The three config reads (`getAuthMetadata` / `listStoredUsers` / `readConfig`) inside `gatherStatusData` remain — those are cheap file I/O and unifying them would require restructuring `auth.ts`, which feels out of scope for this PR. Happy to follow up if that materially shows up in profiling.

src/commands/auth/login.test.ts
scottlovegrove: Done in 03bb4d7 — `attachAndCapture` now passes `createTodoistTokenStore()` directly. Type-cast and hand stub are gone.

src/commands/auth/status.ts
scottlovegrove: Added in 03bb4d7 — new `persisted-account snapshot path (fetchLive)` describe block in `auth.test.ts` drives a controllable stub `TodoistTokenStore` directly into `attachTodoistStatusCommand` (bypassing `registerAuthCommand` so we do not depend on the real keyring/config state of the test host). Three tests:

1. Text status from the snapshot — asserts `createApiForToken("snapshot_token")` is called and the `✓ Authenticated (default)` / Email / Name / Mode lines render.
2. `--json` envelope from the snapshot path — parses the JSON and matches against the full field set including `source` and `isDefault`.
3. `--user <ref>` re-resolve fallback — asserts that with `--user` set, `createApiForToken` is **not** called and `getApi()` is used instead so the selector is honoured.

src/commands/auth/logout.ts
scottlovegrove: Added in 03bb4d7 — three new tests in the `logout subcommand` block cover the warning surface:

1. Plain mode — `consoleSpy` asserted with `✓ Logged out` (stdout) and `errorSpy` asserted with the `Warning:` + warning message (stderr).
2. `--json` mode — stdout receives exactly the `{"ok": true}` envelope (no "Stored token removed" line), warning still goes to stderr.
3. `--ndjson` mode — stdout is fully silent, warning still goes to stderr.

All three drive `mockClearApiToken.mockResolvedValue({ storage: "config-file", warning: ... })` so the keyring-fallback branch is exercised end-to-end through the new `logTokenStorageResult(result, msg, view.json || view.ndjson)` shape.

PR #337, opened 2026-05-13
fix(comment): make `comment add --file` actually upload the file

engfragui:
### Problem 

`td comment add --file` didn't actually upload the file. Two issues:

1. The SDK builds multipart bodies with the `form-data` npm package, which undici's native fetch can't serialize — it coerces the value to `"[object FormData]"` (17 bytes), so the file never reaches the server.
Note: This same issue was present when creating or importing a template.

2. The SDK's upload-client validates the `/api/v1/uploads` response without first running `camelCaseKeys`, so `resource_type` never lands on `resourceType` and Zod throws.

### Solution

1. Read file as a `Blob` and then feed it to `uploadFile` (and template-related methods)
2. Fixed in https://github.com/Doist/todoist-sdk-typescript/pull/601 (so here we just need a `todoist-sdk-typescript` bump)

## Reference

- Fixes https://github.com/Doist/Issues/issues/20242
- Depends on https://github.com/Doist/todoist-sdk-typescript/pull/601

### Comments

- Create some `sample_file.txt`
- Pick a `task_id` and run the following:
```
node dist/index.js comment add id:{task_id} \
  --content "td comment-add smoke test" \
  --file sample_file.txt \
  --verbose --json
```
- [ ] Verify that the file gets successfully attached 
- Run the same command but with a non existing file
- [ ] Verify that you get a `FILE_NOT_FOUND` error

### Templates

- Export an existing project as csv (save it as `template.csv`)
- Run the following:
```
node dist/index.js template create \
  --name "td template-create smoke test" \
  --file template.csv \
  --verbose --json
```
- [ ] Verify that the project gets successfully created 
- Run the same command but with a non existing file
- [ ] Verify that you get a `FILE_NOT_FOUND` error
- Pick a `project_id` and run the following:
```
node dist/index.js template import-file id:{project_id} \
  --file template.csv \
  --verbose --json
```
- [ ] Verify that the content gets successfully imported into the existing project
- Run the same command but with a non existing file
- [ ] Verify that you get a `FILE_NOT_FOUND` error

### Comments

Successful `commend add`

Screenshot 2026-05-14 at 09 57 33

Failed `commend add` (file not found)

```
{"error":{"code":"FILE_NOT_FOUND","message":"File not found: /Users/francesca/conductor/workspaces/todoist-cli/memphis/sample_file.txt","hints":["Check the file path and try again."]}}
```

### Templates 

`template create`

Screenshot 2026-05-14 at 09 55 51

`template import-file`

Screenshot 2026-05-14 at 09 55 43


src/lib/api/core.ts
```diff
@@ -291,6 +292,11 @@ function analyzeAndEmitApiResponse(
 
 export function createApiForToken(token: string): TodoistApi {
     const rawApi = new TodoistApi(token, { customFetch: createTrackedFetch() })
+    // SDK ≤ 0.10.x validates the `/api/v1/uploads` response without first
+    // running `camelCaseKeys`, so `resource_type` never lands on
+    // `resourceType` and Zod throws. Override with a direct upload until
+    // the SDK is patched.
+    rawApi.uploadFile = (args) => uploadAttachment(args, token)
```
engfragui: Fixed in https://github.com/Doist/todoist-sdk-typescript/pull/601

src/lib/uploads.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const response = await fetchTodoist(UPLOADS_URL, {
+        method: 'POST',
+        body: form,
+        headers: { Authorization: `Bearer ${token}` },
+    })
+
+    if (!response.ok) {
+        const detail = await response.text().catch(() => '')
+        throw new Error(
```
doistbot: [P1] This helper now sits on a user-facing command path, but the new raw `Error` throws here (and in the validation branches below) bypass the CLI's `BaseCliError` handling. In `--json` mode, ordinary cases like 401/413 responses, missing files, or missing `fileName` will be surfaced as `INTERNAL_ERROR`, and the auth/rate-limit hints from `wrapApiError()` are lost. Please convert these failures into `CliError`s or otherwise route them through the structured API-error path.

src/lib/uploads.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    const raw = (await response.json()) as Record<string, unknown>
+    return camelCaseAttachment(raw, fileName)
+}
+
+async function resolveUploadBody(args: UploadFileArgs): Promise<{ body: Blob; fileName: string }> {
+    const { file, fileName } = args
+    if (typeof file === 'string') {
+        const buffer = await readFile(file)
```
doistbot: [P2] `readFile()` makes every path-based upload fully resident in memory before the request is sent. That is a behavior regression from the SDK/form-data path, which could stream file contents, and it makes larger attachments prone to large memory spikes or OOMs. Please switch this branch to a file-backed `Blob`/streaming approach instead of eagerly buffering the whole file.

src/lib/uploads.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (!response.ok) {
+        const detail = await response.text().catch(() => '')
+        throw new Error(
+            `Upload failed: HTTP ${response.status} ${response.statusText}${detail ? ` — ${detail}` : ''}`,
+        )
+    }
+
+    const raw = (await response.json()) as Record<string, unknown>
+    return camelCaseAttachment(raw, fileName)
```
doistbot: [P2] This returns a hand-built `Attachment` after an unchecked cast from `response.json()`, with silent fallbacks like `resourceType: 'file'` and `fileName: fallbackName`. That weakens the `Promise<Attachment>` contract and creates a second schema copy that can drift from the SDK. Since this workaround exists specifically because the SDK validates too early, it would be safer to camel-case first and then validate against the attachment schema (or a local equivalent) instead of returning a best-effort object.

src/lib/usage-tracking.test.ts
```diff
@@ -74,6 +74,41 @@ describe('usage tracking', () => {
         expect(response.ok).toBe(true)
     })
 
+    it('materializes form-data package bodies into a Buffer for undici fetch', async () => {
+        const { default: FormData } = await import('form-data')
```
doistbot: [P2] This dynamic import likely relies on `form-data` being available as a transitive dependency via `@doist/todoist-sdk`. If the SDK replaces or drops this package in a future update, this test will unexpectedly break with a `MODULE_NOT_FOUND` error. Since `isFormDataPackageInstance` detects the package entirely via duck-typing (`getBuffer`, `getHeaders`, `pipe`), consider replacing this import with a plain mock object that matches the expected `FormDataPackage` stream shape to isolate the test from upstream dependency tree changes.

src/lib/usage-tracking.test.ts
```diff
@@ -74,6 +74,41 @@ describe('usage tracking', () => {
...
+            captured = options
+            return new Response('{}', {
+                status: 200,
+                headers: { 'content-type': 'application/json' },
+            })
+        })
+
+        const form = new FormData()
+        form.append('file_name', 'test.bin')
+        form.append('file', Buffer.from('hello world bytes'), {
```
doistbot: [P2] This test only appends a Buffer part, so it would still pass if `formDataToBuffer()` regressed to `getBuffer()`-style handling. The new implementation explicitly exists to support stream-backed multipart parts, which is the higher-value failure mode for real uploads. Please make the `file` part a `Readable`/`fs.createReadStream()` and assert that path still serializes correctly.

src/lib/api/core.ts
doistbot: [P2] Please add a regression test for this wiring in `core.test.ts`. The new tests cover `uploadAttachment()` directly, but the command tests mock the API client, so they will not fail if `createApiForToken()` stops overriding `uploadFile` and the SDK's broken implementation is used again.

doistbot (commented):
This PR thoughtfully restores file upload functionality for comments by resolving serialization incompatibilities with native fetch and working around an SDK validation bug. The adjustments ensure the command successfully attaches files and interacts correctly with the API. A few refinements are noted, specifically around surfacing structured `CliError`s, switching to a streaming approach to avoid memory spikes, strictly validating the attachment schema, and hardening the test suite with stream-backed mocks and coverage for the SDK override.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A316fa470-4eea-11f1-94ca-3a9b991d3c06%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A316fa470-4eea-11f1-94ca-3a9b991d3c06&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

rfgamaral:
Please mention me again when doistbot comments are addressed, and you mark this ready for review :)

engfragui:
@rfgamaral Apologies, I didn't realize a reviewer was going to be added automatically 🙈 Will ping you again tomorrow once this is clean and ready for review.

engfragui:
@doistbot /review

src/lib/usage-tracking.ts
```diff
@@ -104,18 +104,64 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+// (`getBuffer` alone would suffice) and stream parts like
+// `fs.createReadStream`, which `getBuffer` does not serialize.
+function formDataToBuffer(form: FormDataPackage): Promise<Buffer> {
+    return new Promise((resolve, reject) => {
+        const chunks: Buffer[] = []
+        form.on('data', (chunk) => {
+            chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk as string))
+        })
+        form.on('end', () => resolve(Buffer.concat(chunks)))
+        form.on('error', (err) => reject(err as Error))
```
doistbot: [P1] This new `form.on('error')` path rejects with a plain `Error`. For `comment add --file`, local file failures like ENOENT/EACCES will now bubble through the API proxy as generic errors and end up rendered as `INTERNAL_ERROR` in `--json` mode instead of a structured CLI error. Please convert multipart-read failures to an appropriate `CliError` (for example `FILE_NOT_FOUND` / `FILE_READ_ERROR`) before they escape this layer.

src/lib/usage-tracking.ts
```diff
@@ -104,18 +104,64 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
             abortSignal = signal ? AbortSignal.any([signal, timeoutSignal]) : timeoutSignal
         }
 
+        // Undici's fetch (Node's native `fetch`) doesn't recognize the
+        // `form-data` npm package as a body type, so it coerces it to
+        // `"[object FormData]"` — a 17-byte string — and the upload arrives
+        // without the file. Materialize it into a Buffer so undici sends
+        // the real multipart payload. The SDK already populated the
+        // multipart `Content-Type` (with boundary) in `headers`.
+        const resolvedBody = isFormDataPackageInstance(body) ? await formDataToBuffer(body) : body
```
doistbot: [P2] `formDataToBuffer()` runs entirely before `fetch` receives `signal`, so multipart uploads no longer honor cancellation/timeout while the file is being materialized. An already-aborted signal, or a timeout that fires during buffering, will still read the whole form before failing. Please short-circuit on `signal.aborted` and race the bufferization step against the abort signal so uploads keep the same timeout/cancel semantics as other requests.

src/lib/usage-tracking.ts
```diff
@@ -104,18 +104,64 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+    getHeaders: () => Record<string, string>
+    pipe: (...args: unknown[]) => unknown
+    on: (event: 'data' | 'end' | 'error', listener: (arg?: unknown) => void) => unknown
+    resume: () => unknown
+}
+
+function isFormDataPackageInstance(body: unknown): body is FormDataPackage {
+    if (typeof body !== 'object' || body === null) return false
+    const candidate = body as Record<string, unknown>
+    return (
```
doistbot: [P2] The type guard asserts that `body` is a `FormDataPackage`, which includes `.on` and `.resume` methods. Since `formDataToBuffer` strictly invokes these methods, please also add `typeof candidate.on === 'function'` and `typeof candidate.resume === 'function'` to the checks here to ensure complete type safety and prevent runtime errors if an incomplete object or partial mock is intercepted.

src/lib/usage-tracking.ts
```diff
@@ -104,18 +104,64 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+        typeof candidate.getHeaders === 'function' &&
+        typeof candidate.pipe === 'function'
+    )
+}
+
+// `form-data` is a CombinedStream — it emits buffers via 'data' events but
+// is not async-iterable, so we read it manually. Handles both buffer parts
+// (`getBuffer` alone would suffice) and stream parts like
+// `fs.createReadStream`, which `getBuffer` does not serialize.
+function formDataToBuffer(form: FormDataPackage): Promise<Buffer> {
```
doistbot: [P2] Buffering the entire multipart form into memory can cause memory spikes for large attachments, similar to the regression noted in the previous review round. Since `form-data` implements `.pipe()`, consider avoiding fully resident memory by piping the form into a `PassThrough` stream (which is async-iterable) and passing that stream directly to `fetch` (adding `duplex: 'half'` to `fetchOptions` to support the stream body).

src/lib/usage-tracking.ts
doistbot: [P2] This workaround is documented as an undici/native-`fetch` compatibility fix, but it now runs for every `baseFetch` passed to `createTrackedFetch()` and does so by fully buffering the multipart body. That broadens this helper's contract for custom transports and mixes upload serialization into a cross-cutting tracking layer. Please gate the conversion to the native-fetch path (similar to `attachDispatcherIfNative`) and avoid making the generic fetch wrapper own eager multipart materialization.

src/lib/usage-tracking.test.ts
```diff
@@ -74,6 +76,77 @@ describe('usage tracking', () => {
         expect(response.ok).toBe(true)
     })
 
+    it('materializes form-data package bodies into a Buffer for undici fetch', async () => {
```
doistbot: [P2] This only verifies `createTrackedFetch()` in isolation. The `comment add --file` tests still mock `getApi()`/`uploadFile()`, so they would all keep passing if `createApiForToken()` stopped supplying `customFetch: createTrackedFetch()` and uploads regressed again. Please add a small regression test in `src/lib/api/core.test.ts` for that wiring; otherwise the command path this PR fixes is still effectively untested.

doistbot (commented):
This PR addresses the file upload issue in `comment add --file` by updating the SDK and properly handling `form-data` serialization to ensure payloads correctly reach the server. These fixes effectively bridge the compatibility gap with native fetch and restore full functionality for file attachments. A few refinements are noted for further improving this implementation, specifically around optimizing memory management by streaming instead of fully buffering, maintaining expected timeout and cancellation semantics, narrowing the scope of the serialization fix strictly to the native fetch path, and tightening up type guards, error handling, and test coverage.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A186767a0-4ef5-11f1-9d2a-f8a5e1c89eca%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A186767a0-4ef5-11f1-9d2a-f8a5e1c89eca&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
@doistbot /review

src/lib/usage-tracking.ts
```diff
@@ -104,25 +106,102 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
-        const fetchOptions: RequestInit = {
+        // Undici's fetch doesn't recognize the `form-data` npm package as
+        // a body type — it coerces it to `"[object FormData]"` (17 bytes)
+        // and the upload arrives empty. Bridge it to a WHATWG stream so
+        // undici streams the real multipart payload. Gated to the native
+        // fetch path: test stubs and other custom transports get the body
+        // as-is and decide for themselves how to handle it.
+        const useNativeFetch = baseFetch === globalThis.fetch
+        const needsBodyBridge = useNativeFetch && isFormDataPackageInstance(body)
+        const resolvedBody = needsBodyBridge ? formDataToWebStream(body) : body
```
doistbot: [P2] This still starts the multipart stream before an already-aborted `signal` is honored: `formDataToWebStream(body)` calls `form.pipe()` immediately, which puts the form into flowing mode before `fetch` sees `abortSignal`. A canceled upload can therefore still open/read the local file and may even race into `FILE_NOT_FOUND` / `FILE_READ_ERROR` instead of aborting cleanly. Please short-circuit when `abortSignal?.aborted` before bridging, or defer starting `pipe()` until the request is actually dispatched.

doistbot (commented):
This PR successfully fixes file uploads in `comment add` by adapting `form-data` payloads for native fetch and bumping the SDK to resolve response parsing issues. It is a great step toward fully functional file attachments, though the stream bridging logic and testing setup require some refinement to ensure robustness. A few adjustments are needed to properly handle abort signals and synchronous pipe errors, extract a shared file-error helper to reduce duplication, and improve the resilience and cleanup of the mock fetch tests.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aa2172330-4ef7-11f1-8318-fe249860283a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aa2172330-4ef7-11f1-8318-fe249860283a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/usage-tracking.ts
```diff
@@ -104,25 +106,102 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+    const pass = new PassThrough()
+    form.pipe(pass)
+    form.on('error', (err) => pass.destroy(err))
+    return Readable.toWeb(pass) as ReadableStream<Uint8Array>
+}
+
+// Walk the error/cause chain looking for a Node fs error so local file
+// problems (the SDK opens uploads via `fs.createReadStream`) surface as a
+// structured `CliError` rather than a generic `INTERNAL_ERROR`.
+function mapStreamErrorToCliError(error: unknown): CliError | undefined {
```
doistbot: [P2] `mapStreamErrorToCliError()` adds a third copy of the CLI's `ENOENT`/file-read → `CliError` translation. The same mapping is already hand-rolled in `src/commands/template/create.ts` and `src/commands/template/import-file.ts`, and the user-facing messages are already drifting. Please extract one shared file-error helper in `src/lib/` and reuse it here instead of introducing another variant.

src/lib/usage-tracking.ts
```diff
@@ -104,25 +106,102 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+// Bridge the `form-data` CombinedStream — which undici's native fetch
+// doesn't recognize — into a WHATWG ReadableStream it does. Piping
+// through a PassThrough keeps the body lazy (no in-process buffering of
+// the whole file) and lets undici honor `signal` / cancel the upstream
+// stream when the request is aborted. Stream errors (e.g. ENOENT from
+// `fs.createReadStream`) propagate to the PassThrough so undici rejects
+// the fetch with the original error as its `cause`.
+function formDataToWebStream(form: FormDataPackage): ReadableStream<Uint8Array> {
+    const pass = new PassThrough()
+    form.pipe(pass)
```
doistbot: [P2] `form.pipe(pass)` runs before the `'error'` listener is attached, so any synchronous error emitted during `pipe()` (or by another duck-typed body implementation accepted by `isFormDataPackageInstance`) will bypass the bridge and escape as an uncaught transport error. Register the error handler before starting the pipe, and ideally wrap `pipe()` itself so this adapter is robust to both sync and async failures.

src/lib/api/core.test.ts
```diff
@@ -50,3 +50,49 @@ describe('buildRescheduleDate', () => {
...
+
+        const api = createApiForToken('test-token')
+        await api.uploadFile({ file: Buffer.from('data'), fileName: 'test.bin' })
+
+        if (!captured) throw new Error('global fetch was not called — customFetch not wired')
+        // ReadableStream body + duplex: 'half' is the signature of
+        // `createTrackedFetch`'s form-data → stream bridge. If a future
+        // refactor drops `customFetch`, undici sees the raw form-data
+        // instance and `captured.body` is no longer a ReadableStream.
+        expect(captured.body).toBeInstanceOf(ReadableStream)
```
doistbot: [P2] Asserting that the body becomes a `ReadableStream` makes this test brittle to upstream SDK changes. If the SDK ever switches to using native WHATWG `FormData` instead of the `form-data` package, `createTrackedFetch` will correctly bypass the stream bridge, but this test will fail. Verifying the `doist-platform: cli` header below is sufficient to prove that `customFetch` is wired correctly without coupling to the SDK's internal payload serialization.

src/lib/api/core.test.ts
```diff
@@ -50,3 +50,49 @@ describe('buildRescheduleDate', () => {
...
+        expect((captured as RequestInit & { duplex?: string }).duplex).toBe('half')
+        // Tracking headers also come from `createTrackedFetch`.
+        const headers = captured.headers as Record<string, string>
+        expect(headers['doist-platform']).toBe('cli')
+
+        vi.restoreAllMocks()
```
doistbot: [P2] If any `expect` assertion fails, this manual restore is skipped, which can leak the `globalThis.fetch` mock to other test suites. Please remove this if Vitest is configured to automatically restore mocks, or move it to an `afterEach` hook or `finally` block.

src/lib/usage-tracking.test.ts
```diff
@@ -74,6 +75,175 @@ describe('usage tracking', () => {
...
+        }
+
+        const fileBytes = Buffer.from('hello world bytes from a stream')
+        const form = new FormDataStub(fileBytes)
+
+        let captured: RequestInit | undefined
+        // The bridge only kicks in on the native-fetch path. Mock
+        // `globalThis.fetch` so the gate (`baseFetch === globalThis.fetch`)
+        // matches.
+        vi.spyOn(globalThis, 'fetch').mockImplementation((async (
```
doistbot: [P2] These new multipart tests spy on `globalThis.fetch`, but this top-level `describe` has no matching cleanup; the only `afterEach(vi.restoreAllMocks)` in the file lives inside the later nested `proxy dispatcher injection` suite. If one of these cases fails, the mocked fetch will bleed into following tests. Please add a describe-level `afterEach(() => vi.restoreAllMocks())` for this block.

src/lib/usage-tracking.test.ts
```diff
@@ -74,6 +75,175 @@ describe('usage tracking', () => {
...
+            captured = options
+            return new Response('{}', {
+                status: 200,
+                headers: { 'content-type': 'application/json' },
+            })
+        }) as typeof fetch)
+
+        const trackedFetch = createTrackedFetch()
+
+        await trackedFetch('https://api.todoist.com/api/v1/uploads', {
```
doistbot: [P2] The new multipart coverage still never exercises the bridged path with an already-aborted or mid-stream-aborted `signal`. The existing timeout tests only cover plain `GET` requests, so a regression where multipart uploads stop honoring cancellation would still pass. Please add one abort-focused case on this stream-backed upload path.

src/lib/usage-tracking.ts
```diff
@@ -104,25 +106,118 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
     }
 }
 
+// Minimal shape of the `form-data` npm package's FormData class. The SDK
+// uses it internally for multipart uploads. Distinct from the global
+// WHATWG `FormData`, which has no `getHeaders`/`pipe`. We require `pipe`
+// + `on` because the streaming workaround below uses both.
+type FormDataPackage = {
```
gnapse: nit: make it a jsdoc

```suggestion
/**
 * Minimal shape of the `form-data` npm package's FormData class. The SDK
 * uses it internally for multipart uploads. Distinct from the global
 * WHATWG `FormData`, which has no `getHeaders`/`pipe`. We require `pipe`
 * + `on` because the streaming workaround below uses both.
 */
type FormDataPackage = {
```

src/lib/usage-tracking.ts
```diff
@@ -104,25 +106,118 @@ function toCustomFetchResponse(response: Response): CustomFetchResponse {
...
+
+// Bridge the `form-data` CombinedStream — which undici's native fetch
+// doesn't recognize — into a WHATWG ReadableStream it does. Piping
+// through a PassThrough keeps the body lazy (no in-process buffering of
+// the whole file) and lets undici honor `signal` by destroying the
+// PassThrough on abort, which propagates upstream and closes any open
+// `fs.createReadStream` handles. Stream errors (e.g. ENOENT) propagate
+// the same way so undici rejects the fetch with the original error in
+// its `cause`.
+function formDataToWebStream(
```
gnapse: Same, make it jsdoc.

engfragui:
@doistbot /review

src/commands/comment/add.ts
```diff
@@ -67,13 +68,28 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
         | undefined
 
     if (options.file) {
-        const uploadResult = await api.uploadFile({ file: options.file })
+        let buffer: Buffer
+        try {
+            buffer = await readFile(options.file)
```
doistbot: [P2] Reading the entire file into memory via `readFile()` causes memory spikes for large attachments, which is a regression from the older streaming behavior. Since the project uses Node >= 20.18.1, please replace `readFile()` with `openAsBlob()` (imported from `node:fs`). It returns a file-backed `Blob` that streams directly from disk without eagerly loading the payload into memory.

You can then remove the `buffer` variable and `new Blob(...)` wrapping below, and pass the returned blob directly to the API.

src/lib/api/core.test.ts
```diff
@@ -50,3 +50,35 @@ describe('buildRescheduleDate', () => {
         expect(result).toBe('2026-03-20T10:00:00')
     })
 })
+
+describe('createApiForToken', () => {
+    it('serializes Blob uploads as a real multipart body (not "[object FormData]")', async () => {
+        let captured: RequestInit | undefined
+        vi.spyOn(globalThis, 'fetch').mockImplementation((async (
```
doistbot: [P2] The `vi.spyOn(globalThis, 'fetch')` mock is never restored. Since `restoreMocks` is not enabled in `vitest.config.ts`, this mocked fetch will leak to any future tests added to this file. Please add an `afterEach(() => vi.restoreAllMocks())` hook or use `mockRestore()` at the end of the test to ensure cleanup.

src/lib/api/core.test.ts
```diff
@@ -50,3 +50,35 @@ describe('buildRescheduleDate', () => {
...
+                { status: 200, headers: { 'content-type': 'application/json' } },
+            )
+        }) as typeof fetch)
+
+        const api = createApiForToken('test-token')
+        const blob = new Blob([new Uint8Array(Buffer.from('data'))])
+        await api.uploadFile({ file: blob, fileName: 'test.bin' })
+
+        if (!captured) throw new Error('global fetch was not called')
+        expect(captured.body).toBeInstanceOf(FormData)
```
doistbot: [P2] Asserting `captured.body` is a native `FormData` couples this test to the SDK's current multipart implementation, not to behavior owned by `createApiForToken()`. If the SDK later changes its internal body type while uploads still work, this test will fail for the wrong reason. Prefer asserting the CLI contract here (for example, that `customFetch` is wired by checking the tracking header and that the upload response is still parsed successfully) without pinning the exact body class.

src/commands/comment/comment.test.ts
```diff
@@ -350,6 +370,29 @@ describe('comment add with attachment', () => {
...
+                '--content',
+                'See attached',
+                '--file',
+                join(attachmentTmpDir, 'does-not-exist.pdf'),
+            ]),
+        ).rejects.toMatchObject({
+            code: 'FILE_NOT_FOUND',
+        })
+        expect(mockApi.uploadFile).not.toHaveBeenCalled()
+    })
```
doistbot: [P2] The new error handling in `addComment` includes a fallback branch for `FILE_READ_ERROR` (e.g., permission denied, attempting to read a directory), but this scenario remains untested. Since real filesystem IO is already configured for this suite, you can easily cover this branch by adding a test that passes `attachmentTmpDir` (the directory itself) as the `--file` flag, which will reliably trigger an `EISDIR` read error.

doistbot (commented):
This PR successfully resolves the file upload bug for `td comment add` by transitioning to a `Blob`-based payload and updating the Todoist SDK. The changes restore a crucial feature and make the multipart integration significantly more reliable. A few adjustments are needed around optimizing memory usage by using `openAsBlob()` rather than eagerly reading files into memory, ensuring global fetch mocks are properly cleaned up to prevent test leakage, decoupling a test assertion from the SDK's internal body type, and covering the new read error fallback branch with an additional test.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A67d9ea60-4f18-11f1-962a-5693f0731999%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A67d9ea60-4f18-11f1-962a-5693f0731999&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/template/create.ts
```diff
@@ -49,10 +45,16 @@ export async function createFromTemplate(options: CreateFromTemplateOptions): Pr
         workspaceId = workspace.id
     }
 
-    let file: Buffer
+    let file: Blob
     try {
-        file = fs.readFileSync(filePath) as Buffer
+        await stat(filePath)
```
doistbot: [P2] Moving the `stat`/`openAsBlob` check down here changes error precedence: `template create --file missing.csv` used to fail locally before any auth or workspace lookup, but now it calls `getApi()` (and may resolve the workspace) first. That means a missing file can surface `NO_TOKEN`/workspace errors or trigger unnecessary API work instead of `FILE_NOT_FOUND`. Please restore the local file validation before API/ref resolution here; `import-file.ts` has the same regression.

src/commands/comment/add.ts
```diff
@@ -67,13 +69,33 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
...
     if (options.file) {
-        const uploadResult = await api.uploadFile({ file: options.file })
+        // `openAsBlob` returns a file-backed Blob that streams from disk
+        // on demand — no eager full-file read. Passing the Blob to the
+        // SDK makes it take its native-FormData branch (which undici can
+        // serialize). `stat` first because `openAsBlob` rewraps fs
+        // errors as opaque `ERR_INVALID_ARG_VALUE` TypeErrors, which
+        // would lose our `FILE_NOT_FOUND` vs `FILE_READ_ERROR` mapping.
+        let file: Blob
+        try {
```
doistbot: [P2] This `try/catch` block that handles `stat()` + `openAsBlob()` and maps `ENOENT` / read errors into `CliError` instances is now duplicated identically across `src/commands/comment/add.ts`, `src/commands/template/create.ts`, and `src/commands/template/import-file.ts`. Please extract a shared utility (e.g., `openFileAsBlob(path, typeName)`) in `src/lib/` to handle this initialization and standard error mapping centrally.

src/commands/comment/add.ts
```diff
@@ -67,13 +69,33 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
...
+            await stat(options.file)
```
doistbot: [P2] `openAsBlob()` internally performs its own `stat` operation to populate the resulting `Blob`'s `.size` property. Calling `stat()` immediately beforehand creates a redundant filesystem hit on every successful upload. To avoid this duplicate work on the happy path, consider calling `openAsBlob()` directly and moving the `stat()` call into the `catch` block to disambiguate the opaque `ERR_INVALID_ARG_VALUE` error only when an error actually occurs. (This same optimization applies to the identical patterns in `src/commands/template/create.ts` and `src/commands/template/import-file.ts`).

src/commands/comment/comment.test.ts
doistbot: [P2] The `FILE_READ_ERROR` fallback branch in `addComment` (which handles generic read failures like permissions or EISDIR) remains untested. Since this suite already manages real filesystem artifacts, please add a case that passes the directory `attachmentTmpDir` itself as the `--file` argument to reliably trigger an `EISDIR` read error and exercise this path.

src/commands/comment/comment.test.ts
```diff
@@ -350,6 +370,29 @@ describe('comment add with attachment', () => {
         consoleSpy.mockRestore()
     })
 
+    it('returns FILE_NOT_FOUND when the --file path does not exist', async () => {
+        const program = createProgram()
+
+        mockApi.getTask.mockResolvedValue({ id: 'task-1', content: 'Buy milk' })
```
doistbot, line 407: [P3] This new test hand-builds a Todoist task object. The repo’s test guidance is to use shared fixtures from `src/test-support/fixtures.ts` instead of inline mock entities so tests stay aligned with canonical shapes. Please switch this to `fixtures.tasks.basic` (or add a fixture if you need a different shape).

doistbot (commented):
This PR resolves a crucial issue with multipart file uploads for comments and templates by adopting `Blob` and bumping the SDK to handle the updated responses. These changes provide a solid fix that ensures file attachments reach the server reliably while adding helpful test coverage for the upload flows. A few refinements would help polish the implementation, specifically by restoring the original error precedence to prevent unnecessary API calls, extracting the duplicated file-reading logic into an optimized shared utility, and tweaking the tests to cover directory read errors and use standard fixtures.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A975c83d0-4f1b-11f1-9a08-06566843384f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A975c83d0-4f1b-11f1-9a08-06566843384f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
@rfgamaral This is ready for review 🙏 I did my best!
Edit: Spoke too soon, had to address one last batch of doistbot comments. All done now ☑️

@gnapse Feel free to chime in as well if/as needed.

src/lib/local-file.ts
```diff
@@ -0,0 +1,33 @@
...
+ * `stat` runs before `openAsBlob` because `openAsBlob` rewraps fs
+ * errors as opaque `ERR_INVALID_ARG_VALUE` TypeErrors, which would
+ * collapse the `FILE_NOT_FOUND` / `FILE_READ_ERROR` distinction.
+ */
+export async function openLocalFileAsBlob(
+    rawPath: string,
+): Promise<{ blob: Blob; filePath: string; fileName: string }> {
+    const filePath = resolve(rawPath)
+    try {
+        await stat(filePath)
```
doistbot: [P1] `stat()` only proves that the path exists; it does not verify that the file can actually be read. Because `openAsBlob()` returns a file-backed `Blob` and defers the real read until fetch serializes the multipart body, a path like a `chmod 000` file can get past this helper and then fail later as a plain transport error, which the CLI will render as `INTERNAL_ERROR` instead of `FILE_READ_ERROR`. Please add a real readability preflight here (for example, open the file for reading and close it) before returning the `Blob`, so permission-denied uploads still stay on the structured `CliError` path.

src/commands/template/create.ts
```diff
@@ -24,12 +23,12 @@ export async function createFromTemplate(options: CreateFromTemplateOptions): Pr
         throw new CliError('MISSING_FILE', 'Template file path is required (--file)')
     }
 
-    const filePath = path.resolve(options.file)
-    if (!fs.existsSync(filePath)) {
-        throw new CliError('FILE_NOT_FOUND', `Template file not found: ${filePath}`, [
-            'Check the file path and try again.',
-        ])
-    }
+    const {
```
doistbot, line 26: [P2] This file-preparation block is now duplicated here and in `src/commands/template/import-file.ts` (`openLocalFileAsBlob(...)`, `filePath`, `defaultFileName`, then `options.fileName || defaultFileName`). Since both commands already share `template/helpers.ts`, please extract a small shared helper for template uploads so the two paths don't drift the next time file handling changes.

src/commands/comment/comment.test.ts
```diff
@@ -350,6 +370,57 @@ describe('comment add with attachment', () => {
...
```
doistbot, line 407: [P3] These new tests hand-build the same task stub inline, but the repo's test guidance is to reuse `src/test-support/fixtures.ts` entities. Please switch this to `fixtures.tasks.basic` (or a small helper derived from it) instead of duplicating `{ id: 'task-1', content: 'Buy milk' }` here and again in the following test.

src/lib/local-file.ts
doistbot: [P2] `openAsBlob()` performs its own `stat` operation internally to determine the file size. Calling `await stat(filePath)` immediately before it adds a redundant filesystem hit on every successful upload. To avoid this duplicate work on the happy path, call `openAsBlob()` directly and move the `stat()` call inside the `catch` block to disambiguate the opaque `ERR_INVALID_ARG_VALUE` error only when a failure actually occurs.

src/commands/comment/comment.test.ts
```diff
@@ -350,6 +370,57 @@ describe('comment add with attachment', () => {
...
+                '--file',
+                join(attachmentTmpDir, 'does-not-exist.pdf'),
+            ]),
+        ).rejects.toMatchObject({
+            code: 'FILE_NOT_FOUND',
+        })
+        expect(mockApi.uploadFile).not.toHaveBeenCalled()
+    })
+
+    it('returns FILE_READ_ERROR when the --file path is unreadable', async () => {
```
doistbot: [P2] Since the granular file error mapping was extracted to the shared `openLocalFileAsBlob` helper (and is fully tested in `local-file.test.ts`), re-testing the `FILE_READ_ERROR` fallback here using the embedded null-byte trick duplicates that coverage. Please remove this test to simplify the command suite, as we can now trust the shared helper to throw the correct error.

src/commands/template/template.test.ts
```diff
@@ -294,7 +300,9 @@ describe('template', () => {
 
         it('errors when file does not exist', async () => {
             const program = createProgram()
-            vi.mocked(fs.existsSync).mockReturnValue(false)
+            vi.mocked(stat).mockRejectedValue(
```
doistbot: [P2] These new file-error tests should also assert that `getApi()` and the ref resolver mocks were never called (same for the `import-file` cases below). The command was reordered to validate the file before any API/ref lookup, and without that assertion this suite would still pass if the old error-ordering regression came back.

doistbot (commented):
This PR successfully resolves file upload failures for comments and templates by switching to `Blob`-based reading and updating the SDK dependency. These changes nicely ensure that multipart payloads are properly serialized by native fetch, restoring a critical piece of CLI functionality. There are a few remaining adjustments to make around the new file helper, including verifying true file readability before returning, optimizing `stat` checks to avoid redundant filesystem calls on the happy path, extracting duplicated template logic, refining test coverage to fully protect the new operation order and fixtures, and adding the shared helper to `CODEBASE.md`.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F337&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A57b6f8a0-4f6b-11f1-9656-5a76d556794f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A57b6f8a0-4f6b-11f1-9656-5a76d556794f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/local-file.test.ts
```diff
@@ -0,0 +1,72 @@
...
+        const missing = join(tmpDir, 'does-not-exist.bin')
+
+        await expect(openLocalFileAsBlob(missing)).rejects.toMatchObject({
+            code: 'FILE_NOT_FOUND',
+            message: `File not found: ${missing}`,
+            hints: ['Check the file path and try again.'],
+        })
+    })
+
+    it('throws FILE_READ_ERROR for non-ENOENT failures, preserving the underlying message as a hint', async () => {
```
doistbot, line 77: [P2] This only exercises `FILE_READ_ERROR` when `stat()` rejects on an invalid path. The other important branch in `openLocalFileAsBlob()` is `openAsBlob()` failing after a successful `stat()`, and none of the new tests cover that. Please add a case that keeps `stat()` successful and forces `openAsBlob()` to reject so the helper’s full error mapping is protected.

src/lib/local-file.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Open a local file as a streaming `Blob` for upload, with CLI-grade
+ * error reporting. The returned Blob is file-backed — undici reads it
+ * lazily when serializing the multipart request body, so the payload
+ * never has to fit in memory all at once.
+ *
+ * `stat` runs before `openAsBlob` because `openAsBlob` rewraps fs
+ * errors as opaque `ERR_INVALID_ARG_VALUE` TypeErrors, which would
+ * collapse the `FILE_NOT_FOUND` / `FILE_READ_ERROR` distinction.
+ */
+export async function openLocalFileAsBlob(
```
doistbot, line 33: [P2] `openLocalFileAsBlob()` is a new shared helper under `src/lib/` and is already reused by multiple commands. Per the repo guidance, broadly reusable helpers must be added to `CODEBASE.md`’s `src/lib/` catalog so the repo map stays accurate.

rfgamaral (approved):

src/commands/template/create.ts
```diff
@@ -24,12 +23,12 @@ export async function createFromTemplate(options: CreateFromTemplateOptions): Pr
...
+        blob: file,
+        filePath,
+        fileName: defaultFileName,
+    } = await openLocalFileAsBlob(options.file)
+    const fileName = options.fileName || defaultFileName
```
scottlovegrove: I wonder whether, instead of passing `options.file` in, we pass just `options`, and then `openLocalFileAsBlob` can return the filename as well, since we always want it, and we can check against the options passed in then.

scottlovegrove (approved):

src/commands/template/create.ts
engfragui: 👍 Makes sense to me to delegate the decision to `openLocalFileAsBlob`. Done.
Minor: I ended up passing `options.file` and `options.fileName` to `openLocalFileAsBlob` (instead of the entire `options`) just because we didn't need the entire `options` object and also `options` differs slightly in shape between command/add, template/create, template/import-file, etc.

💭 I also added `options.fileName` to `command/add` which didn't have it as an option, but if template/create and template/import-file have it, then I think `command/add` should have it too! (There's no reason)

PR #340, opened 2026-05-15
feat(auth): adopt cli-core 0.12.0 multi-user TokenStore contract

scottlovegrove:
## Summary
- Bump `@doist/cli-core` from `0.10.0` → `0.12.0`, which reshapes `TokenStore<T>` as multi-user (`active(ref?)`, `clear(ref?)`, required `list()` + `setDefault(ref)`) and adds an exported `stripUserFlag` helper.
- Implement the new contract on `TodoistTokenStore` by wiring through todoist-cli's existing per-user primitives (`findUserByRef`, `getStoredUsers`, `getDefaultUser`, `setDefaultUserId`, `clearApiToken({ ref })`).
- Drop the local `stripUserFlag` implementation and its tests; import cli-core's instead. cli-core 0.12.0's `stripUserFlag` stops at the first positional, which matches the production pre-subcommand `--user` pattern in `src/index.ts` and lets commander's root-program `--user` option handle the post-subcommand case unchanged.

## Test plan
- [x] `npm run check` clean
- [x] `npm test` — 1592 / 1592 pass (4 removed `stripUserFlag` tests now covered by cli-core)
- [x] `npm run build` succeeds
- [x] Smoke: `td task list --user scott@doist.com` works (post-subcommand `--user` still parsed by commander's root option)

### DIFF
--- a/src/lib/auth-store.ts
+++ b/src/lib/auth-store.ts
@@ -1,8 +1,9 @@
-import type { AuthAccount, TokenStore } from '@doist/cli-core/auth'
+import type { AccountRef, AuthAccount, TokenStore } from '@doist/cli-core/auth'
 import {
     clearApiToken,
     loadTokenForStoredUser,
     NoTokenError,
+    setDefaultUserId,
     type TokenStorageResult,
     TOKEN_ENV_VAR,
     upsertUser,
@@ -10,7 +11,7 @@ import {
 } from './auth.js'
 import { type AuthFlag, type AuthMode, readConfig, type StoredUser } from './config.js'
 import { SecureStoreUnavailableError } from './secure-store.js'
-import { getDefaultUser, getStoredUsers } from './users.js'
+import { findUserByRef, getDefaultUser, getStoredUsers } from './users.js'
 
 /**
  * Account shape stored by todoist-cli. Extends cli-core's `AuthAccount` with
@@ -95,23 +96,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
 
     return {
         /**
-         * Pure view of persisted state. Deliberately ignores the global
-         * `--user` selector (a CLI-invocation concern, not storage) and
-         * returns `null` when `TODOIST_API_TOKEN` is in play (env tokens
-         * don't represent a persisted account). When multiple accounts
-         * are stored without a default, returns `null` rather than
-         * throwing a selection error — the runtime caller can react to
-         * "no active persisted account" but shouldn't see CLI-arg errors
-         * leaking out of a store read.
+         * Pure view of persisted state. Returns `null` when `TODOIST_API_TOKEN`
+         * is in play (env tokens don't represent a persisted account). With
+         * `ref`, returns that specific stored account; without, returns the
+         * default-or-only account. A `ref` that does not match any stored
+         * account returns `null` so cli-core's resolver layer can translate
+         * the miss into a typed error without storage reads leaking CLI-arg
+         * errors.
          */
-        async active() {
+        async active(ref?: AccountRef) {
             if (process.env[TOKEN_ENV_VAR]) return null
 
             const config = await readConfig()
             const users = getStoredUsers(config)
             if (users.length === 0) return null
 
-            const target = getDefaultUser(config) ?? (users.length === 1 ? users[0] : null)
+            let target: StoredUser | null
+            if (ref !== undefined) {
+                target = findUserByRef(config, ref)?.user ?? null
+            } else {
+                target = getDefaultUser(config) ?? (users.length === 1 ? users[0] : null)
+            }
             if (!target) return null
 
             try {
@@ -134,8 +139,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
             lastStorageResult = result
         },
 
-        async clear() {
-            lastClearResult = await clearApiToken()
+        async clear(ref?: AccountRef) {
+            lastClearResult = await clearApiToken({ ref })
+        },
+
+        async list() {
+            const config = await readConfig()
+            const users = getStoredUsers(config)
+            if (users.length === 0) return []
+            const defaultUser = getDefaultUser(config)
+            const defaultId = defaultUser?.id ?? (users.length === 1 ? users[0].id : undefined)
+            return users.map((user) => ({
+                account: storedUserToAccount(user),
+                isDefault: user.id === defaultId,
+            }))
+        },
+
+        async setDefault(ref: AccountRef) {
+            // `setDefaultUserId` accepts any ref (id or email) and throws
+            // `UserNotFoundError` (a `CliError` with code `USER_NOT_FOUND`)
+            // when the ref does not match any stored account.
+            await setDefaultUserId(ref)
         },
 
         getLastStorageResult() {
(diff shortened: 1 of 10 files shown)


doistbot (commented):
This PR nicely bumps `@doist/cli-core` to 0.12.0, wiring up the new multi-user `TokenStore` contract and adopting the upstream `stripUserFlag` implementation. These updates cleanly bridge our existing per-user primitives with the reshaped library interface. There are just a few details to review around validating references to prevent accidental legacy token clearance, consolidating duplicated default-user fallback logic, distinguishing between lookup misses and credential read errors, and expanding test coverage for the newly implemented methods.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F340&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1044f240-506c-11f1-9a14-dee75c99c75b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1044f240-506c-11f1-9a14-dee75c99c75b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth-store.ts
```diff
@@ -134,8 +139,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
             lastStorageResult = result
         },
 
-        async clear() {
-            lastClearResult = await clearApiToken()
+        async clear(ref?: AccountRef) {
+            lastClearResult = await clearApiToken({ ref })
```
doistbot, line 130: [P2] Forwarding `ref` straight into `clearApiToken()` changes logout behavior on legacy single-user installs. In `clearApiToken`, the legacy fallback runs before `requestedRef` is validated, so a call like `td auth logout --user missing@example.com` will clear the legacy token instead of raising `USER_NOT_FOUND`. Please validate the ref before taking the legacy-clear path (or avoid passing `ref` through here until that fallback is fixed).

src/lib/auth-store.ts
```diff
@@ -134,8 +139,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
...
+        },
+
+        async list() {
+            const config = await readConfig()
+            const users = getStoredUsers(config)
+            if (users.length === 0) return []
+            const defaultUser = getDefaultUser(config)
+            const defaultId = defaultUser?.id ?? (users.length === 1 ? users[0].id : undefined)
```
doistbot: [P2] The fallback logic to determine the effective default user (`getDefaultUser` falling back to the only stored user) is duplicated here and on line 118 in `active()`. Consider extracting a helper like `getEffectiveDefaultUser(config)` in `src/lib/users.ts` to DRY this up. (This new helper could also replace the identical fallback blocks in `resolveActiveUser` and `clearApiToken` over in `src/lib/auth.ts`).

src/lib/auth-store.ts
```diff
@@ -95,23 +96,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
...
+         * Pure view of persisted state. Returns `null` when `TODOIST_API_TOKEN`
+         * is in play (env tokens don't represent a persisted account). With
+         * `ref`, returns that specific stored account; without, returns the
+         * default-or-only account. A `ref` that does not match any stored
+         * account returns `null` so cli-core's resolver layer can translate
+         * the miss into a typed error without storage reads leaking CLI-arg
+         * errors.
          */
-        async active() {
+        async active(ref?: AccountRef) {
```
doistbot, line 108: [P1] Adding `ref` here makes `null` mean two different things: "ref did not match any stored account" and "matched account but its token/keyring could not be read" (the existing catch still collapses `NoTokenError` / `SecureStoreUnavailableError` to `null`). cli-core uses `null` for the lookup-miss path, so a real account with an unavailable credential will now be misreported as an unknown user. Only return `null` for an actual ref miss; once a user was matched, propagate a typed auth/storage error instead.

src/lib/auth-store.ts
```diff
@@ -134,8 +139,27 @@ export function createTodoistTokenStore(): TodoistTokenStore {
             lastStorageResult = result
         },
 
-        async clear() {
-            lastClearResult = await clearApiToken()
+        async clear(ref?: AccountRef) {
+            lastClearResult = await clearApiToken({ ref })
+        },
+
+        async list() {
```
doistbot, line 133: [P2] The new multi-user `TokenStore` contract methods introduced here are entirely untested.

Please update `auth-store.test.ts` to cover `list()` (especially the `isDefault` fallback logic), `setDefault()` (which will require adding `setDefaultUserId` to the `./auth.js` vi.mock in that test file), and the new `ref` arguments on `active()` and `clear()`.

src/lib/auth-store.ts
scottlovegrove: Fixed in fd71eec — dropped the swallow on `loadTokenForStoredUser` failures so `NoTokenError` / `SecureStoreUnavailableError` propagate once a user is matched. `null` now strictly means "ref miss" per cli-core's contract. New tests cover both the matched-user-broken-keyring case (with and without `--user`) and the ref-miss case.

src/lib/auth-store.ts
scottlovegrove: Fixed in fd71eec — reordered `clearApiToken` so `requestedRef` is checked before the legacy-config fallback. `td auth logout --user missing@x` now throws `USER_NOT_FOUND` instead of wiping the legacy token. Verified the existing `auth.test.ts` suite still passes.

src/lib/auth-store.ts
scottlovegrove: Done in fd71eec — added `getEffectiveDefaultUser(config)` in `src/lib/users.ts` and routed both `active()` and `list()` through it. Left `resolveActiveUser` / `clearApiToken` in `auth.ts` alone for now — their fallback shapes are more nuanced (multi-user-no-default throws `NoUserSelectedError`, default-pointing-at-missing-user fallback, legacy-config fallback), so retrofitting them deserves its own pass.

src/lib/auth-store.ts
scottlovegrove: Added in fd71eec — extended `auth-store.test.ts` with coverage for: `list()` (empty, single-user implicit default, multi-user explicit default, multi-user no default → none marked), `setDefault()` (delegates to `setDefaultUserId` + `UserNotFoundError` propagation), `active(ref)` (id match, email match, ref miss → null, matched-but-broken-keyring → throws), and `clear(ref)` (forwarded vs absent ref). `setDefaultUserId` mocked alongside the existing `./auth.js` mock.

PR #341, opened 2026-05-16
refactor(auth): adopt cli-core 0.16 keyring TokenStore

scottlovegrove:
## What gets deleted

- `src/lib/secure-store.ts` — replaced by cli-core's `createSecureStore` / `SecureStoreUnavailableError`
- The multi-account write path inside `src/lib/auth.ts` (`upsertUser`, `clearApiToken`, `removeUserById`, `setDefaultUserId`, `loadTokenForStoredUser`, `ensureV2Shape`, `stripLegacyAuthFields`, `buildFallbackWarning`, `buildConfigCleanupWarning`) — cli-core's `createKeyringTokenStore` owns this surface now
- `src/lib/auth.ts` `resolveLegacyToken` + `clearLegacyToken` — the v1 fallback for unmigrated installs (see "Behavioural changes" below)
- `src/commands/auth/token-view.ts` — replaced by cli-core's `attachTokenViewCommand`
- Unused `users.ts` exports (`getDefaultUser`, `getEffectiveDefaultUser`, `FindUserResult`, `UserRef`)
- ~14 paranoid / cli-core-covered tests across `auth.test.ts`, `auth-store.test.ts` (whole file), `user-records.test.ts`, `migrate-auth.test.ts`, and `commands/auth/auth.test.ts` (token-view + legacy-fallback cases)

## What gets added

- `src/lib/user-records.ts` — `UserRecordStore<TodoistAccount>` adapter cli-core's `createKeyringTokenStore` reads/writes through. REPLACE-not-merge `upsert`; every write runs through `ensureV2` (stamps `config_version`, strips legacy fields). Array manipulation delegates to `users.ts`.
- `src/lib/migrate-auth.ts` rewrite — thin wrapper around cli-core's `migrateLegacyAuth` supplying the Todoist-specific callbacks (`/api/v1/user` `identifyAccount`, `loadLegacyPlaintextToken`, `cleanupLegacyConfig`). One-way gate is `config.config_version === CONFIG_VERSION`, which survives logout so a reinstall over a logged-out v2 install can't re-migrate a stale legacy slot.
- `src/commands/auth/store-wrap.ts` (`withUserRefAware`) — shared wrap that substitutes `getRequestedUserRef()` into `active`/`clear` and turns ref-misses into `UserNotFoundError`. Used by both `auth logout` and `auth token view` because `index.ts` strips `--user` from argv before commander runs.
- `src/lib/config.ts` `stripLegacyAuthFields` — single source of truth for the v1 fields scrubbed on every v2 write.
- Adapter-level tests in `user-records.test.ts` (REPLACE-not-merge, `ensureV2`, default-pointer cleanup on remove) and `migrate-auth.test.ts` (the consumer callbacks + `toLocalResult` translation).

## Behavioural changes

- **Dropped the v1 read + clear fallback.** A user on an unmigrated v1 install whose postinstall never ran (offline first-run only) now has to run `td auth login` again instead of staying authenticated. Acceptable one-time UX cost — anyone who's seen ≥ one release with postinstall migration is already on v2.
- **`td auth token` no longer prints "Updated stored token for ..." vs "Saved token for ...".** cli-core's `TokenStore.set` is `void`, so the verb distinction is dropped. Always prints "Saved token for &lt;email&gt;".
- **`td user use no-such-user` exits with `ACCOUNT_NOT_FOUND` (cli-core's code) instead of `USER_NOT_FOUND`.** Same UX; different code string.

## File map

| File | Status | Role |
|---|---|---|
| `src/lib/secure-store.ts` | **deleted** | replaced by `@doist/cli-core/auth.createSecureStore` |
| `src/commands/auth/token-view.ts` | **deleted** | replaced by `@doist/cli-core/auth.attachTokenViewCommand` |
| `src/lib/auth.ts` | gutted (320 → ~165 LOC) | read-side only: `resolveActiveUser`, `getApiToken`, `probeApiToken`, `getAuthMetadata`, `listStoredUsers`, `NoTokenError` |
| `src/lib/auth-store.ts` | rewritten | `createTodoistTokenStore()` + shared `SERVICE_NAME` / `LEGACY_ACCOUNT` / `accountForUser` constants + `toTodoistAccount` |
| `src/lib/migrate-auth.ts` | rewritten | cli-core wrapper |
| `src/lib/user-records.ts` | **new** | `UserRecordStore<TodoistAccount>` adapter |
| `src/commands/auth/store-wrap.ts` | **new** | shared `--user` injection wrap |
| `src/commands/auth/logout.ts` | rewritten | wraps the store with `withUserRefAware` |
| `src/lib/users.ts` | slimmed | drops 4 unused exports |
| `src/commands/user/{use,remove}.ts` | slimmed | route directly through the store |
| `CODEBASE.md` | updated | auth-storage section rewritten |

## Test plan

- [x] `npm run type-check`, `npm test` (1564/1564), `npm run check` — all clean
- [ ] Manual smoke on Linux with libsecret available: `td auth login` writes to `user-<id>` keyring slot, `td auth logout` removes it
- [ ] Manual smoke on WSL without D-Bus: fallback warning on stderr, `users[0].api_token` set in config
- [ ] Manual smoke v1 → v2 migration: hand-craft a v1 config (`{ "api_token": "...", "auth_mode": "read-write" }`), confirm postinstall migrates it and sets `config_version: 2`
- [ ] Manual smoke `td auth logout --user <ref>`: ensure the explicit ref is honoured (regression covered by the review fixup)


src/lib/auth.ts
```diff
@@ -211,181 +177,6 @@ export async function getAuthMetadata(): Promise<AuthMetadata> {
...
-}
-
-/**
- * Remove the active user (resolved via `--user` or default). For multi-user
- * installs without a default, callers must pass `--user <ref>` to disambiguate.
- */
-export async function clearApiToken(opts: { ref?: string } = {}): Promise<TokenStorageResult> {
-    const config = await readConfig()
-    const users = getStoredUsers(config)
-    const requestedRef = opts.ref ?? getRequestedUserRef()
```
doistbot, line 304: [P1] Removing this `getRequestedUserRef()` path drops Todoist CLI’s `--user <id|email>` handling from logout. `index.ts` strips `--user` out of `process.argv` before commander runs, so the generic `attachLogoutCommand`/`TokenStore.clear()` path cannot recover it on its own. As written, `td auth logout --user ...` will now act on the default/sole account instead of the requested one. Please keep a local wrapper that threads the parsed selector into clear, or preserve an equivalent `clearApiToken({ ref: getRequestedUserRef() })` path.

src/lib/auth.ts
```diff
@@ -462,60 +256,6 @@ async function resolveLegacyToken(config: Config): Promise<ResolvedUser> {
     throw new NoTokenError()
 }
 
-async function clearLegacyToken(config: Config): Promise<TokenStorageResult> {
```
doistbot, line 465: [P2] Dropping the legacy-clear path leaves logout inconsistent with the still-present legacy read fallback above. `resolveActiveUser()` can continue authenticating from top-level `api_token` / the legacy `api-token` keyring slot for one release, but `td auth logout` no longer clears either of those sources. On an unmigrated/offline-upgrade install, logout will not actually log the user out. Please retain a legacy-aware clear path until the v1 fallback reader is removed.

src/lib/user-records.ts
```diff
@@ -0,0 +1,120 @@
...
+        ...rest
+    } = config
+    return {
+        ...rest,
+        config_version: CONFIG_VERSION,
+        users: Array.isArray(rest.users) ? rest.users : [],
+    }
+}
+
+function withoutDefaultUser(config: Config): Config {
```
doistbot: [P2] Missed reuse opportunity: this helper and the array-manipulation logic in `upsert()`, `remove()`, and `setDefaultId()` below reinvent the `removeStoredUser`, `upsertStoredUser`, and `setDefaultUser` functions already available in `src/lib/users.ts`. You can reuse those existing helpers to keep config mutation logic centralized (and export a `clearDefaultUser` from there for the `id === null` case).

src/lib/migrate-auth.ts
```diff
@@ -30,124 +31,70 @@ interface MigrateOptions {
...
+                id: user.id,
+                email: user.email,
+                authMode: config.auth_mode ?? 'unknown',
+                authScope: config.auth_scope,
+                authFlags: config.auth_flags,
+            })
+        },
+        cleanupLegacyConfig: async () => {
+            const config = await readConfig()
+            const {
```
doistbot: [P2] This destructured list of legacy fields duplicates the cleanup logic in `ensureV2` (`src/lib/user-records.ts`). Consider preserving the old `stripLegacyAuthFields` helper (e.g., in `src/lib/config.ts`) instead of duplicating the knowledge of which legacy fields to strip across multiple files.

src/lib/user-records.ts
```diff
@@ -0,0 +1,120 @@
...
+    }
+    if (record.fallbackToken !== undefined) {
+        stored.api_token = record.fallbackToken
+    }
+    return stored
+}
+
+function storedUserToRecord(user: StoredUser): UserRecord<TodoistAccount> {
+    const record: UserRecord<TodoistAccount> = {
+        account: {
```
doistbot: [P2] `storedUserToRecord()` is rebuilding the `TodoistAccount` shape instead of reusing `toTodoistAccount()`, which this PR already positions as the single source of truth for that mapping. Please delegate here as well so `label`, the `'unknown'` auth-mode fallback, and any future account fields stay aligned across login, migration, and record-store reads.

src/lib/user-records.ts
```diff
@@ -0,0 +1,120 @@
+import type { UserRecord, UserRecordStore } from '@doist/cli-core/auth'
```
doistbot, line 1: [P2] Because this PR changes the auth/token storage architecture and drops `secure-store.ts`, please update `CODEBASE.md` to reflect these structural changes. `AGENTS.md` explicitly lists changes to auth/token storage and the addition/removal of shared helpers as triggers for updating the map.

doistbot (commented):
This PR effectively transitions the CLI's auth and token storage to the new `@doist/cli-core` primitives, drastically reducing local boilerplate and centralizing our keyring interactions. The architectural simplification is a great improvement for long-term maintainability, cleanly aligning our storage mechanisms with the core library. A few adjustments are needed to finalize the implementation, including restoring user-specific targeting and legacy cleanup during logout, resolving some logic duplication around user records and account shapes, filling test coverage gaps for the new adapter, and updating the codebase documentation to reflect the structural changes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F341&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9d114750-5143-11f1-855e-43733262b062%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9d114750-5143-11f1-855e-43733262b062&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/config/view.ts
```diff
@@ -1,3 +1,6 @@
+import { SecureStoreUnavailableError } from '@doist/cli-core/auth'
+
+const SECURE_STORE_DESCRIPTION = 'system credential manager'
```
doistbot: [P3] This constant declaration is interspersed between imports. Please move `SECURE_STORE_DESCRIPTION` below all the `import` statements to match standard grouping conventions.

src/lib/auth.ts
```diff
@@ -18,31 +13,20 @@ export {
...
-    removeStoredUser,
-    setDefaultUser as setDefaultUserInConfig,
-    upsertStoredUser,
-    UserNotFoundError,
-} from './users.js'
+import { findUserByRef, getStoredUsers, NoUserSelectedError, UserNotFoundError } from './users.js'
 
 export const TOKEN_ENV_VAR = 'TODOIST_API_TOKEN'
 
+const SERVICE_NAME = 'todoist-cli'
```
doistbot: [P2] `SERVICE_NAME`/`LEGACY_ACCOUNT` are now hardcoded here, while the same persisted identifiers are also hardcoded in `auth-store.ts` and `migrate-auth.ts`. These strings are part of the keyring/config ABI, so they should live in one shared auth-storage constant/helper instead of being duplicated across read, write, and migration paths.

src/lib/auth-store.ts
```diff
@@ -48,112 +41,19 @@ export function toTodoistAccount(input: TodoistAccountInput): TodoistAccount {
...
-
-        getLastStorageResult() {
-            return lastStorageResult
-        },
-
-        getLastClearResult() {
-            return lastClearResult
-        },
-    }
+    return createKeyringTokenStore<TodoistAccount>({
```
doistbot: [P2] This adapter is depending on `createKeyringTokenStore`'s default `accountForUser` and `matchAccount` behavior for Todoist's persisted storage format. Please pass those callbacks explicitly here instead of encoding the assumption in a comment; otherwise a future cli-core default change could silently start writing different keyring slots or stop resolving email refs for existing installs.

src/lib/migrate-auth.ts
```diff
@@ -30,124 +31,70 @@ interface MigrateOptions {
...
+        serviceName: 'todoist-cli',
+        legacyAccount: 'api-token',
+        userRecords: createTodoistUserRecordStore(),
+        silent: opts.silent,
+        logPrefix: 'todoist-cli',
+        // `config_version === CONFIG_VERSION` is the durable, one-way gate. It
+        // survives logout (which clears `users[]` but leaves `config_version`),
+        // so a reinstall over a logged-out v2 install cannot re-migrate a
+        // stale legacy keyring entry.
+        hasMigrated: async () => (await readConfig()).config_version === CONFIG_VERSION,
```
doistbot: [P2] This wrapper now reparses `config.json` independently in `hasMigrated`, `markMigrated`, `loadLegacyPlaintextToken`, `identifyAccount`, and `cleanupLegacyConfig`. A successful migration will hit several of these callbacks back-to-back, so one postinstall run does multiple reads of the same file where the previous implementation worked from one snapshot. Cache the config in this wrapper and refresh that cache after each `writeConfig()` to avoid the extra I/O.

src/lib/user-records.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * `UserRecordStore<TodoistAccount>` backed by the existing JSON config file.
+ *
+ * REPLACE, do not merge: `upsert` rebuilds the `StoredUser` entirely from the
+ * incoming record so a stale `api_token` from a prior offline-fallback write
+ * cannot outlive a later keyring-online write (cli-core preferentially reads
+ * `fallbackToken` over the keyring).
+ */
+export function createTodoistUserRecordStore(): UserRecordStore<TodoistAccount> {
```
doistbot, line 70: [P1] The new `createTodoistUserRecordStore` adapter lacks unit tests. The PR deletes `auth-store.test.ts` and config-modification tests from `auth.test.ts` (e.g., `upsertUser` and `removeUserById`), leaving the config read/write, default user tracking, and legacy stripping logic in this file uncovered. Add `src/lib/user-records.test.ts` to restore coverage for this critical configuration logic.

src/lib/migrate-auth.test.ts
```diff
@@ -51,27 +48,28 @@ describe('migrateLegacyAuth', () => {
...
+        // Stub cli-core's migration so we observe what the wrapper passes in
+        // and can drive each consumer callback directly.
+        vi.doMock('@doist/cli-core/auth', async (importOriginal) => {
+            const actual = await importOriginal<typeof import('@doist/cli-core/auth')>()
+            return {
+                ...actual,
+                migrateLegacyAuth: vi.fn(
+                    async (options: MigrateLegacyAuthOptions<TodoistAccount>) => {
+                        capturedOptions = options
+                        return { status: 'no-legacy-state' as const }
```
doistbot: [P2] The mocked `migrateLegacyAuth` is hardcoded to always return `{ status: 'no-legacy-state' }`. This prevents testing the wrapper's `toLocalResult` translation logic, leaving the 'migrated' and 'skipped' mapping branches completely uncovered. Consider allowing tests to override the mock's return value to verify these branches.

src/commands/auth/auth.test.ts
```diff
@@ -444,46 +434,17 @@ describe('auth command', () => {
...
-
-            expect(mockClearApiToken).toHaveBeenCalled()
-            expect(consoleSpy).toHaveBeenCalledWith('✓ Logged out')
+        beforeEach(() => {
+            clearMock.mockReset().mockResolvedValue(undefined)
+            lastClearMock.mockReset().mockReturnValue({ storage: 'secure-store' })
         })
 
-        it('surfaces keyring-fallback warning to stderr in plain mode', async () => {
+        it('surfaces keyring-fallback warning to stderr', async () => {
```
doistbot, line 436: [P2] This rewrite drops the `auth logout --json` and `--ndjson` cases even though `attachTodoistLogoutCommand` still has Todoist-local branching on `view.json || view.ndjson`. Please keep one machine-output assertion here so we catch regressions where the warning path starts writing human text to stdout in JSON/NDJSON modes.

src/lib/auth-store.ts
doistbot: [P2] This rewires auth/token storage to a new cli-core-backed path, but `CODEBASE.md` wasn’t updated. Repo guidance explicitly calls out auth/token storage changes and new broadly reusable helpers under `src/lib/` as triggers for updating that map; it still documents the deleted `secure-store.ts` and the old `auth-store.ts` behavior.

src/lib/auth.ts
scottlovegrove: Fixed in 8c93cd5. `attachTodoistLogoutCommand` now wraps the store to thread `getRequestedUserRef()` into `store.clear()` and pre-checks via `active()` so a non-matching ref surfaces `UserNotFoundError` instead of silently no-op-ing. Covered by the new `threads --user from global args through to store.clear` + `throws UserNotFoundError when --user does not match` tests in `commands/auth/auth.test.ts`.

src/lib/auth.ts
scottlovegrove: Restored in 8c93cd5 as `clearLegacyToken()` in `auth.ts`, invoked by the logout wrap when cli-core's v2 clear is a no-op (empty store + no explicit ref). Pairs with the still-present `resolveLegacyToken` read fallback; both drop together when the v1 fallback reader is removed next cycle. Covered by `falls back to clearLegacyToken on an unmigrated install`.

src/lib/user-records.ts
scottlovegrove: Done in 8c93cd5. The adapter now delegates to `upsertStoredUser` / `removeStoredUser` / `setDefaultUser` from `users.ts` (plus a new `clearDefaultUser` for the `id === null` case).

src/lib/migrate-auth.ts
scottlovegrove: Done in 8c93cd5. `stripLegacyAuthFields` lives in `config.ts` now, called from both `user-records.ts`'s `ensureV2` and migrate-auth's `cleanupLegacyConfig`.

src/lib/user-records.ts
scottlovegrove: Done in 8c93cd5. `storedUserToRecord` now calls `toTodoistAccount` for the record→account mapping.

src/lib/user-records.ts
scottlovegrove: Updated in 8c93cd5. The auth & token storage section was rewritten — covers the new `auth-store.ts` / `user-records.ts` / `migrate-auth.ts` layout, the read/write boundary, and the migration gate.

src/commands/config/view.ts
scottlovegrove: Moved below imports in 8c93cd5.

src/lib/auth.ts
scottlovegrove: Centralized in 8c93cd5. `SERVICE_NAME`, `LEGACY_ACCOUNT`, and `accountForUser` live in `auth-store.ts`; `auth.ts` and `migrate-auth.ts` import them.

src/lib/auth-store.ts
scottlovegrove: Done in 8c93cd5. Both passed explicitly. `matchAccount` is case-insensitive id-or-label so the prior `findUserByRef` semantics (case-insensitive email) are pinned regardless of any future cli-core default change.

src/lib/migrate-auth.ts
scottlovegrove: Done in 8c93cd5. The wrapper now memoizes `readConfig()` and refreshes on `writeConfig()`, so one successful postinstall run does one read+write pass instead of four reparses.

src/lib/user-records.ts
scottlovegrove: Added `src/lib/user-records.test.ts` in 8c93cd5: REPLACE-not-merge upsert (stale `fallbackToken` dropped), `fallbackToken` preserved when supplied, append-without-overwrite, `ensureV2` legacy-fields strip on every write, `remove` cleaning the default pointer when it matched the removed user, `setDefaultId(null)` dropping the wrapper, and `list` translation.

src/lib/migrate-auth.test.ts
scottlovegrove: Done in 8c93cd5. The mocked `migrateLegacyAuth` now reads a per-test `coreResult` variable. Added a `toLocalResult translation` describe block covering the `migrated` (account → migratedUserId/migratedEmail), `skipped` (reason + detail flattened), and `already-migrated` passthrough branches.

src/commands/auth/auth.test.ts
scottlovegrove: Re-added in 8c93cd5. New `keeps the keyring-fallback warning on stderr in --json mode (and out of the JSON envelope)` test pins the view-mode branching in `onCleared` so a regression that leaked human prose into stdout would surface.

src/lib/auth-store.ts
scottlovegrove: Updated in 8c93cd5 — same fix as the other CODEBASE.md item: the auth & token storage section now reflects the new cli-core-backed layout.

scottlovegrove:
@doistbot /review

src/commands/auth/store-wrap.ts
```diff
@@ -0,0 +1,28 @@
...
+ * instead of cli-core's null-on-miss contract, so `logout --user mistake`
+ * doesn't silently print `✓ Logged out`.
+ */
+export function withUserRefAware(store: TodoistTokenStore): TodoistTokenStore {
+    return {
+        ...store,
+        active: (ref) => store.active(ref ?? getRequestedUserRef()),
+        async clear(ref) {
+            const targetRef = ref ?? getRequestedUserRef()
+            if (targetRef !== undefined && !(await store.active(targetRef))) {
```
doistbot: [P2] Using `store.active()` to verify if a user exists is risky because it actively attempts to load the token. If the keyring is offline or the token is missing, `active()` will propagate `SecureStoreUnavailableError` or `NoTokenError` instead of returning null. This will cause `td auth logout --user <ref>` to crash rather than clearing the broken credential. Consider checking the underlying config or `store.list()` for existence instead to avoid side effects.

src/lib/auth-store.ts
```diff
@@ -48,112 +55,36 @@ export function toTodoistAccount(input: TodoistAccountInput): TodoistAccount {
...
```
doistbot: [P1] Delegating `active()` straight to `createKeyringTokenStore()` drops the old `TODOIST_API_TOKEN` short-circuit. `attachTodoistStatusCommand()` still assumes `store.active()` returns `null` in env-token mode; with both an env token and a stored account present, `td auth status` will fetch the user via the persisted token while `getAuthMetadata()` reports `source: 'env'`, so the status output can describe the wrong account. Keep a small local wrapper around `active()` that returns `null` when `TODOIST_API_TOKEN` is set.

src/lib/auth-store.ts
```diff
@@ -48,112 +55,36 @@ export function toTodoistAccount(input: TodoistAccountInput): TodoistAccount {
...
- * since cli-core's `TokenStore.set` signature is `void`).
+ * Case-insensitive email match on top of cli-core's default id-or-label
+ * equality. `findUserByRef` (in `users.ts`) historically lowercased the email
+ * compare; pin that behaviour explicitly here so a cli-core default change
+ * can't drop it.
  */
-export type TodoistTokenStore = TokenStore<TodoistAccount> & {
-    getLastStorageResult(): TokenStorageResult | undefined
-    getLastClearResult(): TokenStorageResult | undefined
+function matchTodoistAccount(account: TodoistAccount, ref: AccountRef): boolean {
```
doistbot: [P2] `matchTodoistAccount()` is now a second implementation of the CLI's user-ref matching rules. `findUserByRef()` in `src/lib/users.ts` already owns exact-id + case-insensitive email matching, and the two copies have already drifted (`findUserByRef` trims the ref, this one does not). Please extract a small shared matcher and use it from both places so `--user <ref>` resolution stays consistent between config reads and keyring-store lookups.

src/commands/auth/index.ts
```diff
@@ -1,43 +1,40 @@
...
         .command('token [token]')
         .description('Save API token for CLI authentication (or use a subcommand: `view`)')
         .action(loginWithToken)
 
-    tokenCmd
-        .command('view')
-        .description(
+    attachTokenViewCommand(tokenCmd, {
+        name: 'view',
+        store: withUserRefAware(store),
```
doistbot: [P3] We're now constructing `withUserRefAware(store)` here and again inside `attachTodoistLogoutCommand()`. Since this wrapper is just the Todoist-specific adapter for cli-core registrars, please hoist one `refAwareStore` in `registerAuthCommand()` and reuse it for both `token view` and `logout` so the `--user` adaptation lives in a single place.

src/commands/auth/store-wrap.ts
```diff
@@ -0,0 +1,28 @@
...
+ * `attachTokenViewCommand`) can't see the flag on their parsed args; this
+ * wrap puts the global selector back into play.
+ *
+ * Also turns "ref didn't match anything" into a typed `UserNotFoundError`
+ * instead of cli-core's null-on-miss contract, so `logout --user mistake`
+ * doesn't silently print `✓ Logged out`.
+ */
+export function withUserRefAware(store: TodoistTokenStore): TodoistTokenStore {
+    return {
+        ...store,
```
doistbot: [P2] Spreading `...store` to wrap the `TokenStore` instance is unsafe. If `createKeyringTokenStore` from `@doist/cli-core` returns a class instance, prototype methods like `getLastClearResult` will not be copied to the wrapper, causing `refAware.getLastClearResult()` to throw at runtime (even if the POJO mock works in tests). Additionally, spreading unbinds `this`, which breaks if any store methods rely on it. Consider delegating explicitly using `Object.assign(Object.create(store), { active, clear })` or a `Proxy`.

src/commands/auth/auth.test.ts
```diff
@@ -444,124 +433,84 @@ describe('auth command', () => {
...
                 ).rejects.toHaveProperty('code', 'USER_NOT_FOUND')
+                expect(clearMock).not.toHaveBeenCalled()
             } finally {
                 process.argv = originalArgv
                 resetGlobalArgs()
             }
-            expect(consoleSpy).not.toHaveBeenCalled()
+        })
+
+        it('keeps the keyring-fallback warning on stderr in --json mode (and out of the JSON envelope)', async () => {
```
doistbot: [P2] This replacement only checks that one human string is absent from stdout. It no longer proves that `td auth logout --json` still emits the expected machine-readable payload. If stdout went empty or the JSON envelope changed/broke, this test would still pass. Please keep a positive assertion on the JSON body in addition to the stderr-warning separation.

src/commands/auth/auth.test.ts
```diff
@@ -444,124 +433,84 @@ describe('auth command', () => {
...
+
+            await program.parseAsync(['node', 'td', 'auth', 'logout', '--json'])
+
+            const stdout = consoleSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('\n')
+            expect(stdout).not.toContain('Stored token removed')
+            expect(errorSpy).toHaveBeenCalledWith('Warning:', WARNING_RESULT.warning)
         })
     })
+
+    // `token view` is now wired via cli-core's `attachTokenViewCommand`;
```
doistbot: [P2] The `--user` wrap is not actually covered by the logout cases above. Those only exercise `withUserRefAware.clear()`, while `td auth token view` goes through `withUserRefAware.active()`. Since switching `token view` to cli-core is a key change in this PR, add one focused `auth token view --user <ref>` case (with stripped `process.argv`) so the local read-side wrapper is still tested.

src/lib/migrate-auth.test.ts
```diff
@@ -1,35 +1,37 @@
...
-        e = { token: null }
-        state.entries.set(account, e)
-    }
-    return e
-}
-
-describe('migrateLegacyAuth', () => {
+const TEST_CONFIG_PATH = `${TEST_HOME}/.config/todoist-cli/config.json`
+
+// Capture the options object the wrapper passes into cli-core. Each test
```
doistbot, line 16: [P2] This adapter suite now covers plaintext loading, account identification, and cleanup, but it still never drives the new Todoist-specific migration gate callbacks (`hasMigrated` / `markMigrated`). Those are the local behaviors that prevent re-migrating a logged-out v2 install, so please add a couple of direct `capturedOptions` assertions to pin that contract.

doistbot (commented):
This update successfully transitions our authentication setup to the new `@doist/cli-core` keyring primitives, replacing complex local implementations with a much cleaner adapter pattern. Shrinking this custom auth surface significantly improves long-term maintainability and standardizes our credential management. There are just a few details to refine before merging, primarily around preserving the environment variable short-circuit, avoiding crashes from offline keyring checks, safely wrapping the store instance without losing prototype methods, consolidating user-reference matching, and expanding test coverage to capture JSON outputs, migration gates, and specific flag interactions.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F341&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A27b14030-514f-11f1-843b-3404a06a5f73%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A27b14030-514f-11f1-843b-3404a06a5f73&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/store-wrap.ts
scottlovegrove: Fixed in 7ea3440. Existence check now runs through `store.list()` (no token load, no `SecureStoreUnavailableError` risk). Implementation matches accounts via the shared `matchUserRef` helper extracted in this commit.

src/lib/auth-store.ts
scottlovegrove: Fixed in 7ea3440. `createTodoistTokenStore.active()` now short-circuits to `null` when `TODOIST_API_TOKEN` is set so `attachTodoistStatusCommand`'s `onNotAuthenticated` branch still fires, keeping the env-mode status display aligned with `getAuthMetadata().source === 'env'`. `TOKEN_ENV_VAR` lives in `auth-store.ts` now (re-exported from `auth.ts` for back-compat).

src/lib/auth-store.ts
scottlovegrove: Fixed in 7ea3440. Extracted `matchUserRef(user, ref)` into `users.ts` (single source of truth: id-exact or case-insensitive trimmed-email). `findUserByRef`, the `store-wrap` existence check, and the keyring store's `matchAccount` all delegate to it — the trim-drift is gone.

src/commands/auth/index.ts
scottlovegrove: Done in 7ea3440. `registerAuthCommand` builds one `refAware = withUserRefAware(store)` and passes it to both `attachTodoistLogoutCommand` and `attachTokenViewCommand`. Logout no longer constructs its own wrap.

src/commands/auth/store-wrap.ts
scottlovegrove: Switched to `Object.assign(Object.create(store) as TodoistTokenStore, { active, clear })` in 7ea3440 so any cli-core method that's later promoted to a prototype still resolves via the prototype chain instead of being silently dropped.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 7ea3440. The `--json` logout test now positively asserts `stdout` contains `"ok": true` in addition to the negative "Stored token removed" check, so an empty stdout or broken envelope would surface.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 7ea3440. Two new tests in `commands/auth/auth.test.ts > token view subcommand`: `routes --user from global args through the active() path` (ref matched → token printed, `activeMock` called with the global ref) and `surfaces UserNotFoundError when --user does not match` (ref missed → `USER_NOT_FOUND`, `activeMock` never called). The wrap's `active` path is now covered explicitly, not just transitively via logout.

src/lib/migrate-auth.test.ts
scottlovegrove: Done in 7ea3440. Added a `hasMigrated / markMigrated (the one-way gate)` describe block in `migrate-auth.test.ts`: gate returns `true` when `config_version === CONFIG_VERSION`, `false` on pre-v2 configs, and `markMigrated` stamps `config_version: 2` without touching the legacy fields (that's `cleanupLegacyConfig`'s job).

pedroalves0 (commented):
✅

PR #349, opened 2026-05-18
feat: add section reorder command

rmartins90:
## Overview

Adds `td section reorder` so sections can be reordered within a project from the CLI. It supports positional moves, plus relative moves before or after another section.

## Changes

- Added `td section reorder` with `--project`, `--section`, `--before`, `--after`, `--position`, `--json`, and `--dry-run`.
- Added a Sync API helper for Todoist's `section_reorder` command.
- Updated section tests, shared fixtures, `SKILL_CONTENT`, and the generated Todoist CLI skill file.

## Test plan

1. Create or pick a test project with at least three sections.
2. Run `node dist/index.js section reorder "Review" --project "<project>" --before "Done" --dry-run` and check the preview shows the expected order.
3. Run `node dist/index.js section reorder "Review" --project "<project>" --position 0 --json` and confirm the JSON output lists `Review` first.
4. List/open the project sections and confirm Todoist reflects the new order.

### DIFF
--- a/src/commands/section/index.ts
+++ b/src/commands/section/index.ts
@@ -1,11 +1,13 @@
 import { Command } from 'commander'
 import { CliError } from '../../lib/errors.js'
 import type { PaginatedViewOptions } from '../../lib/options.js'
+import { parseOrderArg } from '../../lib/order.js'
 import { archiveSection } from './archive.js'
 import { browseSection } from './browse.js'
 import { createSection } from './create.js'
 import { deleteSection } from './delete.js'
 import { listSections } from './list.js'
+import { reorderSection } from './reorder.js'
 import { unarchiveSection } from './unarchive.js'
 import { updateSection } from './update.js'
 
@@ -83,6 +85,40 @@ export function registerSectionCommand(program: Command): void {
                 return
             }
             return updateSection(id, options)
+        })
+
+    const reorderCmd = section
+        .command('reorder [ref]')
+        .description('Reorder a section within a project')
+        .option('--section <ref>', 'Section name or id:xxx')
+        .option('--project <ref>', 'Project name or id:xxx (required)')
+        .option('--before <ref>', 'Place before this sibling section')
+        .option('--after <ref>', 'Place after this sibling section')
+        .option('--position <n>', 'Move to a 0-indexed position within the project', parseOrderArg)
+        .option('--json', 'Output the new ordering as JSON')
+        .option('--dry-run', 'Preview what would happen without executing')
+        .addHelpText(
+            'after',
+            `
+Examples:
+  td section reorder "Review" --project "Roadmap" --before "Done"
+  td section reorder "Review" --project "Roadmap" --after "In Progress"
+  td section reorder --section "Review" --project "Roadmap" --position 0
+  td section reorder "Review" --project "Roadmap" --position 0`,
+        )
+        .action((ref, options) => {
+            if (ref && options.section) {
+                throw new CliError(
+                    'CONFLICTING_OPTIONS',
+                    'Cannot specify section both as argument and --section flag',
+                )
+            }
+            const sectionRef = ref || options.section
+            if (!sectionRef) {
+                reorderCmd.help()
+                return
+            }
+            return reorderSection(sectionRef, options)
         })
 
     const archiveCmd = section
--- a/src/commands/section/index.ts
+++ b/src/commands/section/index.ts
@@ -103,8 +103,8 @@ export function registerSectionCommand(program: Command): void {
 Examples:
   td section reorder "Review" --project "Roadmap" --before "Done"
   td section reorder "Review" --project "Roadmap" --after "In Progress"
-  td section reorder --section "Review" --project "Roadmap" --position 0
-  td section reorder "Review" --project "Roadmap" --position 0`,
+  td section reorder --section "Review" --project "Roadmap" --position 0 --dry-run
+  td section reorder "Review" --project "Roadmap" --position 2 --json`,
         )
         .action((ref, options) => {
             if (ref && options.section) {
--- a/src/commands/section/reorder.ts
+++ b/src/commands/section/reorder.ts
@@ -1,0 +1,172 @@
+import type { Section, TodoistApi } from '@doist/todoist-sdk'
+import { getApi } from '../../lib/api/core.js'
+import { reorderSections } from '../../lib/api/sections-sync.js'
+import { CliError } from '../../lib/errors.js'
+import { isQuiet } from '../../lib/global-args.js'
+import { formatJson } from '../../lib/output.js'
+import { paginate } from '../../lib/pagination.js'
+import { extractId, isIdRef, looksLikeRawId, resolveProjectId } from '../../lib/refs.js'
+
+export type ReorderSectionOptions = {
+    section?: string
+    project?: string
+    before?: string
+    after?: string
+    position?: number
+    json?: boolean
+    dryRun?: boolean
+}
+
+async function loadProjectSections(api: TodoistApi, projectId: string): Promise<Section[]> {
+    const { results } = await paginate(
+        (cursor, limit) => api.getSections({ projectId, cursor: cursor ?? undefined, limit }),
+        { limit: Number.MAX_SAFE_INTEGER },
+    )
+    return results
+}
+
+function resolveSectionFromList(sections: Section[], ref: string): Section {
+    if (!ref.trim()) {
+        throw new CliError('INVALID_SECTION', 'section reference cannot be empty.')
+    }
+
+    if (isIdRef(ref)) {
+        const id = extractId(ref)
+        const match = sections.find((section) => section.id === id)
+        if (!match) {
+            throw new CliError('SECTION_NOT_FOUND', `Section id:${id} not found in project.`)
+        }
+        return match
+    }
+
+    const lower = ref.toLowerCase()
+    const exact = sections.filter((section) => section.name.toLowerCase() === lower)
+    if (exact.length === 1) return exact[0]
+    if (exact.length > 1) {
+        throw new CliError(
+            'AMBIGUOUS_SECTION',
+            `Multiple sections match "${ref}" exactly in project:`,
+            exact.slice(0, 5).map((section) => `"${section.name}" (id:${section.id})`),
+        )
+    }
+
+    const partial = sections.filter((section) => section.name.toLowerCase().includes(lower))
+    if (partial.length === 1) return partial[0]
+    if (partial.length > 1) {
+        throw new CliError(
+            'AMBIGUOUS_SECTION',
+            `Multiple sections match "${ref}" in project:`,
+            partial.slice(0, 5).map((section) => `"${section.name}" (id:${section.id})`),
+        )
+    }
+
+    if (looksLikeRawId(ref)) {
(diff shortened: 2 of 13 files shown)


src/commands/section/reorder.ts
```diff
@@ -0,0 +1,172 @@
...
+
+async function loadProjectSections(api: TodoistApi, projectId: string): Promise<Section[]> {
+    const { results } = await paginate(
+        (cursor, limit) => api.getSections({ projectId, cursor: cursor ?? undefined, limit }),
+        { limit: Number.MAX_SAFE_INTEGER },
+    )
+    return results
+}
+
+function resolveSectionFromList(sections: Section[], ref: string): Section {
```
doistbot, line 37: [P2] This reinvents the in-memory entity resolution logic that already exists as `resolveFromList` in `src/lib/refs.ts`. Please export and reuse `resolveFromList` (updating it to handle exact-match ambiguity if that's an intentional improvement here) instead of duplicating the exact/partial matching and error-handling boilerplate.

src/commands/section/reorder.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+export async function reorderSection(ref: string, options: ReorderSectionOptions): Promise<void> {
+    if (!options.project) {
+        throw new CliError('MISSING_PROJECT', 'Specify --project <ref> to reorder a section.', [
+            'Section names are scoped to a project.',
+        ])
+    }
+
+    const flagCount = [options.before, options.after, options.position].filter(
```
doistbot: [P3] The argument validation for mutually-exclusive placement flags is duplicated verbatim from `src/commands/project/reorder.ts`. Consider extracting a shared helper to validate reorder flags and avoid structural drift across commands.

src/commands/section/index.ts
```diff
@@ -85,6 +87,40 @@ export function registerSectionCommand(program: Command): void {
...
+        .option('--json', 'Output the new ordering as JSON')
+        .option('--dry-run', 'Preview what would happen without executing')
+        .addHelpText(
+            'after',
+            `
+Examples:
+  td section reorder "Review" --project "Roadmap" --before "Done"
+  td section reorder "Review" --project "Roadmap" --after "In Progress"
+  td section reorder --section "Review" --project "Roadmap" --position 0
+  td section reorder "Review" --project "Roadmap" --position 0`,
```
doistbot: [P3] The last two examples are nearly identical. Consider updating these to match the `SKILL_CONTENT` examples which also demonstrate `--dry-run` and `--json` usage.

src/commands/section/reorder.ts
```diff
@@ -0,0 +1,172 @@
...
+    const [removed] = newOrder.splice(oldIndex, 1)
+    newOrder.splice(newIndex, 0, removed)
+
+    const items = newOrder.map((section, index) => ({
+        id: section.id,
+        sectionOrder: index + 1,
+    }))
+
+    if (options.dryRun) {
+        console.log(`Would reorder "${target.name}": position ${oldIndex} -> ${newIndex}`)
```
doistbot: [P3] For visual consistency with `project reorder`, consider using the unicode arrow `→` instead of `->` here (and similarly on line 162, as well as replacing `>` with `→` on line 147).

src/commands/section/reorder.ts
```diff
@@ -0,0 +1,172 @@
...
+    after?: string
+    position?: number
+    json?: boolean
+    dryRun?: boolean
+}
+
+async function loadProjectSections(api: TodoistApi, projectId: string): Promise<Section[]> {
+    const { results } = await paginate(
+        (cursor, limit) => api.getSections({ projectId, cursor: cursor ?? undefined, limit }),
+        { limit: Number.MAX_SAFE_INTEGER },
```
doistbot: [P2] Using `paginate(..., { limit: Number.MAX_SAFE_INTEGER })` makes this path allocate the full `all` array and then clone it again inside `paginate()` via `slice(0, limit)` before returning. Since reorder always needs the complete project order, that doubles the memory/copy work for every run. Please switch this to a dedicated "load all pages" loop (or a helper that returns the accumulated array directly) instead of the huge-limit sentinel here.

src/commands/section/section.test.ts
```diff
@@ -1,15 +1,22 @@
 import { Command } from 'commander'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 vi.mock('../../lib/api/core.js', () => ({
     getApi: vi.fn(),
 }))
 
+vi.mock('../../lib/api/sections-sync.js', () => ({
```
doistbot: [P2] The test relies on unnecessary module mocking of the internal `sections-sync.js` helper instead of mocking the SDK boundary. This creates a brittle coupling to the internal implementation rather than verifying the API payload. Please remove this mock, add `sync: vi.fn()` to `createMockApi()` in `src/test-support/mock-api.ts`, and assert against `mockApi.sync` using `expect.objectContaining` to match the SDK `createCommand` payload.

src/commands/section/section.test.ts
```diff
@@ -264,6 +271,314 @@ describe('section update --json', () => {
...
+                'Review',
+                '--project',
+                'id:proj-work',
+                '--before',
+                'Review',
+            ]),
+        ).rejects.toHaveProperty('code', 'INVALID_OPTIONS')
+    })
+
+    it('--dry-run does not call reorderSections', async () => {
```
doistbot: [P2] This `--dry-run` test only verifies that `reorderSections` is skipped and that some log line is printed. Since the point of `--dry-run` is the preview itself, add assertions for the `New section order` output (or the listed ids/names) so a regression in the preview ordering can't still pass by logging only `Would reorder ...`.

src/commands/section/section.test.ts
```diff
@@ -264,6 +271,314 @@ describe('section update --json', () => {
...
+            '--project',
+            'id:proj-work',
+            '--position',
+            '1',
+        ])
+
+        expect(mockReorderSections).not.toHaveBeenCalled()
+    })
+
+    it('--json outputs the new ordering', async () => {
```
doistbot, line 592: [P2] The suite exercises `--json` only for a real move, but `reorderSection` has a separate `newIndex === oldIndex && options.json` branch. Add a no-op JSON test here so `section reorder --json` is covered when nothing changes and can't regress to emitting the human `No change` message instead of the current ordering.

doistbot (commented):
This PR introduces the `td section reorder` command, offering robust functionality to rearrange sections within a project using positional or relative moves. It is a great addition that significantly expands the CLI's project management capabilities. A few adjustments are noted regarding the reuse of existing entity resolution and validation helpers, optimizing pagination logic to reduce memory overhead, polishing output formatting for consistency, and strengthening the test suite with more robust SDK boundary mocks and coverage for edge cases.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F349&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A74e933c0-5308-11f1-8a3d-c8b75ea4ceba%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A74e933c0-5308-11f1-8a3d-c8b75ea4ceba&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/section/reorder.ts
rmartins90: Extracted shared reorder placement validation and wired both project and section reorder through it.

src/commands/section/reorder.ts
rmartins90: Updated to reuse the shared resolver and keep exact-match ambiguity handling there.

src/commands/section/reorder.ts
rmartins90: Updated section reorder output to use the same `→` marker as project reorder.

src/commands/section/index.ts
rmartins90: Updated the examples to cover `--dry-run` and `--json`.

src/commands/section/reorder.ts
rmartins90: Switched this to an explicit all-pages loop.

src/commands/section/section.test.ts
rmartins90: Added no-op `--json` coverage.

src/commands/section/section.test.ts
rmartins90: Added explicit preview-order assertions.

src/commands/section/section.test.ts
rmartins90: Updated the tests to assert the SDK sync payload instead of mocking the internal helper.

src/commands/section/reorder.ts
```diff
@@ -0,0 +1,132 @@
...
+        sectionOrder: index + 1,
+    }))
+
+    if (options.dryRun) {
+        console.log(`Would reorder "${target.name}": position ${oldIndex} → ${newIndex}`)
+        console.log('New section order:')
+        for (let index = 0; index < newOrder.length; index++) {
+            const marker = newOrder[index].id === target.id ? '→' : ' '
+            console.log(`  ${marker} ${index}: ${newOrder[index].name} (id:${newOrder[index].id})`)
+        }
```
scottlovegrove: Why do we do this for the dry run, but not for the actual run? Shouldn't this match what we display on the actual run?

src/commands/section/section.test.ts
```diff
@@ -264,6 +280,391 @@ describe('section update --json', () => {
...
+        await expect(
+            program.parseAsync([
+                'node',
+                'td',
+                'section',
+                'reorder',
+                'Review',
+                '--project',
+                'id:proj-work',
+            ]),
```
scottlovegrove: We could create a const here for the array as these are used in a lot of the tests and could act as the base args.

scottlovegrove (approved):

src/commands/section/section.test.ts
rmartins90: Done, added shared base args for the reorder tests.

src/commands/section/reorder.ts
rmartins90: Yeah, that makes sense. Updated the real run to print the new section order through the same helper as dry-run.

PR #351, opened 2026-05-19
feat(reminder): support urgent reminders (iOS 26)

### LINKED ISSUE #350: `reminder add` / `reminder update`: support `--urgent` flag for urgent (iOS 26) reminders
### Summary

`td reminder add` and `td reminder update` cannot set the `is_urgent`
flag on a reminder, and `td reminder get` / `td reminder list` drop
`is_urgent` from their output. Urgent reminders are the iOS 26 feature
where time-based reminders fire a full-screen alarm that breaks through
Silent and Do Not Disturb modes ([help center][help]).

The SDK already types the field (`@doist/todoist-sdk` 10.2.0, bumped in
v1.65.0), so this is purely a CLI surface gap.

[help]: https://www.todoist.com/help/articles/add-an-urgent-reminder-with-the-todoist-ios-app-WeBYdY5ra

### Requested changes

**Write path**
- `td reminder add`: add `--urgent` (and `--no-urgent` for symmetry)
  that threads through to `is_urgent` on the `reminder_add` sync
  command.
- `td reminder update`: same flag.

**Read path**
- `td reminder get` and `td reminder list`: include `is_urgent` in
  plain output and as a key in `--json` output.

### Current behaviour (tip of `main`, v1.65.0)

`src/lib/api/reminders.ts` defines local types that omit the field:

```ts
export interface Reminder {
    id: string
    itemId: string
    type: 'absolute' | 'relative' | 'location'
    due?: ReminderDue
    minuteOffset?: number
    isDeleted: boolean
}

export interface AddReminderArgs {
    itemId: string
    minuteOffset?: number
    due?: ReminderDue
}

export interface UpdateReminderArgs {
    minuteOffset?: number
    due?: ReminderDue
}
```

`toReminder()` hand-picks fields from the SDK's `Reminder`, so
`isUrgent` is dropped on read. `addReminder()` / `updateReminder()`
build the sync commands without it, so the CLI never sends it on
writes. Help output for both commands lacks any urgent flag:

```
$ td reminder add --help
Options:
  --task <ref>         Task reference (name or id:xxx)
  --before <duration>  Time before due (e.g., 30m, 1h)
  --at <datetime>      Specific time (e.g., 2024-01-15 10:00)
  --json               Output the created reminder as JSON
  --dry-run            Preview what would happen without executing

$ td reminder update --help
Options:
  --before <duration>  Time before due (e.g., 30m, 1h)
  --at <datetime>      Specific time (e.g., 2024-01-15 10:00)
  --dry-run            Preview what would happen without executing
```

### SDK + server side already accept it

The SDK types `isUrgent: z.ZodOptional<z.ZodBoolean>` on reminder
request and response schemas across:

- `dist/types/types/reminders/requests.d.ts` (multiple request shapes)
- `dist/types/types/sync/resources/reminders.d.ts` (resource shape)
- `dist/types/types/sync/response.d.ts` (sync response)
- `dist/types/types/webhooks/reminders.d.ts` (webhook events — here
  it is `isUrgent: z.ZodBoolean`, non-optional)

I verified the sync API accepts it by direct POST to `/api/v1/sync`:

```json
[{"type":"reminder_add","temp_id":"…","uuid":"…","args":{
  "item_id":"<task id>",
  "type":"absolute",
  "due":{"date":"2026-05-19T10:52:35Z","timezone":null,
         "is_recurring":false,
         "string":"2026-05-19T10:52:35.000000Z","lang":"en"},
  "is_urgent":true
}}]
```

Response: HTTP 200, `sync_status: {"<uuid>": "ok"}`. A subsequent
sync read returns the reminder with `"is_urgent": true` preserved.

### Suggested patch shape

Roughly a dozen lines in `src/lib/api/reminders.ts`:

1. Add `isUrgent?: boolean` to `Reminder`, `AddReminderArgs`,
   `UpdateReminderArgs`.
2. Map `isUrgent: r.isUrgent` in `toReminder()`.
3. Pass `isUrgent` through `pickDefined({...})` in both
   `addReminder()` and `updateReminder()`.

Plus `--urgent` / `--no-urgent` flag wiring in
`src/commands/reminder/add.ts` and `update.ts`, and an `[urgent]`
marker (or similar) in the plain-text formatters.

### Notes / out of scope

- This issue is scoped to the CLI surface gap. Whether iOS currently
  renders or fires urgent alarms in response to server-set
  `is_urgent: true` is a separate concern at the iOS / sync layer.
- Per Todoist docs, urgent applies only to time-based reminders;
  location reminders are excluded.

### Environment

- todoist-cli: 1.62.2 (gap confirmed on `main` @ v1.65.0)
- @doist/todoist-sdk: 10.2.0
- Plan: Pro (legacy personal)

scottlovegrove:
Closes #350.

## Summary

- `reminder add` / `reminder update` now accept `--urgent` / `--no-urgent`, threaded through `is_urgent` on the `reminder_add` / `reminder_update` sync commands.
- `reminder get` / `reminder list` plain output show a red `[urgent]` badge after `[time]`; `--json` includes `isUrgent`.
- `reminder update` no longer requires `--before` / `--at` when only toggling urgency — `td reminder update <id> --no-urgent` is valid on its own.

The SDK (`@doist/todoist-sdk` 10.2.0, already on `main`) types `isUrgent` on absolute + relative reminder schemas. Location reminders are out of scope — the SDK schema confirms the field doesn't exist there.

## Test plan

- [x] `npm run type-check`
- [x] `npm test` (1607 passing, +3 new)
- [x] `npm run check`
- [x] `npm run sync:skill`
- [ ] Live smoke: `td reminder add <task> --at "2026-06-01 09:00" --urgent --json` → JSON shows `isUrgent: true`
- [ ] Live smoke: `td reminder list --task <task>` → row shows `[urgent]` badge
- [ ] Live smoke: `td reminder update <id> --no-urgent` (no time flags) → succeeds; subsequent `get` no longer shows badge

### DIFF
--- a/src/commands/reminder/add.ts
+++ b/src/commands/reminder/add.ts
@@ -10,7 +10,7 @@ import { CliError } from '../../lib/errors.js'
 import { isQuiet } from '../../lib/global-args.js'
 import { formatJson, printDryRun } from '../../lib/output.js'
 import { resolveTaskRef } from '../../lib/refs.js'
-import { parseDateTime } from './helpers.js'
+import { formatUrgentBadge, parseDateTime } from './helpers.js'
 
 interface AddOptions {
     before?: string
@@ -101,11 +101,11 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
         return
     }
 
-    const urgentSuffix = options.urgent ? ` ${chalk.red('[urgent]')}` : ''
+    const urgent = formatUrgentBadge(options.urgent)
     if (minuteOffset !== undefined) {
-        console.log(`Added reminder: ${formatDuration(minuteOffset)} before due${urgentSuffix}`)
+        console.log(`Added reminder: ${formatDuration(minuteOffset)} before due${urgent}`)
     } else if (due) {
-        console.log(`Added reminder: at ${due.date.replace('T', ' ')}${urgentSuffix}`)
+        console.log(`Added reminder: at ${due.date.replace('T', ' ')}${urgent}`)
     }
     console.log(chalk.dim(`ID: ${reminderId}`))
 }
--- a/src/commands/reminder/add.ts
+++ b/src/commands/reminder/add.ts
@@ -15,6 +15,7 @@ import { parseDateTime } from './helpers.js'
 interface AddOptions {
     before?: string
     at?: string
+    urgent?: boolean
     json?: boolean
     dryRun?: boolean
 }
@@ -69,6 +70,7 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
             Task: task.content,
             Before: options.before,
             At: options.at,
+            Urgent: options.urgent === undefined ? undefined : String(options.urgent),
         })
         return
     }
@@ -77,6 +79,7 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
         itemId: task.id,
         minuteOffset,
         due,
+        isUrgent: options.urgent,
     })
 
     if (options.json) {
@@ -86,6 +89,7 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
             type: minuteOffset !== undefined ? 'relative' : 'absolute',
             minuteOffset,
             due,
+            isUrgent: options.urgent,
             isDeleted: false,
         }
         console.log(formatJson(reminder, 'reminder'))
@@ -97,10 +101,11 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
         return
     }
 
+    const urgentSuffix = options.urgent ? ` ${chalk.red('[urgent]')}` : ''
     if (minuteOffset !== undefined) {
-        console.log(`Added reminder: ${formatDuration(minuteOffset)} before due`)
+        console.log(`Added reminder: ${formatDuration(minuteOffset)} before due${urgentSuffix}`)
     } else if (due) {
-        console.log(`Added reminder: at ${due.date.replace('T', ' ')}`)
+        console.log(`Added reminder: at ${due.date.replace('T', ' ')}${urgentSuffix}`)
     }
     console.log(chalk.dim(`ID: ${reminderId}`))
 }
--- a/src/commands/reminder/get.ts
+++ b/src/commands/reminder/get.ts
@@ -1,17 +1,21 @@
+import type { Reminder as SdkReminder } from '@doist/todoist-sdk'
 import chalk from 'chalk'
 import { getReminderById } from '../../lib/api/reminders.js'
 import { formatJson } from '../../lib/output.js'
 import { lenientIdRef } from '../../lib/refs.js'
-import { formatReminderTime } from './helpers.js'
+import { formatReminderTime, formatUrgentBadge } from './helpers.js'
 
 interface GetOptions {
     json?: boolean
     full?: boolean
 }
 
+// `td reminder get` targets the time-reminder endpoint, so location is impossible here.
+type TimeReminder = Extract<SdkReminder, { type: 'absolute' | 'relative' }>
+
 export async function getReminderCmd(reminderId: string, options: GetOptions): Promise<void> {
     const id = lenientIdRef(reminderId, 'reminder')
-    const reminder = await getReminderById(id)
+    const reminder = (await getReminderById(id)) as TimeReminder
 
     if (options.json) {
         console.log(formatJson(reminder, 'reminder', options.full))
@@ -20,13 +24,9 @@ export async function getReminderCmd(reminderId: string, options: GetOptions): P
 
     const idStr = chalk.dim(reminder.id)
     const type = chalk.cyan('[time]')
-    const time = formatReminderTime(
-        reminder as {
-            type: 'relative' | 'absolute'
-            minuteOffset?: number
-            due?: { date: string }
-        },
-    )
-    console.log(`${idStr}  ${type} ${time}`)
+    const time = formatReminderTime(reminder)
+    const urgent = formatUrgentBadge(reminder.isUrgent)
+    const urgentSegment = urgent ? ` ${urgent}` : ''
+    console.log(`${idStr}  ${type}${urgentSegment} ${time}`)
     console.log(chalk.dim(`Task: ${reminder.itemId}`))
 }
--- a/src/commands/reminder/get.ts
+++ b/src/commands/reminder/get.ts
@@ -1,17 +1,13 @@
-import type { Reminder as SdkReminder } from '@doist/todoist-sdk'
 import chalk from 'chalk'
 import { getReminderById } from '../../lib/api/reminders.js'
 import { formatJson } from '../../lib/output.js'
 import { lenientIdRef } from '../../lib/refs.js'
-import { formatReminderTime, formatUrgentBadge } from './helpers.js'
+import { type TimeReminder, formatReminderTime, formatUrgentBadge } from './helpers.js'
 
 interface GetOptions {
     json?: boolean
     full?: boolean
 }
-
-// `td reminder get` targets the time-reminder endpoint, so location is impossible here.
-type TimeReminder = Extract<SdkReminder, { type: 'absolute' | 'relative' }>
 
 export async function getReminderCmd(reminderId: string, options: GetOptions): Promise<void> {
     const id = lenientIdRef(reminderId, 'reminder')
@@ -26,7 +22,6 @@ export async function getReminderCmd(reminderId: string, options: GetOptions): P
     const type = chalk.cyan('[time]')
     const time = formatReminderTime(reminder)
     const urgent = formatUrgentBadge(reminder.isUrgent)
-    const urgentSegment = urgent ? ` ${urgent}` : ''
-    console.log(`${idStr}  ${type}${urgentSegment} ${time}`)
+    console.log(`${idStr}  ${type}${urgent} ${time}`)
     console.log(chalk.dim(`Task: ${reminder.itemId}`))
 }
--- a/src/commands/reminder/helpers.ts
+++ b/src/commands/reminder/helpers.ts
@@ -4,6 +4,7 @@ import {
     type LocationTrigger,
     type Reminder,
 } from '@doist/todoist-sdk'
+import chalk from 'chalk'
 import type { ReminderDue } from '../../lib/api/reminders.js'
 import { formatDuration } from '../../lib/duration.js'
 import { CliError } from '../../lib/errors.js'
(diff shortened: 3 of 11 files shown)


src/commands/reminder/update.ts
```diff
@@ -49,13 +51,16 @@ export async function updateReminderCmd(reminderId: string, options: UpdateOptio
...
 
-    await apiUpdateReminder(id, { minuteOffset, due })
+    await apiUpdateReminder(id, { minuteOffset, due, isUrgent: options.urgent })
 
     if (!isQuiet()) {
         if (minuteOffset !== undefined) {
             console.log(`Updated reminder: ${formatDuration(minuteOffset)} before due (id:${id})`)
         } else if (due) {
             console.log(`Updated reminder: at ${due.date.replace('T', ' ')} (id:${id})`)
+        } else if (options.urgent !== undefined) {
```
doistbot, line 75: [P3] When a user updates both the time and urgency simultaneously (e.g., `--before 10m --urgent` or `--at "..." --no-urgent`), the `else if` structure causes the confirmation message to silently omit the urgency update detail. Consider appending the urgency state to the time confirmation, similar to the suffix approach used in `add.ts`.

src/commands/reminder/add.ts
```diff
@@ -86,6 +89,7 @@ export async function addReminder(taskRef: string, options: AddOptions): Promise
             type: minuteOffset !== undefined ? 'relative' : 'absolute',
             minuteOffset,
             due,
+            isUrgent: options.urgent,
```
doistbot, line 92: [P1] Adding `isUrgent` here is not enough to expose it in normal `--json` output. `formatJson(..., 'reminder')` still filters reminders through the existing essential-field whitelist in `src/lib/output.ts`, which does not include `isUrgent`, so `reminder add/get/list --json` will silently omit the new field unless that whitelist is updated too.

src/commands/reminder/update.ts
```diff
@@ -9,12 +9,13 @@ import { parseDateTime } from './helpers.js'
...
     before?: string
     at?: string
+    urgent?: boolean
     dryRun?: boolean
 }
 
 export async function updateReminderCmd(reminderId: string, options: UpdateOptions): Promise<void> {
-    if (!options.before && !options.at) {
-        throw new CliError('MISSING_TIME', 'Must specify --before or --at')
+    if (!options.before && !options.at && options.urgent === undefined) {
```
doistbot, line 22: [P1] This new urgency-only path mis-serializes relative reminders. When the user runs `td reminder update <id> --urgent/--no-urgent` without `--before` or `--at`, `apiUpdateReminder()` is called with both time fields undefined, and `src/lib/api/reminders.ts` currently derives `type` as `'absolute'` in that case. That means an existing relative reminder is sent as an absolute update with no time. Please preserve the current reminder type (or omit `type` for urgency-only patches) before allowing this path.

src/commands/reminder/helpers.ts
```diff
@@ -14,6 +15,11 @@ interface ReminderLike {
     type: Reminder['type']
     minuteOffset?: number
     due?: { date: string }
+    isUrgent?: boolean
+}
+
+export function formatUrgentBadge(isUrgent: boolean | undefined): string {
+    return isUrgent ? chalk.red('[urgent]') : ''
```
doistbot: [P2] Consider including the leading space in this return value (e.g. `` ` ${chalk.red('[urgent]')}` ``). This allows you to directly reuse `formatUrgentBadge` in `add.ts` (where `chalk.red('[urgent]')` is currently reimplemented) and eliminates the duplicated conditional logic for `urgentSegment` in `get.ts` and `list.ts`.

src/commands/reminder/get.ts
```diff
@@ -1,17 +1,21 @@
...
-import { formatReminderTime } from './helpers.js'
+import { formatReminderTime, formatUrgentBadge } from './helpers.js'
 
 interface GetOptions {
     json?: boolean
     full?: boolean
 }
 
+// `td reminder get` targets the time-reminder endpoint, so location is impossible here.
+type TimeReminder = Extract<SdkReminder, { type: 'absolute' | 'relative' }>
```
doistbot: [P2] The new "time reminder" typing is now duplicated in two different forms: this file defines/casts a local `TimeReminder`, while `list.ts` uses an inline `'isUrgent' in reminder` cast. Please extract a shared time-reminder type or type guard in `helpers.ts` or `lib/api/reminders.ts` and reuse it in both places so future reminder-field changes only need one update.

src/commands/reminder/reminder.test.ts
```diff
@@ -298,6 +298,32 @@ describe('reminder list', () => {
         ).rejects.toThrow('--cursor requires --type')
     })
 
+    it('shows [urgent] badge for urgent reminders', async () => {
```
doistbot, line 301: [P2] This adds badge coverage for `reminder list`, but `reminder get` also changed its plain-text rendering in this PR and still has no urgent case. Please extend the existing `reminder get` test (or add one nearby) with `isUrgent: true` so regressions in that output path are caught.

src/commands/reminder/reminder.test.ts
```diff
@@ -616,6 +642,31 @@ describe('reminder add', () => {
             ]),
         ).rejects.toThrow('Cannot specify task both as argument and --task flag')
     })
+
+    it('passes --urgent through and shows [urgent] in output', async () => {
```
doistbot, line 646: [P2] The new urgency coverage only exercises human-readable output. This PR also changes the default `--json` contract (`isUrgent` on reminder add/get/list), and none of the nearby JSON tests assert that field. Please extend one of them to verify `isUrgent` is present without `--full`.

src/commands/reminder/index.ts
```diff
@@ -90,6 +93,8 @@ export function registerReminderCommand(program: Command): void {
         .description('Update a reminder')
         .option('--before <duration>', 'Time before due (e.g., 30m, 1h)')
         .option('--at <datetime>', 'Specific time (e.g., 2024-01-15 10:00)')
+        .option('--urgent', 'Mark reminder as urgent (iOS full-screen alarm)')
```
doistbot, line 96: [P2] This expands `reminder update`, but the command still has no `--json` mode. Nearby update commands (including `reminder location update`) return the updated entity via `formatJson(...)`, and the repo guidance expects mutating updates to be machine-readable as well. Please add `--json` here and emit the refreshed reminder so the new urgent toggle is automatable.

doistbot (commented):
This PR provides a solid foundation for urgent reminders by introducing intuitive CLI flags and clear visual badging. The implementation is well-aligned with the new SDK, though a few edge cases need attention to ensure full correctness. Specifically, there are serialization issues that can corrupt relative reminders during urgency-only updates, a field whitelist preventing the new urgency property from appearing in JSON output, and a few opportunities to improve type reuse, consolidate badge formatting, support `--json` on the update command, and expand test coverage.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F351&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A07e3a790-5382-11f1-88f6-fef8a33cf732%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A07e3a790-5382-11f1-88f6-fef8a33cf732&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/reminder/update.ts
scottlovegrove: Fixed in de44a5a. `update.ts` now computes `urgent = formatUrgentBadge(options.urgent)` once and appends it to the time-based confirmation lines, so `--before 10m --urgent` shows the badge; urgency-only updates still get the explicit "marked urgent / marked not urgent" message.

src/commands/reminder/add.ts
scottlovegrove: Fixed in de44a5a. `REMINDER_ESSENTIAL_FIELDS` in src/lib/output.ts now includes `isUrgent`, and there's a new test asserting it survives the whitelist in default `--json` output.

src/commands/reminder/update.ts
scottlovegrove: Fixed in de44a5a. The SDK's sync `type` field is a discriminated literal so we can't omit it. `updateReminder()` now fetches the existing reminder via `api.getReminder(id)` when neither `--before` nor `--at` is supplied and reuses its `type`, so a relative reminder stays relative on an urgency-only patch. Errors out cleanly if someone passes a location id to `reminder update`.

src/commands/reminder/helpers.ts
scottlovegrove: Fixed in de44a5a. `formatUrgentBadge` now returns ` ${chalk.red('[urgent]')}` (leading space, empty otherwise). `add.ts`, `update.ts`, `get.ts`, and `list.ts` all just concatenate the return value — no more duplicated conditional / urgentSegment local.

src/commands/reminder/get.ts
scottlovegrove: Fixed in de44a5a. Exported `TimeReminder = Extract<Reminder, { type: 'absolute' | 'relative' }>` from `helpers.ts`. `get.ts` uses it for the post-fetch cast and `list.ts` casts the iterated array, dropping the inline `'isUrgent' in reminder` check.

src/commands/reminder/reminder.test.ts
scottlovegrove: Done in de44a5a — added a `reminder get` test that mocks an `isUrgent: true` reminder and asserts `[urgent]` appears in the plain-text output.

src/commands/reminder/reminder.test.ts
scottlovegrove: Done in de44a5a — added two assertions: a `reminder get --json` test that the output contains `"isUrgent": true`, and a `reminder update --json` test that parses the JSON and asserts `parsed.isUrgent === true`. Both run without `--full` so they exercise the whitelist.

src/commands/reminder/index.ts
scottlovegrove: Added in de44a5a. `reminder update` now has `--json`; after the sync patch we re-fetch the reminder via `getReminderById` and emit it through `formatJson(reminder, 'reminder')`, matching the `reminder location update` pattern.

pedroalves0 (commented):
✅

PR #356, opened 2026-05-22
feat(billing): add billing commands and billing OAuth scope

scottlovegrove:
## Summary

Surfaces the SDK's read-only billing endpoints (shipped in `@doist/todoist-sdk@10.3.0`, [SDK PR #608](https://github.com/Doist/todoist-sdk-typescript/pull/608)) as a new `td billing` command group.

- `td billing subscription` (default) — plan, status, activation, expiry, price, credit balance, billing-portal URLs
- `td billing plan` — Pro plan status, downgrade date, price list
- `td billing prices` — available Pro and Teams prices by billing cycle
- `td billing pricing [--formatted]` — current/legacy pricing keyed by version
- All subcommands support `--json` / `--ndjson` (raw SDK payload).

## OAuth scope

Adds a single opt-in `billing` flag: `td auth login --additional-scopes=billing`.

- Default login grants `billing:read_write`; adding `--read-only` narrows it to `billing:read`.
- Either variant satisfies the read commands (the gate is a `billing:read` substring check, and `billing:read_write` contains it). The read_write default is forward-looking should write commands land later.
- Implemented via a new optional `readOnlyScope` on `ScopeDefinition`; `resolveAuthScope`/`oauthScopeFor` now honor read-only mode (the single chokepoint used by both the login request and the persisted scope).

## Wiring

- `getSubscriptionInfo` is a POST but read-only, so it's added to the read-only safe-method allowlist (`permissions.ts`).
- The four read methods are mapped in `METHOD_REQUIRED_FLAG → billing`, so a 403 surfaces a `MISSING_SCOPE` error with a `--additional-scopes=billing` re-login hint (names `billing:read` for read-only logins).
- `billing` appended to `AUTH_FLAG_ORDER`; command registered in the lazy registry.
- Docs/skill updated (SKILL.md synced), CODEBASE.md tree, README opt-in-scopes note.

## Test plan

- [x] `npm run check` (oxlint + oxfmt) clean
- [x] `npm test` — full suite green (1629 tests), incl. new `billing.test.ts` and `resolveAuthScope` coverage
- [x] `npm run check:skill-sync` passes
- [x] `td billing --help` and `td auth login --help` show the new command/scope
- [ ] Manual end-to-end against a real account: `td auth login --additional-scopes=billing`, then each `td billing` subcommand (+ `--json`); verify read-only login yields `billing:read`; verify a token without the scope fails with `MISSING_SCOPE`

### DIFF
--- a/src/commands/billing/billing.test.ts
+++ b/src/commands/billing/billing.test.ts
@@ -1,4 +1,3 @@
-import { TodoistRequestError } from '@doist/todoist-sdk'
 import { Command } from 'commander'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
@@ -10,85 +9,18 @@ vi.mock('../../lib/api/core.js', async (importOriginal) => {
     }
 })
 
-vi.mock('../../lib/auth.js', async (importOriginal) => {
-    const actual = await importOriginal<typeof import('../../lib/auth.js')>()
-    return {
-        ...actual,
-        getAuthMetadata: vi.fn(),
-    }
-})
-
-import { getApi, wrapApiError } from '../../lib/api/core.js'
-import { getAuthMetadata } from '../../lib/auth.js'
-import { CliError } from '../../lib/errors.js'
+import { getApi } from '../../lib/api/core.js'
+import { fixtures } from '../../test-support/fixtures.js'
 import { createMockApi, type MockApi } from '../../test-support/mock-api.js'
 import { registerBillingCommand } from './index.js'
 
 const mockGetApi = vi.mocked(getApi)
-const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
 
 function createProgram() {
     const program = new Command()
     program.exitOverride()
     registerBillingCommand(program)
     return program
-}
-
-const SUBSCRIPTION = {
-    status: 'autorenew',
-    plan: 'pro',
-    expirationDate: '2026-12-31',
-    activationMethod: 'stripe',
-    planPrice: {
-        amount: '600',
-        rawAmount: 600,
-        currency: 'USD',
-        billingCycle: 'monthly',
-        taxBehavior: 'exclusive',
-    },
-    billingPortalUrl: 'https://billing.stripe.com/portal/abc',
-    billingPortalSwitchToAnnualUrl: null,
-    hasBillingPortal: true,
-    hasBillingPortalSwitchToAnnual: false,
-    invoiceCreditBalance: { usd: 0 },
-    hasSwitchLegacyToCurrent: false,
-}
-
-const PRO_PLAN_DETAILS = {
-    currentPlanStatus: 'Active',
-    downgradeAt: null,
-    priceList: [
-        {
-            billingCycle: 'monthly',
-            prices: [{ currency: 'USD', unitAmount: 600, taxBehavior: 'exclusive' }],
-        },
-    ],
-}
-
-const PRICES = {
-    pro: [
-        {
-            billingCycle: 'yearly',
-            prices: [{ currency: 'USD', unitAmount: 6000, taxBehavior: 'exclusive' }],
-        },
-    ],
-    teams: [
-        {
-            billingCycle: 'monthly',
-            prices: [{ currency: 'USD', unitAmount: 800, taxBehavior: 'inclusive' }],
-        },
-    ],
-}
-
-const PRICING = {
-    latestPro: 'v25',
-    latestBiz: 'v25',
-    sessionPro: 'v25',
-    sessionBiz: 'v25',
-    v25: {
-        pro: { usd: { monthly: 400, yearly: 2900 } },
-        biz: { usd: { monthly: 800, yearly: 7200 } },
-    },
 }
 
 describe('billing subscription', () => {
@@ -103,7 +35,7 @@ describe('billing subscription', () => {
     it('is the default subcommand and renders the subscription summary', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getSubscriptionInfo.mockResolvedValue(SUBSCRIPTION)
+        mockApi.getSubscriptionInfo.mockResolvedValue(fixtures.billing.subscription)
 
         // No subcommand given → default subscription
         await program.parseAsync(['node', 'td', 'billing'])
@@ -114,7 +46,7 @@ describe('billing subscription', () => {
         expect(output).toContain('Status:             autorenew')
         expect(output).toContain('Activation:         stripe')
         expect(output).toContain('Expires:            2026-12-31')
-        expect(output).toContain('$6.00/mo')
+        expect(output).toContain('$6/mo')
         expect(output).toContain('https://billing.stripe.com/portal/abc')
         consoleSpy.mockRestore()
     })
@@ -122,25 +54,25 @@ describe('billing subscription', () => {
     it('outputs the raw payload with --json', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getSubscriptionInfo.mockResolvedValue(SUBSCRIPTION)
+        mockApi.getSubscriptionInfo.mockResolvedValue(fixtures.billing.subscription)
 
         await program.parseAsync(['node', 'td', 'billing', 'subscription', '--json'])
 
         const output = consoleSpy.mock.calls[0][0] as string
-        expect(JSON.parse(output)).toEqual(SUBSCRIPTION)
+        expect(JSON.parse(output)).toEqual(fixtures.billing.subscription)
         consoleSpy.mockRestore()
     })
 
     it('outputs single-line JSON with --ndjson', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getSubscriptionInfo.mockResolvedValue(SUBSCRIPTION)
+        mockApi.getSubscriptionInfo.mockResolvedValue(fixtures.billing.subscription)
 
         await program.parseAsync(['node', 'td', 'billing', 'subscription', '--ndjson'])
 
         const output = consoleSpy.mock.calls[0][0] as string
         expect(output).not.toContain('\n')
-        expect(JSON.parse(output)).toEqual(SUBSCRIPTION)
+        expect(JSON.parse(output)).toEqual(fixtures.billing.subscription)
         consoleSpy.mockRestore()
     })
 
@@ -148,7 +80,7 @@ describe('billing subscription', () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
         mockApi.getSubscriptionInfo.mockResolvedValue({
-            ...SUBSCRIPTION,
+            ...fixtures.billing.subscription,
             status: 'none',
             plan: 'free',
             expirationDate: null,
@@ -179,25 +111,27 @@ describe('billing plan', () => {
     it('renders pro plan details', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getProPlanDetails.mockResolvedValue(PRO_PLAN_DETAILS)
+        mockApi.getProPlanDetails.mockResolvedValue(fixtures.billing.proPlanDetails)
 
         await program.parseAsync(['node', 'td', 'billing', 'plan'])
 
         const output = consoleSpy.mock.calls.map((c) => c[0]).join('\n')
         expect(output).toContain('Status:             Active')
         expect(output).toContain('Downgrade at:       (none)')
-        expect(output).toContain('monthly: $6.00')
+        expect(output).toContain('monthly: $6')
         consoleSpy.mockRestore()
     })
 
     it('outputs the raw payload with --json', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getProPlanDetails.mockResolvedValue(PRO_PLAN_DETAILS)
+        mockApi.getProPlanDetails.mockResolvedValue(fixtures.billing.proPlanDetails)
 
         await program.parseAsync(['node', 'td', 'billing', 'plan', '--json'])
 
-        expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual(PRO_PLAN_DETAILS)
+        expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual(
+            fixtures.billing.proPlanDetails,
+        )
         consoleSpy.mockRestore()
     })
 })
@@ -214,15 +148,15 @@ describe('billing prices', () => {
     it('renders pro and teams price listings', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getPrices.mockResolvedValue(PRICES)
+        mockApi.getPrices.mockResolvedValue(fixtures.billing.prices)
 
         await program.parseAsync(['node', 'td', 'billing', 'prices'])
 
         const output = consoleSpy.mock.calls.map((c) => c[0]).join('\n')
         expect(output).toContain('Pro')
-        expect(output).toContain('yearly: $60.00')
+        expect(output).toContain('yearly: $60')
         expect(output).toContain('Teams')
-        expect(output).toContain('monthly: $8.00')
+        expect(output).toContain('monthly: $8')
         consoleSpy.mockRestore()
     })
 })
@@ -239,7 +173,7 @@ describe('billing pricing', () => {
     it('formats minor-unit numbers as money and lists version entries', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getPricing.mockResolvedValue(PRICING)
+        mockApi.getPricing.mockResolvedValue(fixtures.billing.pricing)
 
         await program.parseAsync(['node', 'td', 'billing', 'pricing'])
 
@@ -247,20 +181,14 @@ describe('billing pricing', () => {
         const output = consoleSpy.mock.calls.map((c) => c[0]).join('\n')
         expect(output).toContain('Latest Pro:         v25')
         expect(output).toContain('v25')
-        expect(output).toContain('pro (USD): $4.00/mo, $29.00/yr')
+        expect(output).toContain('pro (USD): $4/mo, $29/yr')
         consoleSpy.mockRestore()
     })
 
     it('passes --formatted through and prints pre-formatted strings as-is', async () => {
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-        mockApi.getPricing.mockResolvedValue({
-            latestPro: 'v25',
-            latestBiz: 'v25',
-            sessionPro: 'v25',
-            sessionBiz: 'v25',
-            v25: { pro: { usd: { monthly: '$4', yearly: '$29' } } },
-        })
+        mockApi.getPricing.mockResolvedValue(fixtures.billing.pricingFormatted)
 
         await program.parseAsync(['node', 'td', 'billing', 'pricing', '--formatted'])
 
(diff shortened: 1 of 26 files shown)


src/commands/billing/format.ts
```diff
@@ -0,0 +1,37 @@
...
+    console.log(JSON.stringify(payload, null, indent))
+    return true
+}
+
+/** Format an amount given in a currency's minor units (e.g. cents) as money. */
+export function formatMoney(minorUnits: number, currency: string): string {
+    const code = currency.toUpperCase()
+    try {
+        return new Intl.NumberFormat('en-US', { style: 'currency', currency: code }).format(
+            minorUnits / 100,
```
doistbot: [P1] This assumes every currency has 2 fractional digits. `unitAmount`/`rawAmount` are minor units, so currencies like JPY/KRW will render 100× too small here, and 3-decimal currencies will be wrong too. Please derive the divisor from the currency’s actual fraction digits (or use currency metadata) instead of hard-coding `/ 100`.

src/lib/api/core.ts
```diff
@@ -224,7 +228,8 @@ async function getScopeRemediation(methodName: string | undefined): Promise<stri
     }
     const metadata = await getAuthMetadata()
     const command = buildReloginCommand(metadata, requiredFlag)
-    return `This command requires the ${oauthScopeFor(requiredFlag)} scope. Re-run \`${command}\` to grant it.`
+    const scope = oauthScopeFor(requiredFlag, metadata.authMode === 'read-only')
```
doistbot, line 237: [P2] This now personalizes the required scope from `authMode`, but the re-login command still preserves `--read-only` only from `authFlags`. Migrated v1 users can have `auth_mode: 'read-only'` with `auth_flags` unset, so this hint will say `billing:read` while suggesting `td auth login --additional-scopes=billing`, which widens them to read-write. Make the command builder fall back to `authMode`, or keep the scope/message aligned.

src/commands/billing/format.ts
```diff
@@ -0,0 +1,37 @@
...
+    ndjson?: boolean
+}
+
+/**
+ * Emit the raw payload as JSON (`--json`, indented) or NDJSON (`--ndjson`,
+ * single line) and report whether machine output was produced, so callers can
+ * early-return before rendering the human view. Billing responses are not
+ * `output.ts` entity types, so they're serialized verbatim.
+ */
+export function outputMachine(payload: unknown, options: BillingViewOptions): boolean {
```
doistbot: [P3] This helper re-implements JSON/NDJSON stringification. Consider reusing `formatJson` and `formatNdjson` from `../../lib/output.js` to ensure machine output stays consistent with the rest of the CLI.

```typescript
import { formatJson, formatNdjson } from '../../lib/output.js'

export function outputMachine(payload: unknown, options: BillingViewOptions): boolean {
    if (options.json) {
        console.log(formatJson(payload))
        return true
    }
    if (options.ndjson) {
        console.log(formatNdjson([payload]))
        return true
    }
    return false
}
```

src/commands/billing/format.ts
```diff
@@ -0,0 +1,37 @@
+import type { PriceListing } from '@doist/todoist-sdk'
+
+export interface BillingViewOptions {
```
doistbot: [P3] `BillingViewOptions` is a local copy of the shared machine-output shape the repo already uses via `src/lib/options.ts` / cli-core. Reusing `ViewOptions` (or `Pick<ViewOptions, 'json' | 'ndjson'>`) here would avoid another options contract that can drift if those shared flags change.

src/commands/billing/pricing.ts
```diff
@@ -0,0 +1,43 @@
...
+    formatted?: boolean
+}
+
+export async function viewPricing(options: PricingOptions = {}): Promise<void> {
+    const api = await getApi()
+    const pricing = await api.getPricing({ formatted: options.formatted })
+
+    if (outputMachine(pricing, options)) return
+
+    const { latestPro, latestBiz, sessionPro, sessionBiz, ...versions } = pricing
```
doistbot, line 18: [P2] The `...versions` rest split only works as long as the SDK never adds another top-level metadata field. Any new field here will be rendered as if it were a pricing version block. Please separate version entries with an explicit key filter (for example, matching the version-key pattern) or a typed helper that returns `{ metadata, versions }` so this view doesn't silently drift when the SDK response grows.

src/commands/billing/format.ts
```diff
@@ -0,0 +1,37 @@
...
+    const indent = options.json ? 2 : undefined
+    console.log(JSON.stringify(payload, null, indent))
+    return true
+}
+
+/** Format an amount given in a currency's minor units (e.g. cents) as money. */
+export function formatMoney(minorUnits: number, currency: string): string {
+    const code = currency.toUpperCase()
+    try {
+        return new Intl.NumberFormat('en-US', { style: 'currency', currency: code }).format(
```
doistbot: [P3] `Intl.NumberFormat` instantiation is relatively slow in V8. Since `formatMoney` is called inside deeply nested loops in `viewPricing` (for every term, currency, plan, and version), creating a new instance on every call can add avoidable overhead. Consider caching the formatter by currency code.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,312 @@
...
+const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
+
+function createProgram() {
+    const program = new Command()
+    program.exitOverride()
+    registerBillingCommand(program)
+    return program
+}
+
+const SUBSCRIPTION = {
```
doistbot: [P2] `CODEBASE.md` explicitly instructs: "do NOT hand-build mock entities. Use factories from `src/test-support/fixtures.ts`". Consider moving these billing payload mocks (`SUBSCRIPTION`, `PRO_PLAN_DETAILS`, `PRICES`, `PRICING`) into `fixtures.ts` and referencing them here to maintain a consistent test-support pattern.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,312 @@
...
+        await program.parseAsync(['node', 'td', 'billing', 'pricing', '--formatted'])
+
+        expect(mockApi.getPricing).toHaveBeenCalledWith({ formatted: true })
+        const output = consoleSpy.mock.calls.map((c) => c[0]).join('\n')
+        expect(output).toContain('pro (USD): $4/mo, $29/yr')
+        consoleSpy.mockRestore()
+    })
+})
+
+describe('billing wrapApiError → MISSING_SCOPE', () => {
```
doistbot: [P2] The `wrapApiError` function is a global API utility (`src/lib/api/core.ts`). Testing its behavior inside this feature command file violates the `AGENTS.md` rule to "Co-locate tests next to the command or lib module they cover." Consider moving this test block to `src/lib/api/core.test.ts`.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,312 @@
...
+describe('billing subscription', () => {
+    let mockApi: MockApi
+
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockApi = createMockApi()
+        mockGetApi.mockResolvedValue(mockApi)
+    })
+
+    it('is the default subcommand and renders the subscription summary', async () => {
```
doistbot, line 32: [P2] These command tests mock `getApi()` directly, so they never exercise the `permissions.ts` read-only gate. That leaves the riskiest change here untested: `getSubscriptionInfo` is a POST that only works because it was added to the safe-method allowlist. If that entry regresses, this suite still passes. Please add a permission/read-only test for the new billing methods, especially `getSubscriptionInfo`.

src/lib/oauth-scopes.test.ts
```diff
@@ -89,4 +93,30 @@ describe('formatScopesHelp', () => {
...
         expect(help).toContain('--additional-scopes=app-management,backups')
     })
+
+    it('lists the billing scope', () => {
+        expect(formatScopesHelp()).toContain('billing')
+    })
+})
+
+describe('resolveAuthScope', () => {
+    it('grants billing:read_write by default', () => {
```
doistbot, line 103: [P2] The new scope tests only cover `billing` in isolation. This PR also changes `AUTH_FLAG_ORDER` and the read-only scope mapping, so the higher-value case is a mixed set like `--read-only --additional-scopes=app-management,billing`. Add one combined-scope assertion (here or in the auth-provider/auth-flags tests) to catch ordering or `billing:read`/`billing:read_write` regressions when scopes are composed.

doistbot (commented):
This PR successfully introduces the `td billing` command group to surface the SDK's read-only billing endpoints alongside the corresponding opt-in OAuth scopes. The implementation is a great addition that cleanly integrates these new capabilities into the CLI's existing permissions framework. There are a few updates needed before merging, such as properly handling fractional digits for non-decimal currencies, refining the re-login scope hints for migrated users, and ensuring robust test coverage for the read-only API gates. Additionally, reusing shared CLI utilities for output formatting, caching the number formatter for performance, and moving mock entities and tests to their appropriate shared locations will help keep the codebase consistent.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F356&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A6e03b270-5604-11f1-821b-806960971686%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A6e03b270-5604-11f1-821b-806960971686&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/billing/format.ts
scottlovegrove: Fixed in 996fabe — `formatMoney` now derives the minor-unit divisor from the currency's fraction digits via `Intl` (USD→100, JPY→1, 3-decimal→1000) and drops decimals for whole amounts, mirroring todoist-web's `formatPrice`.

src/lib/api/core.ts
scottlovegrove: Fixed in 996fabe — `buildReloginCommand` now adds `--read-only` when `authMode === 'read-only'` even if `auth_flags` is unset, so migrated v1 users stay read-only and the scope/message stays aligned. Added a regression test.

src/commands/billing/format.ts
scottlovegrove: Done in 996fabe — `outputMachine` now uses `formatJson`/`formatNdjson` from `lib/output.js`.

src/commands/billing/format.ts
scottlovegrove: Done in 996fabe — `BillingViewOptions` is now `Pick<ViewOptions, 'json' | 'ndjson'>`.

src/commands/billing/pricing.ts
scottlovegrove: Fixed in 996fabe — the pricing view now skips non-object top-level fields, so any future scalar metadata won't be rendered as a pricing version.

src/commands/billing/format.ts
scottlovegrove: Done in 996fabe — `Intl.NumberFormat` (and the fraction-digit lookup) are now cached per currency.

src/commands/billing/billing.test.ts
scottlovegrove: Done in 996fabe — billing payloads moved to `fixtures.ts` (`fixtures.billing.*`) and referenced from the test.

src/commands/billing/billing.test.ts
scottlovegrove: Done in 996fabe — the `wrapApiError` MISSING_SCOPE block now lives in `src/lib/api/core.test.ts`.

src/commands/billing/billing.test.ts
scottlovegrove: Done in 996fabe — added a `permissions.test.ts` assertion that the four billing read methods (incl. the POST `getSubscriptionInfo`) are non-mutating, guarding the allowlist entry against regressions.

src/lib/oauth-scopes.test.ts
scottlovegrove: Done in 996fabe — added a `resolveAuthScope` test for `--read-only --additional-scopes=app-management,billing` asserting canonical order and `billing:read`, plus the read-write default.

craigcarlyle:
✅

scottlovegrove:
@doistbot /review

src/commands/billing/pricing.ts
```diff
@@ -0,0 +1,48 @@
...
+    console.log(`  Latest Pro:         ${latestPro}`)
+    console.log(`  Latest Business:    ${latestBiz}`)
+    console.log(`  Session Pro:        ${sessionPro}`)
+    console.log(`  Session Business:   ${sessionBiz}`)
+
+    for (const [version, plans] of Object.entries(versions)) {
+        // Version entries are objects (plan → currency → terms); skip any
+        // future top-level scalar metadata the SDK might add so it isn't
+        // rendered as if it were a pricing version.
+        if (typeof plans !== 'object' || plans === null) continue
```
doistbot: [P2] Filtering by `typeof plans === 'object'` handles scalar additions, but any new top-level object metadata added to the SDK response (e.g. `promotions: { ... }`) would still bypass this check and render as a pricing version. Filtering by the known version key pattern is safer and avoids printing garbage data:

```typescript
if (!/^v\d+/.test(version)) continue
```

doistbot (commented):
This PR introduces the `td billing` command group and its associated OAuth scopes to surface read-only billing endpoints from the updated SDK. The implementation provides comprehensive visibility into subscription statuses and pricing structures while integrating cleanly with the existing authentication and permissions model. There are a few areas for refinement, including optimizing network requests via concurrent fetching and deferred locale resolution, preventing future metadata additions from leaking into pricing outputs, and strengthening the test suite with currency-specific formatting cases and safer mock restoration.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F356&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9f1953b0-5622-11f1-9e2c-7020aa292d2b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9f1953b0-5622-11f1-9e2c-7020aa292d2b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/billing/pricing.ts
```diff
@@ -0,0 +1,48 @@
...
+import { getApi } from '../../lib/api/core.js'
+import { type BillingViewOptions, formatMoney, outputMachine, resolveLocale } from './format.js'
+
+interface PricingOptions extends BillingViewOptions {
+    formatted?: boolean
+}
+
+export async function viewPricing(options: PricingOptions = {}): Promise<void> {
+    const api = await getApi()
+    const locale = await resolveLocale(api, options)
```
doistbot: [P2] `viewPricing` resolves the user locale unconditionally, but the `--formatted` path never uses it: the endpoint already returns localized strings and `formatTerm()` ignores `locale` for string values. That adds an unnecessary `getUser()` dependency/request to every human-readable `td billing pricing --formatted` call. Skip `resolveLocale()` when `options.formatted` is true, or move locale resolution to the numeric-formatting branch.

src/commands/billing/subscription.ts
```diff
@@ -0,0 +1,40 @@
+import chalk from 'chalk'
+import { getApi } from '../../lib/api/core.js'
+import { type BillingViewOptions, formatMoney, outputMachine, resolveLocale } from './format.js'
+
+export async function viewSubscription(options: BillingViewOptions = {}): Promise<void> {
+    const api = await getApi()
+    const locale = await resolveLocale(api, options)
+    const info = await api.getSubscriptionInfo()
```
doistbot, line 8: [P2] `resolveLocale` (which fetches the user) and `api.getSubscriptionInfo()` make independent network requests. Await them concurrently with `Promise.all` to avoid unnecessary latency. This sequential pattern also appears in `plan.ts`, `prices.ts`, and `pricing.ts`.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockApi = createMockApi()
+        mockGetApi.mockResolvedValue(mockApi)
+        mockApi.getUser.mockResolvedValue({ lang: 'en' })
+    })
+
+    it('is the default subcommand and renders the subscription summary', async () => {
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
```
doistbot: [P2] Spying on `console.log` and manually restoring it at the end of the test body is brittle. If an assertion fails, the test throws and `mockRestore()` is skipped—leaving `console.log` muted for subsequent tests and hiding valuable error output. Consider managing the spy lifecycle in `beforeEach`/`afterEach` hooks or relying on Vitest's automatic mock restoration.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,229 @@
...
+    let mockApi: MockApi
+
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockApi = createMockApi()
+        mockGetApi.mockResolvedValue(mockApi)
+        mockApi.getUser.mockResolvedValue({ lang: 'en' })
+    })
+
+    it('formats minor-unit numbers as money and lists version entries', async () => {
```
doistbot, line 167: [P2] This only exercises USD/2-decimal formatting. The formatter change in this PR is specifically about currency-specific minor units, so add a 0-decimal currency case (JPY/KRW) and a 3-decimal currency case here or in a small `format.ts` unit test. Otherwise the bug fixed in `formatMoney` can regress while this suite still passes.

src/commands/billing/billing.test.ts
```diff
@@ -0,0 +1,229 @@
...
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+        mockApi.getPricing.mockResolvedValue(fixtures.billing.pricing)
```
doistbot, line 170: [P2] The suite doesn't cover the new `viewPricing` safeguard that skips non-object top-level metadata. Add a fixture with an extra scalar key (for example `revision: 3`) and assert it is not rendered as a version block, so that fix stays protected.

src/commands/billing/pricing.ts
scottlovegrove: Fixed in 520b04e — the loop now filters by version-key pattern (`if (!/^v\d+/.test(version)) continue`), so any added top-level field (scalar or object, e.g. `promotions`) is treated as metadata and never rendered as a version. Added a test asserting a non-version key isn't rendered.

src/commands/billing/pricing.ts
scottlovegrove: Fixed in 520b04e — `viewPricing` now skips `resolveLocale` (and its `getUser` fetch) when `--formatted` is set: `const locale = options.formatted ? undefined : await resolveLocale(...)`. Added a test asserting getUser isn't called on `--formatted`.

src/commands/billing/billing.test.ts
scottlovegrove: Fixed in 520b04e — the `console.log` spy is now created in a top-level `beforeEach` and restored in `afterEach`, so a failing assertion can't skip the restore.

src/commands/billing/billing.test.ts
scottlovegrove: Fixed in 520b04e — added `format.test.ts` covering JPY (0-decimal, asserts ¥700 not ¥7), BHD (3-decimal, asserts 1.5 not 15), and USD integer/decimal, plus the invalid-code fallback.

src/commands/billing/billing.test.ts
scottlovegrove: Fixed in 520b04e — added `fixtures.billing.pricingWithMetadata` (extra `revision: 3`) and a test asserting `revision` is not rendered while the real `v25` block still is.

src/commands/billing/subscription.ts
scottlovegrove: Keeping these sequential intentionally — it's a spinner-coherence tradeoff rather than an oversight. cli-core's spinner works by *adopting* the single early-spinner instance: with the sequential pattern, `getUser` adopts it (text "Checking authentication…"), releases it back on completion, then the billing call re-adopts the same instance (text "Loading billing…"), so the user sees one continuous spinner whose label updates. Under `Promise.all`, the first call adopts the early instance and the second finds none available, so it spins up a *second* concurrent yocto-spinner — two spinners rendering at once. We have a standing convention (see the `apps view` enrichment batch) that concurrent fetches should read as a single spinner. The latency cost is one short round-trip on an infrequent read command, which we considered the better trade. Happy to revisit if you'd prefer the parallelism.

craigcarlyle:
✅

PR #358, opened 2026-05-22
fix: Update @doist/cli-core to 0.21.0

scottlovegrove:
## Summary
- Bump `@doist/cli-core` 0.16.1 → 0.21.0
- `KeyringTokenStore` now requires `setBundle` and `activeBundle`; `createTodoistTokenStore` rebuilds the store as an object literal, so it forwards both methods
- `activeBundle` reuses the same `TODOIST_API_TOKEN` short-circuit as `active()` so the env-var override stays consistent across both read paths
- Updated the `auth.test.ts` snapshot mock store to implement the new methods
- Allows `td update` to check if the CLI was installed via Homebrew

## Test plan
- [x] `npm run build`
- [x] `npm run check`
- [x] `npm test` (1613 passing)
- [x] `tsc --noEmit` (pre-commit)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "CHANGELOG.md"
   ],
   "dependencies": {
-    "@doist/cli-core": "0.16.1",
+    "@doist/cli-core": "0.21.0",
     "@doist/todoist-sdk": "10.2.0",
     "@napi-rs/keyring": "1.3.0",
     "@pnpm/tabtab": "0.5.4",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "date-fns": "4.1.0",
     "marked": "18.0.3",
     "marked-terminal-renderer": "2.2.0",
+    "oauth4webapi": "3.8.6",
     "open": "11.0.0"
   },
   "devDependencies": {
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -341,6 +341,13 @@ describe('auth command', () => {
                         return { token: 'snapshot_token', account: SNAPSHOT_ACCOUNT }
                     },
                     async set() {},
+                    async setBundle() {},
+                    async activeBundle() {
+                        return {
+                            account: SNAPSHOT_ACCOUNT,
+                            bundle: { accessToken: 'snapshot_token' },
+                        }
+                    },
                     async clear() {},
                     async list() {
                         return [{ account: SNAPSHOT_ACCOUNT, isDefault: true }]
--- a/src/lib/auth-store.ts
+++ b/src/lib/auth-store.ts
@@ -64,15 +64,21 @@ export type TodoistTokenStore = KeyringTokenStore<TodoistAccount>
  * `UserRecordStore` adapter. Two Todoist-specific overlays on top of the
  * defaults:
  *
- *   - `active()` short-circuits to `null` when `TODOIST_API_TOKEN` is set.
- *     The env var is the canonical override across the CLI; without this
- *     short-circuit, `td auth status` would render the stored account while
- *     `getAuthMetadata()` reports `source: 'env'` (wrong account, right
- *     diagnostic).
+ *   - `active()` / `activeBundle()` short-circuit to `null` when
+ *     `TODOIST_API_TOKEN` is set. The env var is the canonical override
+ *     across the CLI; without this short-circuit, `td auth status` would
+ *     render the stored account while `getAuthMetadata()` reports
+ *     `source: 'env'` (wrong account, right diagnostic). Both read paths must
+ *     honour it so the bundle-aware `fetchLive` path agrees with `active()`.
  *   - `accountForUser` / `matchAccount` are passed explicitly. `matchAccount`
  *     delegates to `matchUserRef` so the keyring-store path and the
  *     config-driven `findUserByRef` path share one matcher (case-insensitive
  *     email + trim).
+ *
+ * Built with `Object.assign(Object.create(inner), …)` (the same pattern as
+ * `withUserRefAware()`) so every method other than the two env-aware reads is
+ * inherited from `inner` via the prototype chain — a future cli-core method
+ * addition resolves automatically instead of being silently dropped.
  */
 export function createTodoistTokenStore(): TodoistTokenStore {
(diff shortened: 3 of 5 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
     "CHANGELOG.md"
   ],
   "dependencies": {
-    "@doist/cli-core": "0.16.1",
+    "@doist/cli-core": "0.21.0",
```
doistbot, line 54: [P1] `@doist/cli-core@0.21.0` adds `oauth4webapi` as a peer, but this PR doesn't install it. We do use cli-core's PKCE flow via `createPkceProvider()` in `src/lib/auth-provider.ts`, so `td auth login` will hit `ERR_MODULE_NOT_FOUND` in installed builds. Please add `oauth4webapi` to `dependencies` and update the lockfile.

src/lib/auth-store.ts
```diff
@@ -89,6 +89,11 @@ export function createTodoistTokenStore(): TodoistTokenStore {
             return inner.active(ref)
         },
         set: (account, token) => inner.set(account, token),
+        setBundle: (account, bundle, options) => inner.setBundle(account, bundle, options),
+        activeBundle: async (ref) => {
```
doistbot: [P2] This duplicates the `TODOIST_API_TOKEN` short-circuit that's already in `active()`. Since the goal here is to keep both read paths behaving identically, it would be better to route both methods through one small shared wrapper/helper instead of copying the guard in two places.

src/lib/auth-store.ts
```diff
@@ -89,6 +89,11 @@ export function createTodoistTokenStore(): TodoistTokenStore {
             return inner.active(ref)
         },
         set: (account, token) => inner.set(account, token),
+        setBundle: (account, bundle, options) => inner.setBundle(account, bundle, options),
```
doistbot: [P2] This wrapper is still re-declaring the `KeyringTokenStore` surface as a plain object, which is what made the 0.21.0 upgrade miss `setBundle`/`activeBundle` in the first place. Prefer the same prototype-preserving pattern used in `withUserRefAware()`: `Object.assign(Object.create(inner), { active, activeBundle })`, and let the unchanged methods come from `inner` so future cli-core additions don't need another manual sync.

doistbot (commented):
This PR bumps `@doist/cli-core` to `0.21.0` and updates the `KeyringTokenStore` to support the newly required `setBundle` and `activeBundle` methods. These updates are a solid step toward integrating the latest core capabilities. A few refinements are needed before merging, including installing the missing `oauth4webapi` peer dependency to prevent runtime login errors, refactoring the token store to use a prototype-preserving pattern instead of a plain object to future-proof the implementation, deduplicating the environment variable short-circuit logic, and adding test coverage for the `activeBundle` override behavior.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F358&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac77c2bc0-561d-11f1-93e8-b1552bba4956%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac77c2bc0-561d-11f1-93e8-b1552bba4956&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/auth.test.ts
```diff
@@ -341,6 +341,13 @@ describe('auth command', () => {
                         return { token: 'snapshot_token', account: SNAPSHOT_ACCOUNT }
                     },
                     async set() {},
+                    async setBundle() {},
+                    async activeBundle() {
```
doistbot, line 345: [P2] This new stub satisfies the `activeBundle()` interface, but the suite still doesn't test the behavior this PR is fixing: when `TODOIST_API_TOKEN` is set, the bundle lookup must short-circuit the same way `active()` does. Please add a focused test that sets the env var and asserts `auth status` takes the env path (or add a small `auth-store` unit test that `activeBundle()` returns `null` without touching the inner store). Right now a regression here would pass this suite.

package.json
scottlovegrove: Done — added `oauth4webapi@3.8.6` to `dependencies` and updated the lockfile. Confirmed it's a no-engines package, so node 20 is fine.

src/lib/auth-store.ts
scottlovegrove: Done — rebuilt with `Object.assign(Object.create(inner), { active, activeBundle })`, matching `withUserRefAware()`. Only the two env-aware reads are overridden now; everything else inherits from `inner` via the prototype chain.

src/lib/auth-store.ts
scottlovegrove: Done — both paths now route through one `envTokenSet()` helper instead of duplicating the guard.

src/commands/auth/auth.test.ts
scottlovegrove: Done — added `src/lib/auth-store.test.ts` with focused unit tests: `active()` and `activeBundle()` both return `null` without touching the inner store when `TODOIST_API_TOKEN` is set, plus an inner-store delegation case when it isn't.

rmartins90 (commented):
✅

PR #360, opened 2026-05-23
feat: rename `td user` to `td accounts` and adopt cli-core account attachers

scottlovegrove:
## Summary
- Rename the top-level `td user` command group to `td accounts`, keeping `user` and `users` as non-breaking aliases (wired through the lazy command dispatcher; tab-completion suggests all three).
- Move **all four** account subcommands onto cli-core's generic attachers, so they run on the `TokenStore` contract instead of hand-rolled logic:
  - `accounts list` → `attachAccountListCommand`
  - `accounts use` / `default` → `attachAccountUseCommand`
  - `accounts current` → `attachAccountCurrentCommand`
  - `accounts remove` → `attachAccountRemoveCommand`
- Unify the "default account" rule (`getEffectiveDefaultUserId`) across `auth status` and `config view` (the `list`/`current` markers come from the store directly).
- Update user-facing hints across the CLI to point at `td accounts …`.

## cli-core dependency
Requires **@doist/cli-core 0.23.0** (now pinned) — the release containing all four account attachers plus the `TokenStore` contract changes (`clear()` returns `ClearedAccount`, new optional `activeAccount()`).

## Behavior changes
- **`accounts list --json`/`--ndjson`**: cli-core's `{ accounts, default }` envelope (per-account entries keep the existing camelCase fields). The `storage` field is dropped (`store.list()` has no token-source info); empty `--json` is `{"accounts":[],"default":null}`.
- **`accounts current`**: output preserved (stored account renders with accurate `source`; env/legacy via `onNotAuthenticated`). Gains `--ndjson`. An `activeAccount` env-override is added to the store adapter so `TODOIST_API_TOKEN` still wins over a stored default.
- **`accounts remove`**: `store.clear(ref)` resolves + deletes by canonical id and reports `wasDefault` (token-free, so a broken keyring entry stays removable). Gains `--json` (`{ ok, removed }`) / silent `--ndjson`; human output and the keyring-fallback warning preserved.
- **`accounts use` / `remove`**: don't honor `--quiet` and use Commander's required-arg error for a missing `<ref>` — both are cli-core-wide attacher behaviors (also true of `auth login`/`logout`/`status`); proper fixes belong in cli-core's `emitView`.
- **Aliases**: `td user …` and `td users …` continue to work unchanged.

## Versioning
The `accounts list --json` reshape is an intentional, low-impact change to a niche list-users machine surface; shipping as a minor `feat` (not `BREAKING CHANGE`).

## Test plan
- [x] `npm run type-check`
- [x] `npm test` (1654 pass)
- [x] `check:skill-sync`
- [x] Smoke-tested `accounts list/current/remove` (+ `--json`/`--ndjson`, env-token path) and the `user`/`users` aliases against the linked cli-core build

### DIFF
--- a/CODEBASE.md
+++ b/CODEBASE.md
@@ -127,6 +127,12 @@ New subcommand? Copy a sibling in the target group, wire it in that group's
 - **`auth.ts`** — read-side resolver: `resolveActiveUser`, `getApiToken`,
   `probeApiToken`, `getAuthMetadata`, `listStoredUsers`, `NoTokenError`. All
   write/clear paths go through `auth-store.ts`.
+- **`users.ts`** — config-side account helpers shared across commands:
+  `matchUserRef`/`findUserByRef`/`requireUserByRef`, the stored-user mutators
+  (`upsert`/`remove`/`set`/`clearDefaultUser`), `getDefaultUserId` (raw pinned
+  pointer) and `getEffectiveDefaultUserId` (pinned-else-sole-account, used for
+  the `(default)` marker in `accounts list`/`current`, `auth status`,
+  `config view`).
 - **`auth-flags.ts`** — `buildReloginCommand()` (rebuilds `td auth login`
   with `--read-only` / `--additional-scopes=...` preserved)
 - **`config.ts`** — `~/.config/todoist-cli/config.json` read/write,
--- a/CODEBASE.md
+++ b/CODEBASE.md
@@ -109,9 +109,10 @@ lives in `src/lib/skills/content.ts` (SKILL_CONTENT) — don't duplicate here.
 - **Notifications** — list/view/accept/reject/dismiss
 - **Productivity & activity** — `activity`, `stats`, `completed`
 - **Top-level views** — `today`, `upcoming`, `inbox`, `view` (URL router)
-- **Infra** — `auth` (login/logout/token/status), `settings`, `apps`,
-  `backup`, `attachment`, `hc` (Help Center), `skill`, `completion`, `update`,
-  `doctor`, `changelog`
+- **Infra** — `auth` (login/logout/token/status), `accounts`
+  (multi-user list/use/current/remove; aliases `user`/`users`), `settings`,
+  `apps`, `backup`, `attachment`, `hc` (Help Center), `skill`, `completion`,
+  `update`, `doctor`, `changelog`
 
 New subcommand? Copy a sibling in the target group, wire it in that group's
 `index.ts`, update `SKILL_CONTENT`, run `npm run sync:skill`. See AGENTS.md.
@@ -219,9 +220,11 @@ All live in `src/lib/refs.ts`:
 ## Auth & token storage
 
 `@doist/cli-core/auth` owns the keyring, multi-user `TokenStore`, OAuth flow,
-and the `login` / `logout` / `status` / `token view` registrars. todoist-cli
-supplies a `UserRecordStore<TodoistAccount>` adapter (`user-records.ts`) over
-its config file plus a Todoist-specific `validateToken` (`auth-provider.ts`).
+and the `login` / `logout` / `status` / `token view` registrars, plus the
+`account list` / `account use` attachers that back `td accounts list` and
+`td accounts use` (`commands/user/{list,use}.ts`). todoist-cli supplies a
+`UserRecordStore<TodoistAccount>` adapter (`user-records.ts`) over its config
+file plus a Todoist-specific `validateToken` (`auth-provider.ts`).
 
 Read path (`auth.ts` — `resolveActiveUser` / `getApiToken` / `probeApiToken`):
 env `TODOIST_API_TOKEN` first, then a config-derived target user, then either
--- a/src/commands/user/current.ts
+++ b/src/commands/user/current.ts
@@ -1,50 +1,75 @@
+import { formatJson, formatNdjson } from '@doist/cli-core'
+import { attachAccountCurrentCommand } from '@doist/cli-core/auth'
 import chalk from 'chalk'
-import { readConfig, resolveActiveUser } from '../../lib/auth.js'
-import { getEffectiveDefaultUserId } from '../../lib/users.js'
+import type { Command } from 'commander'
+import type { TodoistAccount, TodoistTokenStore } from '../../lib/auth-store.js'
+import { resolveActiveUser } from '../../lib/auth.js'
 
-export interface CurrentUserOptions {
-    json?: boolean
-}
-
-export async function currentUserCommand(options: CurrentUserOptions): Promise<void> {
-    const resolved = await resolveActiveUser()
-    const config = await readConfig()
-    const defaultId = getEffectiveDefaultUserId(config)
-    const isDefault = resolved.id === defaultId
-
-    if (options.json) {
-        console.log(
-            JSON.stringify(
-                {
+/**
+ * Attach `td accounts current` via cli-core's `attachAccountCurrentCommand`.
+ *
+ * A signed-in stored account resolves through `store.activeAccount()` and is
+ * rendered by `renderText` / `renderJson`. The env-token and legacy
+ * single-user sources live outside the store (so `activeAccount()` returns
+ * `null` for them — env via the adapter's short-circuit, legacy because it has
+ * no record), and are rendered in `onNotAuthenticated` via the read-side
+ * resolver. The stored-case `--json` `source` is reported as `secure-store`;
+ * the env/legacy sources are surfaced accurately by the fallback branch.
+ */
+export function attachTodoistUserCurrentCommand(
+    parent: Command,
+    store: TodoistTokenStore,
+): Command {
+    return attachAccountCurrentCommand<TodoistAccount>(parent, {
+        store,
+        description: 'Show the active account (resolved from --user, default, or single login)',
+        renderText: ({ account, isDefault }) => {
+            const marker = isDefault ? chalk.green(' (default)') : ''
+            return `${account.email} ${chalk.dim(`(id:${account.id})`)}${marker}`
+        },
+        renderJson: ({ account, isDefault }) => ({
+            id: account.id,
+            email: account.email,
+            source: 'secure-store',
+            isDefault,
+            authMode: account.auth_mode,
+            authScope: account.auth_scope,
+            authFlags: account.auth_flags,
+        }),
+        onNotAuthenticated: async ({ view }) => {
+            // Nothing resolved in the store, so the active credential is
+            // out-of-store (TODOIST_API_TOKEN or legacy single-user creds) — or
+            // there's nothing at all, in which case `resolveActiveUser` throws
+            // `NoTokenError`, matching the previous behaviour.
+            const resolved = await resolveActiveUser()
+            if (view.json || view.ndjson) {
+                const payload = {
                     id: resolved.id === 'env' || resolved.id === 'legacy' ? null : resolved.id,
                     email: resolved.email || null,
                     source: resolved.source,
-                    isDefault,
+                    isDefault: false,
                     authMode: resolved.authMode,
                     authScope: resolved.authScope,
                     authFlags: resolved.authFlags,
-                },
-                null,
-                2,
-            ),
-        )
-        return
-    }
-
-    if (resolved.source === 'env') {
-        console.log(chalk.dim('Using TODOIST_API_TOKEN environment variable (no stored user).'))
-        return
-    }
-
-    if (resolved.id === 'legacy') {
-        console.log(
-            chalk.dim(
-                'Using legacy single-user credentials. Run `td auth login` to migrate to a stored account.',
-            ),
-        )
-        return
-    }
-
-    const marker = isDefault ? chalk.green(' (default)') : ''
-    console.log(`${resolved.email} ${chalk.dim(`(id:${resolved.id})`)}${marker}`)
+                }
+                console.log(view.json ? formatJson(payload) : formatNdjson([payload]))
+                return
+            }
+            if (resolved.source === 'env') {
+                console.log(
+                    chalk.dim('Using TODOIST_API_TOKEN environment variable (no stored user).'),
+                )
+                return
+            }
+            if (resolved.id === 'legacy') {
+                console.log(
+                    chalk.dim(
+                        'Using legacy single-user credentials. Run `td auth login` to migrate to a stored account.',
+                    ),
+                )
+                return
+            }
+            console.log(`${resolved.email} ${chalk.dim(`(id:${resolved.id})`)}`)
+        },
+    })
 }
--- a/src/commands/user/current.ts
+++ b/src/commands/user/current.ts
@@ -4,6 +4,7 @@ import chalk from 'chalk'
 import type { Command } from 'commander'
 import type { TodoistAccount, TodoistTokenStore } from '../../lib/auth-store.js'
 import { resolveActiveUser } from '../../lib/auth.js'
+import { projectAccount } from './project-account.js'
 
 /**
  * Attach `td accounts current` via cli-core's `attachAccountCurrentCommand`.
@@ -28,13 +29,8 @@ export function attachTodoistUserCurrentCommand(
             return `${account.email} ${chalk.dim(`(id:${account.id})`)}${marker}`
         },
         renderJson: ({ account, isDefault }) => ({
-            id: account.id,
-            email: account.email,
+            ...projectAccount(account, isDefault),
             source: 'secure-store',
-            isDefault,
-            authMode: account.auth_mode,
-            authScope: account.auth_scope,
-            authFlags: account.auth_flags,
         }),
         onNotAuthenticated: async ({ view }) => {
             // Nothing resolved in the store, so the active credential is
--- a/src/commands/user/current.ts
+++ b/src/commands/user/current.ts
@@ -1,6 +1,6 @@
 import chalk from 'chalk'
 import { readConfig, resolveActiveUser } from '../../lib/auth.js'
-import { getDefaultUserId } from '../../lib/users.js'
+import { getEffectiveDefaultUserId } from '../../lib/users.js'
 
 export interface CurrentUserOptions {
     json?: boolean
@@ -9,7 +9,7 @@ export interface CurrentUserOptions {
 export async function currentUserCommand(options: CurrentUserOptions): Promise<void> {
     const resolved = await resolveActiveUser()
     const config = await readConfig()
-    const defaultId = getDefaultUserId(config)
+    const defaultId = getEffectiveDefaultUserId(config)
     const isDefault = resolved.id === defaultId
 
     if (options.json) {
--- a/src/commands/user/current.ts
+++ b/src/commands/user/current.ts
@@ -14,8 +14,9 @@ import { projectAccount } from './project-account.js'
  * single-user sources live outside the store (so `activeAccount()` returns
  * `null` for them — env via the adapter's short-circuit, legacy because it has
  * no record), and are rendered in `onNotAuthenticated` via the read-side
- * resolver. The stored-case `--json` `source` is reported as `secure-store`;
- * the env/legacy sources are surfaced accurately by the fallback branch.
+ * resolver. The stored-case `--json` `source` is read off the account (the
+ * adapter's `activeAccount` annotates `secure-store` vs `config-file`); the
+ * env/legacy sources come from the fallback branch.
  */
 export function attachTodoistUserCurrentCommand(
     parent: Command,
@@ -30,7 +31,7 @@ export function attachTodoistUserCurrentCommand(
         },
         renderJson: ({ account, isDefault }) => ({
             ...projectAccount(account, isDefault),
-            source: 'secure-store',
+            source: (account.source as string | undefined) ?? 'secure-store',
         }),
         onNotAuthenticated: async ({ view }) => {
             // Nothing resolved in the store, so the active credential is
--- a/src/commands/user/index.ts
+++ b/src/commands/user/index.ts
@@ -1,5 +1,6 @@
 import { Command } from 'commander'
 import { createTodoistTokenStore } from '../../lib/auth-store.js'
+import { ACCOUNT_COMMAND_ALIASES } from './aliases.js'
 import { currentUserCommand } from './current.js'
 import { attachTodoistUserListCommand } from './list.js'
 import { removeUserCommand } from './remove.js'
@@ -10,7 +11,7 @@ export function registerUserCommand(program: Command): void {
     // existing scripts (and the docs that long said `td user …`) keep working.
     const account = program
         .command('accounts')
-        .aliases(['user', 'users'])
+        .aliases(ACCOUNT_COMMAND_ALIASES)
         .description('Manage stored Todoist accounts (multi-user)')
 
     // `list` and `use`/`default` delegate to cli-core's generic account
--- a/src/commands/user/index.ts
+++ b/src/commands/user/index.ts
@@ -1,9 +1,9 @@
 import { Command } from 'commander'
 import { createTodoistTokenStore } from '../../lib/auth-store.js'
 import { ACCOUNT_COMMAND_ALIASES } from './aliases.js'
-import { currentUserCommand } from './current.js'
+import { attachTodoistUserCurrentCommand } from './current.js'
 import { attachTodoistUserListCommand } from './list.js'
-import { removeUserCommand } from './remove.js'
+import { attachTodoistUserRemoveCommand } from './remove.js'
 import { attachTodoistUserUseCommand } from './use.js'
 
 export function registerUserCommand(program: Command): void {
@@ -14,23 +14,13 @@ export function registerUserCommand(program: Command): void {
         .aliases(ACCOUNT_COMMAND_ALIASES)
         .description('Manage stored Todoist accounts (multi-user)')
 
-    // `list` and `use`/`default` delegate to cli-core's generic account
-    // attachers (consuming `store.list()` / `store.setDefault()`); `current`
-    // and `remove` stay hand-rolled — cli-core ships no attacher for them.
+    // All four subcommands delegate to cli-core's generic account attachers
+    // (list/use/current/remove), consuming the `TokenStore` contract directly.
     const store = createTodoistTokenStore()
     attachTodoistUserListCommand(account, store)
     attachTodoistUserUseCommand(account, store)
-
-    account
-        .command('current')
-        .description('Show the active account (resolved from --user, default, or single login)')
-        .option('--json', 'Output as JSON')
-        .action(currentUserCommand)
-
-    account
-        .command('remove <ref>')
-        .description('Remove a stored account (deletes its token and config entry)')
-        .action((ref: string) => removeUserCommand(ref))
+    attachTodoistUserCurrentCommand(account, store)
+    attachTodoistUserRemoveCommand(account, store)
 
     account.addHelpText(
         'after',
--- a/src/commands/user/index.ts
+++ b/src/commands/user/index.ts
@@ -1,45 +1,45 @@
 import { Command } from 'commander'
+import { createTodoistTokenStore } from '../../lib/auth-store.js'
 import { currentUserCommand } from './current.js'
-import { listUsersCommand } from './list.js'
+import { attachTodoistUserListCommand } from './list.js'
 import { removeUserCommand } from './remove.js'
-import { useUserCommand } from './use.js'
+import { attachTodoistUserUseCommand } from './use.js'
 
 export function registerUserCommand(program: Command): void {
-    const user = program.command('user').description('Manage stored Todoist accounts (multi-user)')
+    // Renamed from `user` to `accounts`; `user`/`users` stay as aliases so
+    // existing scripts (and the docs that long said `td user …`) keep working.
+    const account = program
+        .command('accounts')
+        .aliases(['user', 'users'])
+        .description('Manage stored Todoist accounts (multi-user)')
 
-    user.command('list')
-        .description('List all stored Todoist accounts')
-        .option('--json', 'Output as JSON')
-        .option('--ndjson', 'Output as newline-delimited JSON')
-        .action(listUsersCommand)
+    // `list` and `use`/`default` delegate to cli-core's generic account
+    // attachers (consuming `store.list()` / `store.setDefault()`); `current`
+    // and `remove` stay hand-rolled — cli-core ships no attacher for them.
+    const store = createTodoistTokenStore()
+    attachTodoistUserListCommand(account, store)
+    attachTodoistUserUseCommand(account, store)
 
-    user.command('use <ref>')
-        .description('Set the default account used when --user is not provided')
-        .action((ref: string) => useUserCommand(ref))
-
-    // `default` is an explicit alias for `use` — same behavior, different verb.
-    user.command('default <ref>')
-        .description('Alias of `td user use <ref>`')
-        .action((ref: string) => useUserCommand(ref))
-
-    user.command('current')
+    account
+        .command('current')
         .description('Show the active account (resolved from --user, default, or single login)')
         .option('--json', 'Output as JSON')
         .action(currentUserCommand)
 
-    user.command('remove <ref>')
+    account
+        .command('remove <ref>')
         .description('Remove a stored account (deletes its token and config entry)')
         .action((ref: string) => removeUserCommand(ref))
 
-    user.addHelpText(
+    account.addHelpText(
         'after',
         `
 Examples:
-  $ td auth login                 # add an account (sets it as default if first)
-  $ td user list                  # see all stored accounts
-  $ td user use scott@doist.com   # set default
-  $ td user current               # show the active account
+  $ td auth login                     # add an account (sets it as default if first)
+  $ td accounts list                  # see all stored accounts
+  $ td accounts use scott@doist.com   # set default
+  $ td accounts current               # show the active account
   $ td --user other@example.com task list   # one-off override
-  $ td user remove old@example.com`,
+  $ td accounts remove old@example.com`,
     )
 }
--- a/src/commands/user/list.ts
+++ b/src/commands/user/list.ts
@@ -1,53 +1,56 @@
-import { formatJson, formatNdjson, printEmpty } from '@doist/cli-core'
+import { attachAccountListCommand } from '@doist/cli-core/auth'
 import chalk from 'chalk'
-import { listStoredUsers, readConfig, type StoredUser } from '../../lib/auth.js'
+import type { Command } from 'commander'
+import type { TodoistAccount, TodoistTokenStore } from '../../lib/auth-store.js'
 import { isAccessible } from '../../lib/global-args.js'
-import { getDefaultUserId } from '../../lib/users.js'
 
-export interface ListUsersOptions {
-    json?: boolean
-    ndjson?: boolean
-}
-
-function projectUser(u: StoredUser, defaultId: string | undefined) {
+/**
+ * Per-account `--json` / `--ndjson` shape. Flattens the stored
+ * `TodoistAccount` into the camelCase fields the CLI has always surfaced,
+ * dropping the internal `label` duplicate and snake_case config keys. The
+ * `{ accounts, default }` envelope around these entries is owned by cli-core's
+ * `attachAccountListCommand`.
+ */
+function projectAccount(account: TodoistAccount, isDefault: boolean) {
     return {
-        id: u.id,
-        email: u.email,
-        isDefault: u.id === defaultId,
-        authMode: u.auth_mode,
-        authScope: u.auth_scope,
-        authFlags: u.auth_flags,
-        storage: u.api_token ? 'config-file' : 'secure-store',
+        id: account.id,
+        email: account.email,
+        isDefault,
+        authMode: account.auth_mode,
+        authScope: account.auth_scope,
+        authFlags: account.auth_flags,
     }
 }
 
-export async function listUsersCommand(options: ListUsersOptions): Promise<void> {
-    const users = await listStoredUsers()
-    const config = await readConfig()
-    const defaultId = getDefaultUserId(config)
-
-    if (users.length === 0) {
-        printEmpty({
-            options,
-            message: chalk.dim('No stored Todoist accounts. Run `td auth login` to add one.'),
-        })
-        return
-    }
-
-    if (options.json) {
-        console.log(formatJson(users.map((u) => projectUser(u, defaultId))))
-        return
-    }
-
-    if (options.ndjson) {
-        console.log(formatNdjson(users.map((u) => projectUser(u, defaultId))))
-        return
-    }
-
-    const accessible = isAccessible()
-    for (const u of users) {
-        const isDefault = u.id === defaultId
-        const marker = isDefault ? (accessible ? ' [default]' : chalk.green(' (default)')) : ''
-        console.log(`${u.email} ${chalk.dim(`(id:${u.id})`)}${marker}`)
-    }
+/**
+ * Attach `td accounts list` via cli-core's generic `attachAccountListCommand`,
+ * which reads `store.list()` and owns the `{ accounts, default }` machine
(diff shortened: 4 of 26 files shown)


src/commands/user/use.ts
```diff
@@ -1,22 +1,17 @@
...
+ * generic `attachAccountUseCommand`, which calls `store.setDefault(ref)` and
+ * owns the success line / `--json` (`{ ok, default }`) / silent-`--ndjson`
+ * output. `setDefault`'s `CliError('ACCOUNT_NOT_FOUND', …)` on a ref miss
+ * propagates to the top-level handler unchanged.
+ */
+export function attachTodoistUserUseCommand(parent: Command, store: TodoistTokenStore): Command {
+    return attachAccountUseCommand<TodoistAccount>(parent, {
+        store,
+        description: 'Set the default account used when --user is not provided',
+    }).alias('default')
```
doistbot, line 16: [P2] Turning `default` into an alias regresses command-name completion. `src/lib/completion.ts` only suggests `c.name()`, not `c.aliases()`, so `td user def<TAB>` will no longer complete even though `default` used to be a real subcommand. Either keep `default` as its own command or make subcommand completion include aliases.

src/commands/workspace/helpers.ts
```diff
@@ -63,7 +63,7 @@ export async function resolveWorkspaceUserRef(
         const user = await findUser(workspaceId, (u) => u.userId === id)
         if (user) return user
         throw new CliError('NOT_FOUND', `No user with id "${id}" found in workspace.`, [
-            'Use `td workspace users <workspace>` to list members.',
+            'Use `td workspace accounts <workspace>` to list members.',
```
doistbot: [P2] This command hint is still duplicated elsewhere in this helper, and the copies have already drifted (`accounts` here, `users` in the other NOT_FOUND branch below). Please pull the list-members hint into a shared constant/helper and reuse it in both branches so future renames stay in sync.

src/commands/user/list.ts
```diff
@@ -1,53 +1,56 @@
...
+ * envelope. The Todoist overrides keep human output identical to the
+ * hand-rolled version (`<email> (id:<id>)` + a default marker that respects
+ * `--accessible`) and flatten each machine entry via `projectAccount`.
+ *
+ * `isDefault` now reflects the store's *effective* default (a lone account
+ * counts as default even with no pinned `defaultUser`), matching what
+ * `active()` would resolve.
+ */
+export function attachTodoistUserListCommand(parent: Command, store: TodoistTokenStore): Command {
+    return attachAccountListCommand<TodoistAccount>(parent, {
```
doistbot, line 20: [P1] This makes `td user list --json` adopt cli-core's `{ accounts, default }` envelope instead of the CLI's existing array contract. `user list` is already a public machine interface, and the rest of this repo usually curates JSON shapes independently of helper internals. Please keep the attacher internal and reformat back to the legacy array shape (or version the output) so a dependency swap doesn't become a user-visible API break.

src/commands/user/list.ts
```diff
@@ -1,53 +1,56 @@
...
-        console.log(`${u.email} ${chalk.dim(`(id:${u.id})`)}${marker}`)
-    }
+/**
+ * Attach `td user list` via cli-core's generic `attachAccountListCommand`,
+ * which reads `store.list()` and owns the `{ accounts, default }` machine
+ * envelope. The Todoist overrides keep human output identical to the
+ * hand-rolled version (`<email> (id:<id>)` + a default marker that respects
+ * `--accessible`) and flatten each machine entry via `projectAccount`.
+ *
+ * `isDefault` now reflects the store's *effective* default (a lone account
```
doistbot, line 15: [P2] Changing `isDefault` to the store's effective default only here creates two definitions of "default" across the CLI. `td user current`, `td auth status`, and `td config view` still key off the pinned `defaultUser`, so a single stored account with no pinned default will now show different default markers depending on which command you run. Please move this rule into shared auth/user logic and apply it consistently before switching this one command.

doistbot (commented):
This PR nicely updates the multi-account commands by adopting generic attachers from `cli-core` 0.22.0 and smoothly transitioning `workspace users` to `workspace accounts`, which reduces boilerplate and improves alignment with the core library. However, a few areas need further refinement, specifically around preserving the existing `user list --json` output contract to avoid breaking changes and ensuring consistent "default" account logic across the CLI. Additional adjustments are also noted to address minor regressions in autocomplete, consolidate duplicated command hints, and ensure that both test coverage and `CODEBASE.md` accurately reflect the new capabilities and structural changes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F360&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac74b00a0-5680-11f1-8c7f-b623bf5fb731%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac74b00a0-5680-11f1-8c7f-b623bf5fb731&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/user/user.test.ts
```diff
@@ -85,59 +95,49 @@ describe('user command', () => {
...
+            })
         })
     })
 
     describe('use', () => {
         beforeEach(() => {
             setDefaultMock.mockReset().mockResolvedValue(undefined)
+            // The attacher re-reads `store.list()` after `setDefault` to
+            // resolve the canonical default id for `--json` output.
+            listMock.mockReset().mockResolvedValue([])
```
doistbot, line 164: [P2] The PR description notes that `user use` gains `--json`/`--ndjson` support. Per `AGENTS.md` ("When adding features, add corresponding tests"), please add test coverage for this new capability. You will likely need to update this mock to return the set account rather than `[]` so that the canonical default resolves correctly in the JSON output.

src/commands/user/user.test.ts
```diff
@@ -85,59 +95,49 @@ describe('user command', () => {
             expect(lines).toContain('default')
         })
 
-        it('emits one JSON value per line for --ndjson with stored accounts', async () => {
-            mockListStoredUsers.mockResolvedValue([
-                { id: '1', email: 'a@b.c', auth_mode: 'read-write' },
-                { id: '2', email: 'd@e.f', auth_mode: 'read-only' },
+        it('streams one JSON value per console.log for --ndjson with stored accounts', async () => {
```
doistbot: [P2] This NDJSON case now asserts the number of `console.log` calls instead of the emitted line-delimited payload. That makes the test fail if the implementation switches between batching lines into one write vs logging each line separately, even though the CLI output is still correct. Assert the joined stdout content split on `\n` instead of the call count.

src/commands/user/index.ts
```diff
@@ -1,26 +1,19 @@
...
-
-    // `default` is an explicit alias for `use` — same behavior, different verb.
-    user.command('default <ref>')
-        .description('Alias of `td user use <ref>`')
-        .action((ref: string) => useUserCommand(ref))
+    // `list` and `use`/`default` delegate to cli-core's generic account
+    // attachers (consuming `store.list()` / `store.setDefault()`); `current`
+    // and `remove` stay hand-rolled — cli-core ships no attacher for them.
+    const store = createTodoistTokenStore()
+    attachTodoistUserListCommand(user, store)
```
doistbot: [P2] `AGENTS.md` says `CODEBASE.md` must be updated when command registration or auth/token-storage wiring changes. This PR does both: `user list`/`user use` now delegate through cli-core attachers, and `workspace users` is being renamed to `workspace accounts`, but `CODEBASE.md` still documents the old `workspace users` naming and the older auth delegation. Please update `CODEBASE.md` in this PR so the repo map stays accurate.

scottlovegrove:
@doistbot /review

src/commands/user/use.ts
scottlovegrove: Fixed in eb7aca1 — `getSubcommandCompletions` now emits each command's aliases, so `default` completes again (along with the `user`/`users` group aliases). Kept `default` as an alias rather than a standalone command since both route through the same cli-core `attachAccountUseCommand`; the completion layer is where the gap was.

src/commands/workspace/helpers.ts
scottlovegrove: No longer applies — the `workspace users` → `workspace accounts` rename was an incorrect interpretation and has been reverted, so this helper is back to `main` untouched. The pre-existing duplicate hint lives in `main` and is out of scope for this PR.

src/commands/user/list.ts
scottlovegrove: Intentional, and shipping as a minor `feat` (not `BREAKING CHANGE`). The envelope only affects `accounts list --json`/`--ndjson` — a niche list-users machine surface whose consumers are scripts/agents that adapt in one line; the human command contract (names, flags, human output) is unchanged. A `BREAKING CHANGE` footer would trigger a major bump that over-signals for this surface, so we're documenting the reshape in the PR body / release notes instead of versioning it as a break.

src/commands/user/list.ts
scottlovegrove: Fixed in eb7aca1 — added `getEffectiveDefaultUserId` (pinned `defaultUser` ?? the sole stored account) and applied it in `accounts current`, `auth status`, and `config view`, so the `(default)` marker matches what `accounts list` derives from the store. `doctor` and `accounts remove` deliberately keep the raw pinned pointer, where the pin itself is what's being diagnosed/cleared.

src/commands/user/user.test.ts
scottlovegrove: Added in eb7aca1 — `accounts use --json` now has coverage asserting `{ ok: true, default }` (with the canonical id resolved from the post-`setDefault` `list()` re-read, so the mock returns the set account), and a `--ndjson` case asserting it stays silent.

src/commands/user/user.test.ts
scottlovegrove: Fixed in eb7aca1 — the ndjson list test now joins stdout and splits on newlines, asserting the parsed per-line payloads rather than the `console.log` call count, so it's agnostic to batching vs per-line writes.

src/commands/user/index.ts
scottlovegrove: Updated CODEBASE.md in eb7aca1 for the `accounts` rename (+ `user`/`users` aliases) and the `account list`/`account use` attacher delegation. The `workspace users` naming in CODEBASE.md is accurate again since that rename was reverted out of this PR.

src/commands/user/list.ts
```diff
@@ -1,53 +1,56 @@
...
+ * counts as default even with no pinned `defaultUser`), matching what
+ * `active()` would resolve.
+ */
+export function attachTodoistUserListCommand(parent: Command, store: TodoistTokenStore): Command {
+    return attachAccountListCommand<TodoistAccount>(parent, {
```
doistbot, line 20: [P1] This now exposes cli-core's `{ accounts, default }` envelope and drops the existing `storage` field from `td user/accounts list --json`. That turns an internal helper swap into a breaking change for scripts consuming the long-standing array payload. Please keep the legacy JSON shape here (or gate the new shape behind a versioned flag) instead of leaking cli-core's list contract directly.

src/commands/user/use.ts
```diff
@@ -1,22 +1,17 @@
...
-    }
+/**
+ * Attach `td accounts use <ref>` (alias `td accounts default <ref>`) via cli-core's
+ * generic `attachAccountUseCommand`, which calls `store.setDefault(ref)` and
+ * owns the success line / `--json` (`{ ok, default }`) / silent-`--ndjson`
+ * output. `setDefault`'s `CliError('ACCOUNT_NOT_FOUND', …)` on a ref miss
+ * propagates to the top-level handler unchanged.
+ */
+export function attachTodoistUserUseCommand(parent: Command, store: TodoistTokenStore): Command {
+    return attachAccountUseCommand<TodoistAccount>(parent, {
```
doistbot, line 13: [P2] Delegating straight to `attachAccountUseCommand` regresses `--quiet`: `td accounts use <ref> --quiet` now prints a success line again, even though the root CLI still documents `--quiet` as suppressing success messages and this command used to honor it. Please preserve the quiet guard in human mode when wiring this attacher.

src/index.ts
```diff
@@ -125,9 +125,10 @@ const commands: Record<string, [string, () => Promise<(p: Command) => void>]> =
         'Manage authentication',
         async () => (await import('./commands/auth/index.js')).registerAuthCommand,
     ],
-    user: [
+    accounts: [
         'Manage stored Todoist accounts (multi-user)',
         async () => (await import('./commands/user/index.js')).registerUserCommand,
+        ['user', 'users'],
```
doistbot: [P2] `['user', 'users']` is now duplicated here and in `src/commands/user/index.ts` (`.aliases(['user', 'users'])`). Those two copies drive different paths (lazy loading/completion vs the real Commander command), so they can silently drift. Please lift the alias list into one shared constant/helper and reuse it in both places.

src/lib/users.ts
```diff
@@ -36,6 +38,24 @@ export function getDefaultUserId(config: Config): string | undefined {
...
+ * `defaultUser`, or — when nothing is pinned — the sole stored account (a lone
+ * account is implicitly default). Mirrors cli-core's `TokenStore.list()`
+ * effective-default rule so the `(default)` marker is identical across
+ * `accounts list`, `accounts current`, `auth status`, and `config view`.
+ *
+ * `getDefaultUserId` (the raw pinned pointer) is still the right call where the
+ * pin itself matters — e.g. `accounts remove` deciding whether it cleared a
+ * pin, or `doctor` diagnosing whether the pin resolves.
+ */
+export function getEffectiveDefaultUserId(config: Config): string | undefined {
```
doistbot: [P2] `getEffectiveDefaultUserId()` doesn't actually mirror `resolveActiveUser()` yet. If `user.defaultUser` points at a missing record, `resolveActiveUser()` falls through to the sole stored account, but this helper returns the stale id immediately. That puts `accounts current` / `auth status` back out of sync on orphaned-default configs. Validate the pinned id against `getStoredUsers(config)` before returning it, or split raw-vs-effective default handling so `config view` can still report stale pointers separately.

src/index.ts
```diff
@@ -247,7 +258,10 @@ if (process.argv[2] === 'completion-server') {
 } else {
     // Find which command (if any) is being invoked — match only known command names
     // to avoid treating option values (e.g. --progress-jsonl /tmp/out) as commands
-    const commandName = process.argv.slice(2).find((a) => !a.startsWith('-') && a in commands)
+    const commandToken = process.argv
```
doistbot, line 262: [P2] Expanding the lazy-loader scan from canonical names to every alias makes pre-command option values ambiguous. With this code, `td --progress-jsonl users task list` will pick `users` as the command token, load the accounts module, and then parse `task list` against placeholders. The alias map is fine once you've identified the command token, but this scan needs to skip option arguments (or otherwise parse the global options first) before checking aliases.

src/commands/user/user.test.ts
```diff
@@ -56,88 +54,112 @@ describe('user command', () => {
...
             expect(lines).toContain('a@b.c')
             expect(lines).toContain('d@e.f')
             expect(lines).toContain('default')
         })
 
-        it('emits one JSON value per line for --ndjson with stored accounts', async () => {
-            mockListStoredUsers.mockResolvedValue([
-                { id: '1', email: 'a@b.c', auth_mode: 'read-write' },
-                { id: '2', email: 'd@e.f', auth_mode: 'read-only' },
+        it('streams one JSON value per console.log for --ndjson with stored accounts', async () => {
```
doistbot: [P3] The test description still says "per console.log", but the assertion was correctly updated to verify the line-delimited payload regardless of how it's batched. Consider updating the description to "per line" to match the new assertion logic.

src/commands/user/user.test.ts
```diff
@@ -56,88 +54,112 @@ describe('user command', () => {
...
-            setup: () => {
-                mockListStoredUsers.mockResolvedValue([])
-            },
-            run: async (extraArgs) => {
-                await createProgram().parseAsync(['node', 'td', 'user', 'list', ...extraArgs])
+        beforeEach(() => {
+            listMock.mockReset().mockResolvedValue([])
+        })
+
+        it.each([['user'], ['users']])(
```
doistbot, line 63: [P2] These alias cases only exercise Commander after `registerUserCommand()` is already loaded. The new behavior in this PR is the lazy alias resolution in `src/index.ts`, so this test won't catch a regression where `td user …` / `td users …` stop loading the `accounts` module from the real entrypoint. Please add one entrypoint-level test for the alias loader instead of relying on the group-level harness alone.

src/commands/user/user.test.ts
```diff
@@ -193,13 +240,33 @@ describe('user command', () => {
...
 
-            await createProgram().parseAsync(['node', 'td', 'user', 'current'])
+            await createProgram().parseAsync(['node', 'td', 'accounts', 'current'])
 
             const out = consoleSpy.mock.calls.flat().join('\n')
             expect(out).toContain('a@b.c')
             expect(out).toContain('default')
         })
 
+        it('marks a lone account as default even with no pinned defaultUser', async () => {
```
doistbot: [P2] This new regression test only covers `accounts current`, but the same effective-default rule was also introduced in `auth status` and `config view`. Their suites still only cover the pinned-default case, so the cross-command consistency this PR is aiming for can drift again without failing tests. Add matching lone-account/no-`defaultUser` cases there too.

src/commands/user/use.ts
doistbot, line 13: [P2] This delegation also changes the missing-`<ref>` path to Commander's built-in required-arg error. `AGENTS.md` says user-facing errors should go through `CliError(...)` so they render consistently in pretty and JSON modes via the global handler. Please keep a thin wrapper here (or otherwise intercept the missing-arg case) so `accounts use` still uses the repo-standard error path.

CODEBASE.md
```diff
@@ -219,9 +220,11 @@ All live in `src/lib/refs.ts`:
 ## Auth & token storage
 
 `@doist/cli-core/auth` owns the keyring, multi-user `TokenStore`, OAuth flow,
-and the `login` / `logout` / `status` / `token view` registrars. todoist-cli
-supplies a `UserRecordStore<TodoistAccount>` adapter (`user-records.ts`) over
-its config file plus a Todoist-specific `validateToken` (`auth-provider.ts`).
+and the `login` / `logout` / `status` / `token view` registrars, plus the
```
doistbot, line 229: [P2] `AGENTS.md` calls out updating `CODEBASE.md` when a new broadly reusable helper is added under `src/lib/`. This PR introduces `getEffectiveDefaultUserId` in `src/lib/users.ts` and reuses it across multiple commands, but the shared-lib catalog still doesn't mention `users.ts`. Please add a short entry there so the repo map stays accurate.

doistbot (commented):
This PR effectively updates the command hierarchy by renaming the user group to `td accounts` and seamlessly retaining legacy aliases. Adopting the generic `cli-core` attachers and unifying the default account logic nicely improves the maintainability of the authentication surface. However, some adjustments are needed to restore the legacy JSON payload shape for `accounts list`, preserve expected `--quiet` and error-handling behaviors during delegation, address edge cases in the alias parsing and effective default logic, and ensure the test coverage and documentation remain fully in sync.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F360&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac8e5d2a0-5685-11f1-96ee-b6602bc23927%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac8e5d2a0-5685-11f1-96ee-b6602bc23927&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/index.ts
scottlovegrove: Fixed in fe4d0ff — extracted `ACCOUNT_COMMAND_ALIASES` into a dependency-free `src/commands/user/aliases.ts` (so `src/index.ts` can import it without eagerly loading the command module), and both the Commander command (`commands/user/index.ts`) and the lazy-loader/completion dispatcher (`src/index.ts`) now consume that single source.

src/lib/users.ts
scottlovegrove: Fixed in fe4d0ff — `getEffectiveDefaultUserId` now validates the pinned id against `getStoredUsers(config)` and falls through to the sole account when the pointer is orphaned, mirroring `resolveActiveUser`. Added a test covering the stale-`defaultUser` case so `current`/`status`/`config view` stay in sync.

src/commands/user/user.test.ts
scottlovegrove: Done in fe4d0ff — description now reads "streams one JSON value per line".

src/commands/user/user.test.ts
scottlovegrove: Added in fe4d0ff — lone-account / no-`defaultUser` effective-default cases now live in the `auth status` (`auth.test.ts`) and `config view` (`config.test.ts`) suites too, so the cross-command consistency is locked.

CODEBASE.md
scottlovegrove: Done in fe4d0ff — added a `users.ts` entry to the CODEBASE.md shared-lib catalog, calling out `getEffectiveDefaultUserId` and the other shared account helpers.

src/commands/user/list.ts
scottlovegrove: Intentional and already settled with the maintainer: keeping cli-core's `{ accounts, default }` envelope and shipping as a minor `feat` (not `BREAKING CHANGE`, no versioned flag). `accounts list --json` is a niche list-users surface whose consumers are agents/scripts that adapt trivially; the human command contract is unchanged, and a major bump would over-signal. Documented under "Note on versioning" in the PR description.

src/commands/user/use.ts
scottlovegrove: Keeping the attacher here by design. `--quiet` isn't honored by *any* cli-core attacher — its `ViewOptions` is only `{ json, ndjson }` and `emitView` has no silent mode — so `td auth login`/`logout`/`status` already ignore `--quiet` in this CLI. Adopting `attachAccountUseCommand` makes `accounts use` consistent with them; the old hand-rolled `use` was the lone exception. The right fix is adding quiet support to cli-core's `emitView` (fixes all attacher-backed commands at once), which we'll track as a separate cli-core change rather than re-diverging just `use` here.

scottlovegrove: Same call — keeping the attacher. The missing-`<ref>` path uses Commander's required-arg error because `attachAccountUseCommand` declares `<ref>` as a required positional and doesn't know todoist's `CliError` convention; this is consistent with the other cli-core-attached commands. Routing it through `CliError` (JSON-rendered) would be a cli-core enhancement (optional positional + typed throw), tracked alongside the `--quiet`/`emitView` gap above rather than worked around locally for this one command.

src/index.ts
scottlovegrove: Deferring — this is a pre-existing class, not introduced here: the command-token scan already mis-detects when an option *value* collides with a command name (`td --progress-jsonl task list` picks `task` today). The alias change only adds `user`/`users` to that colliding set. The proper fix — parse global options before command detection (or skip option arguments in the scan) — is a standalone change to load-bearing dispatch code, so we're tracking it separately rather than expanding this PR.

src/commands/user/user.test.ts
scottlovegrove: Deferring, paired with the dispatch fix above. `src/index.ts` parses argv at import time, so there's no entrypoint unit harness today — a real entrypoint-loader test needs the resolution logic extracted into a testable helper (the same refactor the option-value fix wants). The alias→canonical resolution is covered at the unit level (completion.test.ts) and the group-level command behaviour is covered (user.test.ts), so the gap is specifically the entrypoint wiring; tracking it with the dispatch refactor.

scottlovegrove:
@doistbot /review

src/index.ts
```diff
@@ -247,7 +259,10 @@ if (process.argv[2] === 'completion-server') {
...
```
doistbot, line 262: [P2] Scanning `process.argv` against `commandAliases` without skipping option arguments introduces a real regression for the new aliases. For example, `td --progress-jsonl users task list` now picks `users` as the command token, loads `accounts`, and then errors, whereas this used to run `task list` with `users` as the option value. Parse the global options first, or at least skip argv entries consumed by options that take a value before checking aliases.

src/commands/user/current.ts
```diff
@@ -1,50 +1,75 @@
...
+        store,
+        description: 'Show the active account (resolved from --user, default, or single login)',
+        renderText: ({ account, isDefault }) => {
+            const marker = isDefault ? chalk.green(' (default)') : ''
+            return `${account.email} ${chalk.dim(`(id:${account.id})`)}${marker}`
+        },
+        renderJson: ({ account, isDefault }) => ({
+            id: account.id,
+            email: account.email,
+            source: 'secure-store',
```
doistbot: [P2] This hardcodes stored accounts as `source: 'secure-store'`, but stored users can still be backed by the config-file fallback token (`users[].api_token`). The old implementation surfaced `config-file` in that case via `resolveActiveUser()`. As written, `td accounts current --json` now lies about the credential source on keyring-fallback installs. Please derive `source` from the resolved auth path instead of hardcoding it.

src/commands/user/use.ts
doistbot, line 13: [P2] Delegating straight to the generic attacher regresses the missing-`<ref>` path: Commander handles it before `CliError`, so `td accounts use --json` no longer goes through the repo’s normal JSON/pretty error formatter. Please keep a thin local arg check before delegating here (and in `remove`) so user-facing errors stay on the standard error path.

src/lib/users.ts
```diff
@@ -36,6 +38,29 @@ export function getDefaultUserId(config: Config): string | undefined {
...
```
doistbot: [P2] `getEffectiveDefaultUserId()` now reimplements the same "pinned default else sole stored account" selection rule that `resolveActiveUser()` already carries in `auth.ts`. This PR already had to fix one drift between those paths. Please extract the default-account resolution into one shared helper and have both callers use it instead of maintaining a second copy here.

src/commands/user/current.ts
```diff
@@ -1,50 +1,75 @@
...
+    store: TodoistTokenStore,
+): Command {
+    return attachAccountCurrentCommand<TodoistAccount>(parent, {
+        store,
+        description: 'Show the active account (resolved from --user, default, or single login)',
+        renderText: ({ account, isDefault }) => {
+            const marker = isDefault ? chalk.green(' (default)') : ''
+            return `${account.email} ${chalk.dim(`(id:${account.id})`)}${marker}`
+        },
+        renderJson: ({ account, isDefault }) => ({
```
doistbot, line 32: [P2] `renderJson()` is duplicating the `TodoistAccount` → CLI JSON projection that `projectAccount()` now defines in `list.ts` (`id`, `email`, `isDefault`, `authMode`, `authScope`, `authFlags`), with `source` as the only extra field. Please pull the shared projection into one helper and extend it here so `accounts list` and `accounts current` don't drift on future field changes.

src/commands/user/use.ts
doistbot, line 13: [P2] Moving `accounts use` onto cli-core's `ACCOUNT_NOT_FOUND` path introduces a second public error-code family for the same account-selection miss. `resolveActiveUser`, `auth logout`, and `auth token view` still surface `USER_NOT_FOUND`, so JSON consumers now have to special-case both. Please normalize on one code family here (or migrate the remaining local helpers in the same change) instead of exposing both.

src/lib/users.ts
```diff
@@ -36,6 +38,29 @@ export function getDefaultUserId(config: Config): string | undefined {
...
+    const users = getStoredUsers(config)
+    const pinned = getDefaultUserId(config)
+    // A pinned pointer only counts when it resolves to a stored account —
+    // mirrors `resolveActiveUser`, which ignores an orphaned `defaultUser` and
+    // falls through to the sole stored account. Returning a stale id here would
+    // put `accounts current` / `auth status` back out of sync on a config whose
+    // `defaultUser` points at a removed account.
+    if (pinned && users.some((u) => u.id === pinned)) return pinned
```
doistbot: [P2] The logic handling a stale/orphaned `defaultUser` pointer is untested. While the test suites cover the "no pinned defaultUser" scenario, they do not verify that an orphaned pin correctly falls through to the single available account. Please add a test for this edge case to prevent the fallback behavior from regressing.

src/commands/user/user.test.ts
```diff
@@ -56,88 +56,112 @@ describe('user command', () => {
...
-            setup: () => {
-                mockListStoredUsers.mockResolvedValue([])
-            },
-            run: async (extraArgs) => {
-                await createProgram().parseAsync(['node', 'td', 'user', 'list', ...extraArgs])
+        beforeEach(() => {
+            listMock.mockReset().mockResolvedValue([])
+        })
+
+        it.each([['user'], ['users']])(
```
doistbot, line 63: [P2] These alias cases only exercise Commander after `registerUserCommand()` has already been loaded. The new behavior in this PR is the lazy alias resolution in `src/index.ts`, so a regression there wouldn’t fail this suite. Add one entrypoint-level test that runs `td user …` / `td users …` through the lazy dispatcher.

src/commands/user/user.test.ts
```diff
@@ -172,35 +196,78 @@ describe('user command', () => {
...
 
-            await createProgram().parseAsync(['node', 'td', 'user', 'current'])
+            await createProgram().parseAsync(['node', 'td', 'accounts', 'current'])
 
             const out = consoleSpy.mock.calls.flat().join('\n')
             expect(out).toContain('a@b.c')
             expect(out).toContain('default')
         })
 
+        it('emits the stored account shape for --json', async () => {
```
doistbot, line 251: [P2] `accounts current` gained `--ndjson` here, but this suite still only exercises pretty and `--json`. The fallback path in `current.ts` formats NDJSON itself for env/legacy sources, so add one `--ndjson` case there to lock down the line-delimited payload.

src/commands/user/user.test.ts
```diff
@@ -209,42 +276,71 @@ describe('user command', () => {
...
+            clearMock.mockReset()
             lastClearMock.mockReset().mockReturnValue({ storage: 'secure-store' })
         })
 
-        it('removes the user by id and clears default', async () => {
-            mockReadConfig.mockResolvedValue({
-                config_version: 2,
-                user: { defaultUser: '111' },
-                users: [{ id: '111', email: 'a@b.c' }],
+        it('removes the resolved account and notes the cleared default', async () => {
```
doistbot, line 341: [P2] `accounts remove` now exposes `--json` and silent `--ndjson`, but this block still only covers human/error paths. Please add one `--json` assertion for the `{ ok, removed }` payload and one `--ndjson` assertion for the empty output so the new contract is tested in this repo instead of only through cli-core.

src/lib/skills/content.ts
```diff
@@ -68,17 +68,19 @@ Tokens are stored in the OS credential manager when available, with fallback to
...
-td user current                     # show the active account
-td user remove <id|email>           # delete an account (and its token)
-td --user <id|email> task list      # one-off override for any command
-td auth logout --user <id|email>    # log out a specific account
+td auth login                          # adds the account; first one becomes default
+td accounts list                       # all stored accounts (with default marker)
+td accounts list --json                # { accounts: [...], default } envelope; --ndjson streams one account per line
+td accounts use <id|email>             # set the default account (alias: td accounts default; --json/--ndjson supported)
+td accounts current                    # show the active account
+td accounts remove <id|email>          # delete an account (and its token)
```
doistbot: [P2] `AGENTS.md` requires that whenever flags are added, `SKILL_CONTENT` must be updated to match. The PR description notes that `accounts remove` gains `--json` and `--ndjson` support, and `accounts current` gains `--ndjson` support. Please update the examples for these commands to reflect their new machine-output capabilities, just as you did for `accounts use`.

doistbot (commented):
This PR brings a great refactor that renames the `user` command group to `accounts` and successfully delegates its subcommands to the generic `@doist/cli-core` attachers. Moving these commands to the `TokenStore` contract greatly simplifies the local implementation and cleanly expands machine-readable output support. A few areas require adjustments, specifically around refining the entrypoint alias parsing to ignore option values, ensuring fallback credential sources are reported accurately, consolidating duplicated logic and error codes, updating the agent skill content for the new flags, and rounding out test coverage for the new outputs and edge cases.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F360&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4a4bc5c0-56c2-11f1-94eb-3d730e17dde5%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4a4bc5c0-56c2-11f1-94eb-3d730e17dde5&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/users.ts
scottlovegrove: Fixed in e4021bf — extracted `getEffectiveDefaultUser(config)` in `users.ts` as the single default-selection rule. `resolveActiveUser` (which had the duplicate `pickDefault`) and `getEffectiveDefaultUserId` now both delegate to it, so the resolver and the `(default)` marker share one implementation.

src/commands/user/current.ts
scottlovegrove: Fixed in e4021bf — pulled the projection into `src/commands/user/project-account.ts` (`projectAccount`). `accounts list` uses it directly; `accounts current` spreads it and adds `source`, so the shared fields can't drift.

src/lib/users.ts
scottlovegrove: Added in e4021bf — new `src/lib/users.test.ts` unit-tests `getEffectiveDefaultUser`/`Id` directly, including the orphaned-pin fall-through to the sole stored account (plus pinned, lone-account, multi-no-pin, and empty cases).

src/commands/user/user.test.ts
scottlovegrove: Added in e4021bf — an `accounts current --ndjson` case covering the env fallback (asserts the single line-delimited `{ id: null, source: "env", … }` payload that the `onNotAuthenticated` branch formats).

src/commands/user/user.test.ts
scottlovegrove: Added in e4021bf — `accounts remove --json` asserts `{ ok: true, removed: <id> }` and `--ndjson` asserts empty stdout, so the new contract is covered here rather than only in cli-core.

src/lib/skills/content.ts
scottlovegrove: Done in e4021bf — updated the `content.ts` examples so `accounts current` and `accounts remove` note `--json`/`--ndjson` support, and re-ran `sync:skill` (SKILL.md regenerated, check:skill-sync green).

src/commands/user/current.ts
scottlovegrove: Fixed in 8475f32 — `accounts current --json` now reports the real source. The store adapter's `activeAccount` annotates the resolved account with `secure-store` vs `config-file` (based on whether the stored user has a plaintext `api_token` fallback), and `current` surfaces that instead of the hardcoded value. `activeAccount` is only consumed by `current`, so this doesn't leak into `accounts list`.

src/commands/user/use.ts
scottlovegrove: Fixed in 8475f32 — normalized onto cli-core's `ACCOUNT_NOT_FOUND`. `UserNotFoundError` (the `--user` resolver, `auth logout`/`token view`) now throws `ACCOUNT_NOT_FOUND` instead of `USER_NOT_FOUND`, so JSON consumers see one code family for every account-selection miss (matching the cli-core attachers). Updated the two tests that asserted the old code.

scottlovegrove: Keeping cli-core's behavior here. The missing-`<ref>` message is Commander's required-arg validation (cli-core's attacher declares `<ref>` as a required positional), which fires during parse before any action — it's not a `CliError` omission, and it's consistent with every other `<arg>` command in the CLI. Routing it through `CliError` would mean making `<ref>` optional and hand-rolling the arg check, i.e. dropping the attacher for `use`/`remove`. Per the "follow cli-core's standard" call, we're staying on the attacher.

src/index.ts
scottlovegrove: Tracked separately rather than fixed in this PR. Fair that the alias-widening extends the colliding token set (`user`/`users`), but the underlying scan already mis-detects option values that match a canonical command name (`td --progress-jsonl task list` today), so the proper fix is parsing global options before command detection — a standalone change to the lazy dispatcher, paired with the entrypoint test below.

src/commands/user/user.test.ts
scottlovegrove: Tracked with the dispatcher change above. `src/index.ts` parses argv at import time, so an entrypoint-level test needs the resolution logic extracted into a testable helper (the same refactor the option-value fix wants); they'll land together. The alias→canonical resolution is unit-tested (completion.test.ts) and the group-level command behavior is covered (user.test.ts) in the meantime.

rmartins90 (commented):
✅

PR #373, opened 2026-05-29
feat(project): add `project share` to invite a collaborator

### LINKED ISSUE #372: feat: td project share — add command to invite a collaborator to a project
## Problem

There's no CLI command to share a project with another collaborator. `td project collaborators` lists existing collaborators and `td project join` lets you join a project you've been invited to, but there's no way to *invite* someone programmatically.

## Impact

This blocks clean end-to-end agent automation in DoistOS. The `/cleanup-handbook` skill creates a Todoist project for audit findings but can't share it with the relevant person — the current workaround is prefixing task titles with `[PersonName]` as a fallback, which is a naming convention hack rather than real sharing.

## Requested

A command along the lines of:

```
td project share <project-ref> <email-or-user-ref> [--role viewer|editor]
```

This would mirror what's available in the Todoist UI and REST API (`POST /projects/{id}/collaborators`).

## Context

Filed from DoistOS — both `td` and `tw` feature gaps were reviewed as part of scoping clean agent automation workflows.

scottlovegrove:
## What

Adds `td project share <project-ref> <email>` to invite a collaborator to a project — closes the gap alongside the existing `project join` and `project collaborators` commands.

Closes #372.

```
td project share <project-ref> <email> [--role guest|member|admin] [--message <msg>] [--auto-invite] [--json] [--dry-run]
```

## Behaviour

- **Personal / shared projects** — shares via the `share_project` sync command. `--role` / `--auto-invite` don't apply and are ignored with a warning.
- **Workspace projects** — `--role guest|member|admin` (default `member`) maps to the workspace role. If the email isn't already a workspace member, the command errors with a hint, unless `--auto-invite` is passed — which invites them to the workspace first (via `inviteWorkspaceUsers`) and then shares.
- `--message` attaches an optional invitation message.
- `--dry-run` previews (including whether an auto-invite would fire) without mutating; `--json` outputs the result.

## Notes

- Inviting is only available through the Sync API `share_project` command (there's no REST endpoint), so this adds a `shareProject` sync wrapper in `lib/api/projects-sync.ts`.
- Role vocabulary is the real workspace roles (`guest`/`member`/`admin`), not the `viewer`/`editor` in the original issue text.
- Skill content + `SKILL.md` updated and in sync.
- Share tests live in a new `project.share.test.ts`, and the shared `createProjectProgram` test helper was extracted for reuse.

## Verification

- `npm run type-check` ✅
- `npm run check` (lint + format) ✅
- `npm run check:skill-sync` ✅
- `npm test` — full suite passes (1672)

### DIFF
--- a/src/commands/project/index.ts
+++ b/src/commands/project/index.ts
@@ -20,6 +20,7 @@ import { moveProject } from './move.js'
 import { showPermissions } from './permissions.js'
 import { showProjectProgress } from './progress.js'
 import { reorderProject } from './reorder.js'
+import { shareProject } from './share.js'
 import { unarchiveProject } from './unarchive.js'
 import { updateProject } from './update.js'
 import { viewProject } from './view.js'
@@ -255,6 +256,31 @@ Examples:
         .option('--dry-run', 'Preview what would happen without executing')
         .action((id, options) => {
             return joinProjectCmd(id, options)
+        })
+
+    project
+        .command('share <ref> <email>')
+        .description('Invite a collaborator to a project')
+        .option(
+            '--role <role>',
+            'Workspace role: guest, member, or admin (workspace projects only; default: member)',
+        )
+        .option('--message <msg>', 'Optional invitation message')
+        .option(
+            '--auto-invite',
+            'If the email is not yet a workspace member, invite them to the workspace first (workspace projects only)',
+        )
+        .option('--json', 'Output the result as JSON')
+        .option('--dry-run', 'Preview what would happen without executing')
+        .addHelpText(
+            'after',
+            `
+Examples:
+  td project share "Roadmap" alice@example.com
+  td project share "Team Plan" bob@example.com --role guest --auto-invite`,
+        )
+        .action((ref, email, options) => {
+            return shareProject(ref, email, options)
         })
 
     const progressCmd = project
--- a/src/commands/project/index.ts
+++ b/src/commands/project/index.ts
@@ -260,12 +260,17 @@ Examples:
         })
 
     const shareCmd = project
-        .command('share [ref] [email]')
+        .command('share [project] [email]')
         .description('Invite a collaborator to a project')
-        .option('--project <ref>', 'Project name or id:xxx (alias for the positional ref)')
-        .option(
-            '--role <role>',
-            'Workspace role: guest, member, or admin (workspace projects only; default: member)',
+        .option('--project <ref>', 'Project name or id:xxx (alias for the positional project)')
+        .addOption(
+            withCaseInsensitiveChoices(
+                new Option(
+                    '--role <role>',
+                    'Workspace role (workspace projects only; default: member)',
+                ),
+                ['guest', 'member', 'admin'],
+            ),
         )
         .option('--message <msg>', 'Optional invitation message')
         .option(
@@ -282,29 +287,37 @@ Examples:
   td project share --project "Roadmap" alice@example.com
   td project share "Team Plan" bob@example.com --role guest --auto-invite`,
         )
-        .action((ref, email, options) => {
-            // Project can come from the positional ref or --project, but not both.
+        .action((projectArg, emailArg, options) => {
+            // The project is provided either positionally or via --project, never both.
+            // In --project mode the sole positional is the collaborator email.
             let projectRef: string | undefined
-            let collaboratorEmail: string | undefined
-            if (options.project) {
-                if (ref && email) {
+            let email: string | undefined
+            if (options.project !== undefined) {
+                if (emailArg !== undefined) {
                     throw new CliError(
                         'CONFLICTING_OPTIONS',
-                        'Cannot specify project both as argument and --project flag',
+                        'Cannot specify the project both positionally and via --project.',
                     )
                 }
                 projectRef = options.project
-                collaboratorEmail = ref
+                email = projectArg
             } else {
-                projectRef = ref
-                collaboratorEmail = email
+                projectRef = projectArg
+                email = emailArg
             }
 
-            if (!projectRef || !collaboratorEmail) {
+            if (projectRef === undefined) {
                 shareCmd.help()
                 return
             }
-            return shareProject(projectRef, collaboratorEmail, options)
+            if (email === undefined) {
+                throw new CliError(
+                    'MISSING_EMAIL',
+                    'Specify the email of the collaborator to invite.',
+                    ['Usage: td project share <project> <email>'],
+                )
+            }
+            return shareProject(projectRef, email, options)
         })
 
     const progressCmd = project
--- a/src/commands/project/index.ts
+++ b/src/commands/project/index.ts
@@ -1,6 +1,7 @@
 import { Command, Option } from 'commander'
 import { withCaseInsensitiveChoices } from '../../lib/completion.js'
 import { CURSOR_DESCRIPTION } from '../../lib/constants.js'
+import { CliError } from '../../lib/errors.js'
 import { parseOrderArg } from '../../lib/order.js'
 import { showProjectActivityStats } from './activity-stats.js'
 import { analyzeHealth } from './analyze-health.js'
@@ -258,9 +259,10 @@ Examples:
             return joinProjectCmd(id, options)
         })
 
-    project
-        .command('share <ref> <email>')
+    const shareCmd = project
+        .command('share [ref] [email]')
         .description('Invite a collaborator to a project')
+        .option('--project <ref>', 'Project name or id:xxx (alias for the positional ref)')
         .option(
             '--role <role>',
             'Workspace role: guest, member, or admin (workspace projects only; default: member)',
@@ -277,10 +279,32 @@ Examples:
             `
 Examples:
   td project share "Roadmap" alice@example.com
+  td project share --project "Roadmap" alice@example.com
   td project share "Team Plan" bob@example.com --role guest --auto-invite`,
         )
         .action((ref, email, options) => {
-            return shareProject(ref, email, options)
+            // Project can come from the positional ref or --project, but not both.
+            let projectRef: string | undefined
+            let collaboratorEmail: string | undefined
+            if (options.project) {
+                if (ref && email) {
+                    throw new CliError(
+                        'CONFLICTING_OPTIONS',
+                        'Cannot specify project both as argument and --project flag',
+                    )
+                }
+                projectRef = options.project
+                collaboratorEmail = ref
+            } else {
+                projectRef = ref
+                collaboratorEmail = email
+            }
+
+            if (!projectRef || !collaboratorEmail) {
+                shareCmd.help()
+                return
+            }
+            return shareProject(projectRef, collaboratorEmail, options)
         })
 
     const progressCmd = project
--- a/src/commands/project/project.share.test.ts
+++ b/src/commands/project/project.share.test.ts
@@ -287,4 +287,102 @@ describe('project share', () => {
             autoInvited: false,
         })
     })
+
+    it('finds a member across paginated workspace-user pages', async () => {
+        const program = createProgram()
+
+        mockApi.getProject.mockResolvedValue({
+            id: 'proj-1',
+            name: 'Park Ops',
+            workspaceId: 'ws-1',
+        })
+        mockApi.getWorkspaceUsers
+            .mockResolvedValueOnce({
+                workspaceUsers: [
+                    {
+                        userId: 'user-1',
+                        fullName: 'John Hammond',
+                        userEmail: 'john@ingen.com',
+                        role: 'ADMIN',
+                    },
+                ],
+                hasMore: true,
+                nextCursor: 'page-2',
+            })
+            .mockResolvedValueOnce({
+                workspaceUsers: [
+                    {
+                        userId: 'user-2',
+                        fullName: 'Alan Grant',
+                        userEmail: 'alan@ingen.com',
+                        role: 'MEMBER',
+                    },
+                ],
+                hasMore: false,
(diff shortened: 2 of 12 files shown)


src/commands/project/share.ts
```diff
@@ -0,0 +1,152 @@
...
+            'Auto-invite': !isMember
+                ? options.autoInvite
+                    ? 'yes'
+                    : 'required (not a member)'
+                : undefined,
+        })
+        return
+    }
+
+    if (!isMember && !options.autoInvite) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Move this `NOT_WORKSPACE_MEMBER` validation block above the `if (options.dryRun)` check.

A dry run should reflect the success or failure of the actual command. Placing validation after `dryRun` means `td project share <project> <email> --dry-run` will exit successfully with an "Auto-invite: required" preview message, even though the actual command would fail with an error.

Moving it up also lets you simplify the dry-run output to `'Auto-invite': autoInvited ? 'yes' : undefined`.

src/commands/project/index.ts
```diff
@@ -257,6 +258,31 @@ Examples:
             return joinProjectCmd(id, options)
         })
 
+    project
+        .command('share <ref> <email>')
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This new subcommand takes a project ref positionally, but it doesn't also expose the standard `--project <ref>` alias or reject the positional+flag combination. Right now `td project share --project "Roadmap" alice@example.com` fails as an unknown option, which breaks the established contract for project-scoped commands. Add `--project <ref>` and the same conflict check used by the other positional-or-flag commands.

src/commands/project/project.share.test.ts
```diff
@@ -0,0 +1,290 @@
...
+                    projectId: 'proj-1',
+                    email: 'alan@ingen.com',
+                    message: undefined,
+                    workspaceRole: 'MEMBER',
+                },
+            }),
+        )
+        expect(consoleSpy).toHaveBeenCalledWith('Shared: Park Ops with alan@ingen.com')
+        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining('Role: MEMBER'))
+    })
```
doistbot, line 138: <big><kbd>🟡 P2</kbd></big>  Consider adding a test to verify pagination for `isWorkspaceMember`. The implementation introduces a `while(true)` loop to handle workspaces with over 200 users, but the current tests only cover the single-page scenario (`hasMore: false`). Mocking a response with `hasMore: true` and a `nextCursor` would ensure the loop properly advances the cursor and successfully finds users on subsequent pages.

src/lib/skills/content.ts
```diff
@@ -174,6 +174,8 @@ td project archive "Roadmap"
 td project unarchive "Roadmap"
 td project move "Roadmap" --to-workspace "Acme" --folder "Engineering" --visibility team --yes
 td project join id:abc123
+td project share "Roadmap" alice@example.com
```
doistbot, line 177: <big><kbd>🟡 P2</kbd></big>  This SKILL_CONTENT update still doesn’t fully match the new command surface. `project share` now adds `--message`, `--json`, and `--dry-run` (alongside the workspace-only role behavior), but the skill diff only adds bare examples plus one `--role/--auto-invite` case. Per the repo guidance, the agent reference needs to be updated when flags/options are added, otherwise installed skills won’t expose the actual CLI behavior.

doistbot (commented):
Thanks scottlovegrove for your contribution 🤗. The new `project share` command is a great addition that cleanly handles the nuances between sharing personal and workspace projects.

Few things worth tightening:
- Support the standard `--project <ref>` alias for consistency, and move the workspace member validation above the `--dry-run` check so dry runs accurately reflect failure states.
- Expand the `SKILL_CONTENT` update to cover the newly added `--message`, `--json`, and `--dry-run` flags.
- Add a test with a mocked `hasMore: true` response to exercise the workspace member pagination loop.

<details>

- **[P3] src/commands/project/share.ts:32**: Rather than inferring this type using `Awaited<ReturnType<typeof getApi>>`, you can explicitly type it by importing `type { TodoistApi }` from `@doist/todoist-sdk`.
- **[P3] src/commands/project/index.ts:264**: `--role` is introduced here as a free-form string even though this file already uses `withCaseInsensitiveChoices(...)` for other enum-like options. That forces `share.ts` to hardcode and revalidate the same role list separately, so help text, shell completions, and runtime validation can drift. Build this option from the shared workspace role list and Commander choice helper instead of validating it ad hoc later.
- **[P3] src/commands/project/share.ts:31**: This adds another copy of the `getWorkspaceUsers` cursor loop that already exists in `src/commands/workspace/helpers.ts` (`findUser` / `resolveWorkspaceUserRef`). Pulling the exact-email membership check into that shared helper would keep workspace-user lookup behavior in one place if the paging contract or matching rules ever change.
- **[P3] src/commands/project/share.ts:40**: Per `AGENTS.md`, API paginated responses should always be destructured. Consider destructuring `{ workspaceUsers, hasMore, nextCursor }` directly from the response object.
- **[P3] src/commands/project/share.ts:79**: `AGENTS.md` requires mutating commands to output JSON using `formatJson` from `src/lib/output.ts`. Even though there isn't a strict `EntityType` for a share result, you can use `formatJson({ ... })` instead of `JSON.stringify(...)` to maintain consistency with the core CLI formatter. (This also applies to the JSON output on line 136).

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F373&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab1a25370-5b76-11f1-90d8-9bdeb2be4577%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab1a25370-5b76-11f1-90d8-9bdeb2be4577&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/project/share.ts
scottlovegrove: Done — moved the `NOT_WORKSPACE_MEMBER` validation above the `--dry-run` early return so a dry run now reflects the real failure, and simplified the preview to `'Auto-invite': autoInvited ? 'yes' : undefined`. (b8778ea)

src/commands/project/index.ts
scottlovegrove: Added `--project <ref>` plus the standard positional-or-flag conflict check (mirrors `template`/`section`/`reminder`). Both forms covered by tests. (b8778ea)

src/commands/project/project.share.test.ts
scottlovegrove: Added a test that mocks a first page with `hasMore: true` + `nextCursor` and finds the member on page 2 — asserts the second `getWorkspaceUsers` call advances the cursor. Also switched the lookup to reuse the shared `findUser` paginator from `workspace/helpers.ts`. (b8778ea)

src/lib/skills/content.ts
scottlovegrove: Expanded the SKILL_CONTENT examples to cover `--project`, `--message`, `--json`, and `--dry-run`, and re-ran `npm run sync:skill`. (b8778ea)

scottlovegrove:
@doistbot /review

src/commands/project/index.ts
```diff
@@ -257,6 +259,54 @@ Examples:
             return joinProjectCmd(id, options)
         })
 
+    const shareCmd = project
+        .command('share [ref] [email]')
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `share [ref] [email]` becomes ambiguous once `--project` is added. For example, `td project share id:proj-1 --project id:proj-2` leaves only one positional, and the action code ends up reinterpreting that positional project ref as the collaborator email instead of throwing a conflict/missing-email error. The parser needs to be restructured so `--project` mode doesn't overload the first positional as both project and email.

doistbot (commented):
Thanks scottlovegrove for your contribution 😇. The new `project share` flow looks sound overall, and the test coverage effectively exercises the new workspace role logic and pagination loops.

Few things worth tightening:
- Restructure the argument parsing so that using the `--project` flag doesn't cause the first positional argument to be misidentified as the collaborator email.
- Add a test for `--dry-run` on workspace projects when the email is not already a member and `--auto-invite` is omitted to lock in the early validation logic.

<details>

- **[P3] src/commands/project/test-helpers.ts:1**: This file is test-only, but because it lives under `src/commands/` and is not a `*.test.ts` file, `tsconfig.build.json` will emit it into `dist/` and pull `@doist/cli-core/testing` into the production build. Move it under `src/test-support/` or exclude/rename it so test helpers stay out of release artifacts.
- **[P3] src/commands/project/index.ts:266**: Consider using `withCaseInsensitiveChoices` and `Option` (both already imported in this file) for the `--role` option. This enables shell autocompletion for the valid choices and delegates validation to Commander, which allows you to remove the manual `INVALID_ROLE` check in `share.ts`. ```typescript .addOption( withCaseInsensitiveChoices( new Option('--role <role>', 'Workspace role: guest, member, or admin (workspace projects only; default: member)'), ['guest', 'member', 'admin'] ) ) ```
- **[P3] src/commands/project/project.share.test.ts:31**: Per the test conventions in `CODEBASE.md`, mock entities should come from `src/test-support/fixtures.ts` rather than being hand-built inline. This new suite starts constructing project/workspace-user objects directly here and repeats that pattern throughout the file, which makes it easier for test shapes to drift from the shared fixtures. Please switch these mocks to fixture-based objects, adding missing fixture helpers if needed.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F373&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A48168b80-5b7e-11f1-91ca-462d153a19be%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A48168b80-5b7e-11f1-91ca-462d153a19be&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/project/project.share.test.ts
```diff
@@ -0,0 +1,388 @@
...
+                '--role',
+                'editor',
+            ]),
+        ).rejects.toMatchObject({ code: 'INVALID_ROLE' })
+
+        expect(mockApi.sync).not.toHaveBeenCalled()
+        expect(mockApi.inviteWorkspaceUsers).not.toHaveBeenCalled()
+    })
+
+    it('--dry-run previews without mutating', async () => {
```
doistbot, line 252: <big><kbd>🟡 P2</kbd></big>  Please add a companion test for `--dry-run` on a workspace project when the email is not already a member and `--auto-invite` is not set. This PR explicitly moved that validation above the dry-run early return, but the suite still only covers the plain error path and the successful dry-run path. Without a test here, that regression can come back unnoticed.

src/commands/project/index.ts
scottlovegrove: Reworked the parser into two explicit modes (d75581f): with `--project`, the project comes from the flag and the single positional is the email; without it, the form is `<project> <email>`. Supplying the project both ways now throws `CONFLICTING_OPTIONS`, and a missing email throws `MISSING_EMAIL` instead of being silently inferred. The lone positional in `--project` mode is unambiguously the email (consistent with `--task`/`--workspace` context flags elsewhere).

src/commands/project/project.share.test.ts
scottlovegrove: Added a test asserting that `--dry-run` on a workspace project with a non-member email and no `--auto-invite` throws `NOT_WORKSPACE_MEMBER` (locks in the validation-before-dry-run ordering). (d75581f)

PR #376, opened 2026-06-02
feat(auth): add `--no-browser-open` to `auth login`

scottlovegrove:
## What

Adds a `--no-browser-open` flag to `td auth login`. With it, the OAuth flow prints the authorization URL for manual copy-paste instead of automatically opening a browser.

```
td auth login --no-browser-open
```

## Why

Useful on headless or remote hosts, in SSH sessions, or any time you'd rather not have the desktop browser hijacked. cli-core already auto-skips the spawn on detected headless/WSL environments; this flag forces the skip everywhere.

## How

cli-core surfaces the authorize URL before its `openBrowser` hook fires, so skipping the spawn still leaves a usable login — the printed URL is the path through. cli-core's `openBrowser` hook isn't handed the parsed flags, so the option is read off the command at call time (commander negatable `--no-browser-open` → `browserOpen`, default `true`).

## Tests

- Added two tests: opens the browser by default, skips when the flag is set.
- SKILL content updated and regenerated via `sync:skill`.
- `type-check`, `check` (oxlint + oxfmt), and the auth login suite all pass.

### DIFF
--- a/src/commands/auth/login.test.ts
+++ b/src/commands/auth/login.test.ts
@@ -19,6 +19,9 @@ vi.mock('../../lib/api/core.js', () => ({
 
 vi.mock('chalk')
 
+const openMock = vi.fn()
+vi.mock('open', () => ({ default: (url: string) => openMock(url) }))
+
 // Capture (but don't execute) the options handed to cli-core's
 // `attachLoginCommand` so tests can invoke the Todoist-local callbacks
 // (`resolveScopes`, `onSuccess`) directly — cli-core's own tests don't cover
@@ -48,6 +51,7 @@ type AttachOptions = {
         view: { json: boolean; ndjson: boolean }
         flags: Record<string, unknown>
     }) => void | Promise<void>
+    openBrowser: (url: string) => Promise<void>
     store: { getLastStorageResult: () => unknown }
 }
 
@@ -55,6 +59,20 @@ function attachAndCapture(): AttachOptions {
     capturedAttachOptions.length = 0
     createTestProgram((program) => attachTodoistLoginCommand(program, createTodoistTokenStore()))
     return capturedAttachOptions[capturedAttachOptions.length - 1].options as AttachOptions
+}
+
+function attachAndCaptureLogin(): { options: AttachOptions; login: Command } {
+    capturedAttachOptions.length = 0
+    const captured: { login?: Command } = {}
+    createTestProgram((program) => {
+        captured.login = attachTodoistLoginCommand(program, createTodoistTokenStore())
+    })
+    const login = captured.login
+    if (!login) throw new Error('login command was not attached')
+    return {
+        options: capturedAttachOptions[capturedAttachOptions.length - 1].options as AttachOptions,
+        login,
+    }
 }
 
 const ACCOUNT = {
@@ -139,5 +157,32 @@ describe('attachTodoistLoginCommand: onSuccess output formatting', () => {
         expect(consoleSpy).toHaveBeenCalledTimes(1)
         expect(() => JSON.parse(consoleSpy.mock.calls[0][0] as string)).not.toThrow()
         expect(errorSpy.mock.calls.map((c: unknown[]) => c.join(' ')).join('\n')).toContain(warning)
+    })
+})
+
+describe('attachTodoistLoginCommand: --no-browser-open', () => {
+    beforeEach(() => {
+        openMock.mockClear()
+    })
+
+    afterEach(() => {
+        capturedAttachOptions.length = 0
+    })
+
+    it('opens the browser by default', async () => {
+        const { options } = attachAndCaptureLogin()
+        await options.openBrowser('https://todoist.com/oauth/authorize')
+        expect(openMock).toHaveBeenCalledWith('https://todoist.com/oauth/authorize')
+    })
+
+    it('skips opening the browser when --no-browser-open is set', async () => {
+        const { options, login } = attachAndCaptureLogin()
+        login.setOptionValue('browserOpen', false)
+
+        await options.openBrowser('https://todoist.com/oauth/authorize')
+
+        // cli-core still surfaces the URL itself; the local hook just declines
+        // to spawn a browser, leaving the printed URL for manual copy-paste.
+        expect(openMock).not.toHaveBeenCalled()
     })
 })
--- a/src/commands/auth/login.test.ts
+++ b/src/commands/auth/login.test.ts
@@ -61,17 +61,14 @@ function attachAndCapture(): AttachOptions {
     return capturedAttachOptions[capturedAttachOptions.length - 1].options as AttachOptions
 }
 
-function attachAndCaptureLogin(): { options: AttachOptions; login: Command } {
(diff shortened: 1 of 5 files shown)


src/commands/auth/login.ts
```diff
@@ -72,8 +82,14 @@ export function attachTodoistLoginCommand(auth: Command, store: TodoistTokenStor
         },
     })
 
+    loginCommand = login
+
     return login
         .description('Authenticate with Todoist via OAuth')
+        .option(
+            '--no-browser-open',
```
doistbot, line 92: <big><kbd>🟡 P2</kbd></big>  This adds a new login mode, but the repo's reauth helpers still only rebuild `--read-only` and `--additional-scopes`. After someone signs in with `--no-browser-open`, commands that generate a follow-up `td auth login ...` hint (for example backup access or 403 insufficient-scope remediation) will drop this flag and try to spawn a browser again. On remote/non-desktop sessions that aren't auto-detected as headless, following the CLI's own hint reintroduces the exact behavior this flag is meant to avoid. Persist this flag with the auth metadata or otherwise thread it into the relogin command builder.

src/commands/auth/login.ts
```diff
@@ -43,6 +52,7 @@ export function attachTodoistLoginCommand(auth: Command, store: TodoistTokenStor
         renderSuccess: renderAuthSuccessPage,
         renderError: renderAuthErrorPage,
         openBrowser: async (url) => {
+            if (loginCommand?.opts().browserOpen === false) return
```
doistbot, line 55: <big><kbd>🟡 P2</kbd></big>  This callback adds a second browser-opening path in command code instead of going through the shared `src/lib/browser.ts` helper. That splits browser-launch behavior across two places, so any future tweaks we make there (error handling, platform flags, logging policy, etc.) won't apply to auth login. If login needs a non-announcing variant because cli-core already prints the URL, it's better to extend the helper with an option like `{ announce: false }` and call that here rather than bypassing the shared abstraction.

src/commands/auth/login.test.ts
```diff
@@ -141,3 +159,30 @@ describe('attachTodoistLoginCommand: onSuccess output formatting', () => {
...
+
+    it('opens the browser by default', async () => {
+        const { options } = attachAndCaptureLogin()
+        await options.openBrowser('https://todoist.com/oauth/authorize')
+        expect(openMock).toHaveBeenCalledWith('https://todoist.com/oauth/authorize')
+    })
+
+    it('skips opening the browser when --no-browser-open is set', async () => {
+        const { options, login } = attachAndCaptureLogin()
+        login.setOptionValue('browserOpen', false)
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This test sets Commander’s internal `browserOpen` value directly, so it will still pass if the public `--no-browser-open` flag stops parsing correctly (for example, if the option name changes, gets attached to the wrong command, or negation semantics drift). Since the new behavior is specifically CLI flag wiring, prefer parsing `--no-browser-open` with the existing `createTestProgram` harness and then invoking `options.openBrowser(...)` to verify the real user path.

doistbot (commented):
Thanks scottlovegrove for your contribution 😎 👊. The implementation for the `--no-browser-open` flag is clean, and the proactive updates to the SKILL docs are much appreciated.

Few things worth tightening:
- **Reauth flows:** Follow-up login hints (e.g., for missing scopes) don't preserve the `--no-browser-open` flag, which could unexpectedly spawn a browser later in headless environments.
- **Helper reuse:** The command bypasses `src/lib/browser.ts`; consider extending the shared helper with an option like `{ announce: false }` to keep browser-launch logic unified.
- **Test assertions:** The test verifies the skipped browser behavior by setting Commander's internal state directly rather than parsing the public flag via `createTestProgram`.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F376&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aff36bf60-5e62-11f1-986b-11c18a7e5009%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aff36bf60-5e62-11f1-986b-11c18a7e5009&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/login.ts
scottlovegrove: Done in ad1ba6b — the hook now calls `openInBrowser` from `src/lib/browser.ts` rather than importing `open` directly. Added an `{ announce: false }` option to the helper so it skips its `Opening <url>` line here (cli-core already prints the authorize URL), keeping a single shared browser-launch path.

src/commands/auth/login.test.ts
scottlovegrove: Done in ad1ba6b — the tests now parse the real `--no-browser-open` flag through `createTestProgram` / `parseAsync` and then invoke `options.openBrowser(...)`, so they exercise the actual flag wiring (name, attachment, negation) instead of Commander's internal `browserOpen` value.

src/commands/auth/login.ts
scottlovegrove: Intentionally not persisting this one. `--no-browser-open` describes the *environment the user is sitting in right now* (remote shell, no desktop browser), not a property of the credential — unlike `--read-only` / `--additional-scopes`, it has no effect on the token that gets granted. Baking it into `auth_flags` would make it survive into unrelated future logins: e.g. someone who used it once over SSH would then keep getting `--no-browser-open` suggested on their desktop, which is the opposite of helpful.

The scope-remediation hints fire in the same session the user is already in, and cli-core auto-detects the common non-desktop cases (SSH with no `DISPLAY`/`WAYLAND_DISPLAY`, WSL without interop) and skips the spawn there regardless of the flag. The residual gap is a remote session that *looks* like it has a display but does not — narrow, and the user can re-add the flag on the follow-up command (they just used it, so it's top of mind). Threading a transient session preference through the persisted credential model to close that gap isn't a worthwhile trade. Happy to revisit if it bites in practice.

rfgamaral (commented):
✅

PR #377, opened 2026-06-02
feat(apps): show UI extensions and install URL in `apps view`

scottlovegrove:
## What

Expand `td apps view <ref>` so that when an integration ships **UI extensions**, the view:

- Lists each UI extension as `<name> (<type>[: <sub-type>])`, where type is `context-menu` / `composer` / `settings` and sub-type is the `context-menu` context (`project`/`task`) or `composer` location (`task`/`comment`). Settings extensions have no sub-type.
- Prints the **installable URL** (`https://app.todoist.com/app/install/<distribution_token>`) — the link an integration owner shares so others can install it. This is required for any integration with UI extensions.

When an app has no UI extensions, neither the section nor the install URL is shown.

## Why

The view previously omitted the install URL entirely, which is required to distribute an integration that has UI extensions. The SDK already exposes `getUiExtensionsForApp` (previously unused) to detect them; the install URL mirrors how `todoist-web` builds it (`<origin>/app/install/<distribution_token>`).

## Distribution token reclassified as non-secret

The distribution token is the basis of a **shareable install link** by design, so it is now:
- always fetched (no longer gated behind `--include-secrets`)
- surfaced via the `Install URL` line / `installUrl` JSON field by default

The genuinely sensitive credentials — client secret, verification token, test token — remain hidden by default.

## Output

`--json` / `--ndjson` now always carry `uiExtensions`, `distributionToken`, and `installUrl` (`null` when the app has no UI extensions).

## Testing

- `npm run type-check` ✅
- Full suite: 1686 tests ✅ (new tests cover with/without UI extensions, sub-type formatting, and JSON payload)
- `npm run check` (lint + format) ✅
- Skill content updated + `npm run sync:skill`


doistbot (commented):
Thanks Scott for surfacing UI extensions and shareable install links in the apps view 😎 👊.

Few things worth tightening:
- Register `getUiExtensionsForApp` in `API_SPINNER_MESSAGES` and `METHOD_REQUIRED_FLAG` within `src/lib/api/core.ts` to ensure API errors are formatted correctly and missing permissions prompt for the `app-management` scope.
- Defer fetching the distribution token in the default plain-text flow until after confirming the app has UI extensions to avoid an unnecessary API call.

<details>

- **[P3] src/commands/apps/view.ts:28**: The `default` branch drops exhaustiveness on the SDK’s `UiExtension` discriminated union. If a new extension type is added upstream, this helper will keep compiling and silently render a partial label instead of forcing us to handle the new shape. Prefer explicit branches for the current variants plus a `never` fallback so SDK additions fail at compile time.
- **[P3] src/lib/skills/content.ts:354**: This description now overstates the plain-output behavior. `view.ts` only renders the install link when the app has UI extensions, but this says the distribution token is "always shown." Since `SKILL_CONTENT` is the source of truth for agents, tighten the wording to distinguish plain output (install URL only when UI extensions exist) from JSON, where `distributionToken` is always present.
- **[P3] src/commands/apps/apps.test.ts:551**: This fixture hand-builds the full SDK `UiExtension` shape even though these tests only assert `name`, `extensionType`, and the subtype field. That makes the suite noisy and brittle to unrelated SDK field changes. A small local `makeUiExtension(...)` helper (or shared fixture if you want to reuse it) would keep each case focused on just the fields under test.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F377&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Abaa51720-5e67-11f1-97fb-e405c9d85fc5%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Abaa51720-5e67-11f1-97fb-e405c9d85fc5&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/apps/view.ts
```diff
@@ -15,28 +17,53 @@ function hiddenLine(label: string): string {
...
-        revealSecrets ? api.getAppDistributionToken(app.id) : Promise.resolve(null),
-    ])
+    // Secret-bearing endpoints (getAppSecrets for clientSecret, verification and test
+    // tokens) are gated on `revealSecrets` so we never transport secret data onto the
+    // user's machine unless they asked for it. The distribution token is *not* a secret
+    // — it is the shareable install link's basis — so it is always fetched.
+    const [webhook, uiExtensions, distribution, secrets, verification, testToken] =
+        await Promise.all([
+            api.getAppWebhook(app.id),
+            api.getUiExtensionsForApp(app.id),
```
doistbot: <big><kbd>🟠 P1</kbd></big>  `api.getUiExtensionsForApp` is not registered in `API_SPINNER_MESSAGES` or `METHOD_REQUIRED_FLAG` in `src/lib/api/core.ts`.

Because it is missing from `API_SPINNER_MESSAGES`, the API proxy bypasses `wrapApiError`, meaning any network or API errors from this endpoint will skip the standard `CliError` formatting and potentially crash with a raw stack trace. Additionally, missing from `METHOD_REQUIRED_FLAG` means a 403 error will surface a generic remediation hint instead of prompting for the required `app-management` scope.

Please register `getUiExtensionsForApp` in both dictionaries in `src/lib/api/core.ts`.

src/commands/apps/view.ts
```diff
@@ -15,28 +17,53 @@ function hiddenLine(label: string): string {
...
+            api.getAppDistributionToken(app.id),
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `getAppDistributionToken()` is now called for every `td apps view`, including the default plain-text path where apps without UI extensions never render an install URL or distribution token. That adds an avoidable network round-trip to the common case. Since `getUiExtensionsForApp()` already tells us whether the token is needed, defer this call until after you know `uiExtensions.length > 0`, or only fetch it eagerly for `--json` / `--ndjson` where the payload always includes `distributionToken`.

src/commands/apps/view.ts
scottlovegrove: Done in 818514e — registered `getUiExtensionsForApp` in both `API_SPINNER_MESSAGES` and `METHOD_REQUIRED_FLAG` (`app-management`) in `src/lib/api/core.ts`.

src/commands/apps/view.ts
scottlovegrove: Done in 818514e — the distribution token is now fetched only when the app has UI extensions (plain output) or for `--json`/`--ndjson`; the common no-extensions plain path skips the call.

henningmu (commented):
✅

PR #396, opened 2026-06-10
feat(section): add description to create and update

lmjabreu:
## What

Surfaces section descriptions (26Q2 Project Essentials) in the CLI, mirroring the project/task description surface.

- `td section create` and `td section update` accept `--description <text>` and `--stdin`
- `td section update` no longer requires `--name` — description-only updates work, guarded by `NO_CHANGES`
- Curated `--json` output now includes `description`

## Same contract across descriptions

Identical to projects and tasks: the `--description` / `--stdin` flags, the `CONFLICTING_OPTIONS` guard, and the JSON essential field. Clearing follows the task pattern (pipe empty input via `--stdin`). The typed-args pattern matches the project PR — only the escape-hatch fields sit outside the SDK types (see below), keeping the rest of the payload compile-checked.

## ⚠️ SDK dependency (blocks full read-side behaviour)

`@doist/todoist-sdk` (pinned 10.3.2) does **not** yet model section descriptions, even though the backend shipped them on 2026-06-10:

- `SectionSchema` is a stripping `z.object` with no `description`, so the SDK drops it from section reads. **JSON output will not surface `description` until the SDK adds it.**
- `UpdateSectionArgs` requires `name` and omits `description`, so description-only updates and the `description` field need a localized escape hatch (documented inline).

The **write path works at runtime today** (the REST client forwards the field). The SDK gap is called out with comments in `create.ts` / `update.ts`. This PR needs an SDK release (add `description: z.string().nullable()` to `SectionSchema`; add `description` + optional `name` to `UpdateSectionArgs`) before the read side fully lands. Opening as a draft for that reason.

## Scope

- **In scope:** section descriptions in `create` / `update` / JSON.
- **Non-goals:** a `section view` text block (no such command exists; sections expose detail via `--json`), and the MCP (separate PR).

## Backend reference (asymmetry vs projects)

Section update clears with `null` (`NULL_CLEARS`), where project clears with `""` (`NULL_KEEPS_UNCHANGED`). The CLI sends what `--stdin` yields; the exact section clear wire value is a follow-up to confirm once the SDK exposes the field.

## Testing

- New unit tests: create (`--description`, `--stdin`, conflict), update (description-only, clear via empty `--stdin`, `NO_CHANGES`, conflict), and JSON regression guards asserting `description` survives the curated projection.
- Full suite green (1701 tests), type-check, lint/format, build all pass. Dry-run smoke-tested.

### DIFF
--- a/src/commands/section/create.ts
+++ b/src/commands/section/create.ts
@@ -1,21 +1,36 @@
 import chalk from 'chalk'
 import { getApi } from '../../lib/api/core.js'
+import { CliError } from '../../lib/errors.js'
 import { isQuiet } from '../../lib/global-args.js'
 import { formatJson, printDryRun } from '../../lib/output.js'
 import { resolveProjectId } from '../../lib/refs.js'
+import { readStdin } from '../../lib/stdin.js'
 
 interface CreateOptions {
     name: string
     project: string
+    description?: string
+    stdin?: boolean
     json?: boolean
     dryRun?: boolean
 }
 
 export async function createSection(options: CreateOptions): Promise<void> {
+    if (options.stdin && options.description !== undefined) {
+        throw new CliError('CONFLICTING_OPTIONS', 'Cannot use both --description and --stdin')
+    }
+    let description: string | undefined
+    if (options.stdin) {
+        description = await readStdin()
+    } else if (options.description) {
+        description = options.description
+    }
+
     if (options.dryRun) {
         printDryRun('create section', {
             Name: options.name,
             Project: options.project,
+            Description: description,
         })
         return
     }
@@ -23,10 +38,15 @@ export async function createSection(options: CreateOptions): Promise<void> {
     const api = await getApi()
     const projectId = await resolveProjectId(api, options.project)
 
-    const section = await api.addSection({
+    // `description` is forwarded by the REST client but not yet in the SDK's
+    // AddSectionArgs type (and the SDK strips it from section reads). Tracked as
+    // an SDK dependency. Keep the SDK signature so only `description` escapes typing.
+    const args: Parameters<typeof api.addSection>[0] & { description?: string } = {
         name: options.name,
         projectId,
-    })
+        ...(description !== undefined ? { description } : {}),
+    }
+    const section = await api.addSection(args)
 
     if (options.json) {
         console.log(formatJson(section, 'section'))
--- a/src/commands/section/create.ts
+++ b/src/commands/section/create.ts
@@ -38,15 +38,11 @@ export async function createSection(options: CreateOptions): Promise<void> {
     const api = await getApi()
     const projectId = await resolveProjectId(api, options.project)
 
-    // `description` is forwarded by the REST client but not yet in the SDK's
-    // AddSectionArgs type (and the SDK strips it from section reads). Tracked as
-    // an SDK dependency. Keep the SDK signature so only `description` escapes typing.
-    const args: Parameters<typeof api.addSection>[0] & { description?: string } = {
+    const section = await api.addSection({
         name: options.name,
         projectId,
         ...(description !== undefined ? { description } : {}),
-    }
-    const section = await api.addSection(args)
+    })
 
     if (options.json) {
         console.log(formatJson(section, 'section'))
--- a/src/commands/section/index.ts
+++ b/src/commands/section/index.ts
@@ -50,6 +50,8 @@ export function registerSectionCommand(program: Command): void {
         .description('Create a section')
         .option('--name <name>', 'Section name (required)')
         .option('--project <name>', 'Project name or id:xxx (required)')
+        .option('--description <text>', 'Section description (markdown)')
+        .option('--stdin', 'Read section description from stdin')
         .option('--json', 'Output the created section as JSON')
         .option('--dry-run', 'Preview what would happen without executing')
         .action((options) => {
@@ -76,11 +78,16 @@ export function registerSectionCommand(program: Command): void {
     const updateCmd = section
         .command('update [id]')
         .description('Update a section')
-        .option('--name <name>', 'New section name (required)')
+        .option('--name <name>', 'New section name')
+        .option(
+            '--description <text>',
+            'New description (markdown); pipe empty input via --stdin to clear',
+        )
+        .option('--stdin', 'Read section description from stdin')
         .option('--json', 'Output the updated section as JSON')
         .option('--dry-run', 'Preview what would happen without executing')
         .action((id, options) => {
-            if (!id || !options.name) {
+            if (!id) {
                 updateCmd.help()
                 return
             }
--- a/src/commands/section/section.test.ts
+++ b/src/commands/section/section.test.ts
@@ -5,10 +5,17 @@ vi.mock('../../lib/api/core.js', () => ({
     getApi: vi.fn(),
 }))
 
+vi.mock('../../lib/stdin.js', () => ({
+    readStdin: vi.fn(),
+}))
+
+import { readStdin } from '../../lib/stdin.js'
 import { setupApiMock } from '../../test-support/api-mock.js'
 import { fixtures } from '../../test-support/fixtures.js'
 import { type MockApi } from '../../test-support/mock-api.js'
 import { registerSectionCommand } from './index.js'
+
+const mockReadStdin = vi.mocked(readStdin)
 
 function createProgram() {
     return createTestProgram(registerSectionCommand)
@@ -200,6 +207,7 @@ describe('section create --json', () => {
         mockApi.addSection.mockResolvedValue({
             id: 'sec-new',
             name: 'New Section',
+            description: 'Section notes',
             projectId: 'proj-1',
             sectionOrder: 1,
         })
@@ -220,6 +228,8 @@ describe('section create --json', () => {
         const parsed = JSON.parse(output)
         expect(parsed.id).toBe('sec-new')
         expect(parsed.name).toBe('New Section')
+        // Regression guard: `description` must survive the curated JSON projection.
+        expect(parsed.description).toBe('Section notes')
         expect(consoleSpy).not.toHaveBeenCalledWith(expect.stringContaining('Created:'))
     })
 })
@@ -240,6 +250,7 @@ describe('section update --json', () => {
         mockApi.updateSection.mockResolvedValue({
             id: 'sec-1',
             name: 'New Name',
+            description: 'Updated notes',
             projectId: 'proj-1',
             sectionOrder: 1,
         })
@@ -259,6 +270,8 @@ describe('section update --json', () => {
         const parsed = JSON.parse(output)
         expect(parsed.id).toBe('sec-1')
         expect(parsed.name).toBe('New Name')
+        // Regression guard: `description` must survive the curated JSON projection.
+        expect(parsed.description).toBe('Updated notes')
         expect(consoleSpy).not.toHaveBeenCalledWith(expect.stringContaining('Updated:'))
     })
 })
@@ -646,6 +659,76 @@ describe('section create', () => {
 
         expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining('sec-xyz'))
     })
+
+    it('creates section with --description', async () => {
+        const program = createProgram()
+
+        mockApi.getProject.mockResolvedValue({ id: 'proj-1', name: 'Work' })
+        mockApi.addSection.mockResolvedValue({ id: 'sec-new', name: 'Review' })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'section',
+            'create',
+            '--name',
+            'Review',
+            '--project',
+            'id:proj-1',
+            '--description',
+            'Pending review',
+        ])
+
+        expect(mockApi.addSection).toHaveBeenCalledWith({
+            name: 'Review',
+            projectId: 'proj-1',
+            description: 'Pending review',
+        })
+    })
+
+    it('reads description from stdin with --stdin', async () => {
+        const program = createProgram()
+
+        mockReadStdin.mockResolvedValue('Piped description')
+        mockApi.getProject.mockResolvedValue({ id: 'proj-1', name: 'Work' })
+        mockApi.addSection.mockResolvedValue({ id: 'sec-new', name: 'Review' })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'section',
+            'create',
+            '--name',
+            'Review',
+            '--project',
+            'id:proj-1',
+            '--stdin',
+        ])
+
+        expect(mockApi.addSection).toHaveBeenCalledWith(
+            expect.objectContaining({ description: 'Piped description' }),
+        )
+    })
+
+    it('rejects --description together with --stdin', async () => {
+        const program = createProgram()
+
+        await expect(
+            program.parseAsync([
+                'node',
+                'td',
+                'section',
+                'create',
+                '--name',
+                'Review',
+                '--project',
+                'id:proj-1',
+                '--description',
+                'x',
+                '--stdin',
+            ]),
+        ).rejects.toHaveProperty('code', 'CONFLICTING_OPTIONS')
+    })
 })
 
 describe('section delete', () => {
@@ -752,6 +835,66 @@ describe('section update', () => {
             name: 'New Name',
         })
         expect(consoleSpy).toHaveBeenCalledWith('Updated: Old Name → New Name (id:sec-1)')
+    })
+
+    it('updates description only, without a name change', async () => {
+        const program = createProgram()
+        const consoleSpy = captureConsole()
+
+        mockApi.getSection.mockResolvedValue({ id: 'sec-1', name: 'Planning' })
+        mockApi.updateSection.mockResolvedValue({ id: 'sec-1', name: 'Planning' })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'section',
+            'update',
+            'id:sec-1',
+            '--description',
+            'Sprint backlog',
+        ])
+
+        expect(mockApi.updateSection).toHaveBeenCalledWith('sec-1', {
+            description: 'Sprint backlog',
+        })
+        expect(consoleSpy).toHaveBeenCalledWith('Updated: Planning (id:sec-1)')
+    })
+
+    it('clears description with empty stdin', async () => {
+        const program = createProgram()
+
+        mockReadStdin.mockResolvedValue('')
+        mockApi.getSection.mockResolvedValue({ id: 'sec-1', name: 'Planning' })
+        mockApi.updateSection.mockResolvedValue({ id: 'sec-1', name: 'Planning' })
+
+        await program.parseAsync(['node', 'td', 'section', 'update', 'id:sec-1', '--stdin'])
+
+        expect(mockApi.updateSection).toHaveBeenCalledWith('sec-1', { description: '' })
+    })
+
+    it('throws when no changes specified', async () => {
+        const program = createProgram()
+
+        await expect(
+            program.parseAsync(['node', 'td', 'section', 'update', 'id:sec-1']),
+        ).rejects.toHaveProperty('code', 'NO_CHANGES')
+    })
+
+    it('rejects --description together with --stdin', async () => {
+        const program = createProgram()
+
+        await expect(
+            program.parseAsync([
+                'node',
+                'td',
+                'section',
+                'update',
+                'id:sec-1',
+                '--description',
+                'x',
+                '--stdin',
+            ]),
+        ).rejects.toHaveProperty('code', 'CONFLICTING_OPTIONS')
     })
 })
 
--- a/src/commands/section/section.test.ts
+++ b/src/commands/section/section.test.ts
@@ -228,7 +228,9 @@ describe('section create --json', () => {
         const parsed = JSON.parse(output)
         expect(parsed.id).toBe('sec-new')
         expect(parsed.name).toBe('New Section')
-        // Regression guard: `description` must survive the curated JSON projection.
+        // Regression guard for the CLI's curated JSON projection: once the SDK
+        // exposes section `description` (it strips it today), it must not be
+        // dropped from the output. The mock simulates that post-SDK response.
         expect(parsed.description).toBe('Section notes')
         expect(consoleSpy).not.toHaveBeenCalledWith(expect.stringContaining('Created:'))
     })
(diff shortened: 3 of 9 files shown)


lmjabreu:
@Doistbot /review

src/commands/section/index.ts
```diff
@@ -50,6 +50,8 @@ export function registerSectionCommand(program: Command): void {
         .description('Create a section')
         .option('--name <name>', 'Section name (required)')
         .option('--project <name>', 'Project name or id:xxx (required)')
+        .option('--description <text>', 'Section description (markdown)')
```
doistbot, line 53: <big><kbd>🟠 P1</kbd></big>  Per `AGENTS.md`, whenever options are added in `src/commands/`, the `SKILL_CONTENT` in `src/lib/skills/content.ts` must be updated to match. Please add coverage for the new `--description` and `--stdin` flags for `section create` and `section update` to the agent guidelines.

src/commands/section/create.ts
```diff
@@ -1,32 +1,52 @@
...
 }
 
 export async function createSection(options: CreateOptions): Promise<void> {
+    if (options.stdin && options.description !== undefined) {
+        throw new CliError('CONFLICTING_OPTIONS', 'Cannot use both --description and --stdin')
+    }
+    let description: string | undefined
+    if (options.stdin) {
+        description = await readStdin()
+    } else if (options.description) {
```
doistbot, line 25: <big><kbd>🟡 P2</kbd></big>  Checking `if (options.description)` evaluates to false when a user explicitly passes `--description ""`. Use `if (options.description !== undefined)` to ensure an empty string is faithfully captured rather than silently ignored.

src/commands/section/update.ts
```diff
@@ -1,28 +1,63 @@
...
+    if (options.stdin && options.description !== undefined) {
+        throw new CliError('CONFLICTING_OPTIONS', 'Cannot use both --description and --stdin')
+    }
     const id = lenientIdRef(sectionId, 'section')
 
+    // SDK dependency: UpdateSectionArgs currently requires `name` and omits
+    // `description`. We keep the SDK signature via Partial so only the two
+    // escape hatches (optional name, description) sit outside the SDK types.
+    type SectionUpdateArgs = Parameters<Awaited<ReturnType<typeof getApi>>['updateSection']>[1]
+    const args: Partial<SectionUpdateArgs> & { description?: string } = {}
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `Partial<SectionUpdateArgs>` weakens the entire SDK contract, not just `name`. If a later SDK release adds any other required update field, this command will still compile and silently stop enforcing it. Prefer a narrower escape hatch here, e.g. `Omit<SectionUpdateArgs, 'name'> & { name?: SectionUpdateArgs['name']; description?: string }`, so only the intended fields fall outside the SDK type.

src/commands/section/update.ts
```diff
@@ -1,28 +1,63 @@
...
-        const updated = await api.updateSection(id, { name: options.name })
+        const updated = await api.updateSection(id, updateArgs)
         console.log(formatJson(updated, 'section'))
         return
     }
 
     const section = await api.getSection(id)
-    const updated = await api.updateSection(id, { name: options.name })
-    if (!isQuiet()) console.log(`Updated: ${section.name} → ${updated.name} (id:${id})`)
+    const updated = await api.updateSection(id, updateArgs)
```
doistbot, line 60: <big><kbd>🟡 P2</kbd></big>  Unnecessary blocking network request and missed concurrency.

Since `--name` is now optional, `api.getSection(id)` is being fetched unconditionally on line 57 even though `section.name` is only used when `!isQuiet() && args.name`. For description-only updates, this is a redundant blocking API roundtrip.

You can avoid the unnecessary fetch and parallelize the work when a name change is actually requested:

```typescript
const fetchOldName = !isQuiet() && args.name
const [section, updated] = await Promise.all([
    fetchOldName ? api.getSection(id) : Promise.resolve(null),
    api.updateSection(id, updateArgs)
])

if (!isQuiet()) {
    const label = section ? `${section.name} → ${updated.name}` : updated.name
    console.log(`Updated: ${label} (id:${id})`)
}
```

src/commands/section/section.test.ts
```diff
@@ -646,6 +659,76 @@ describe('section create', () => {
...
+        ])
+
+        expect(mockApi.addSection).toHaveBeenCalledWith({
+            name: 'Review',
+            projectId: 'proj-1',
+            description: 'Pending review',
+        })
+    })
+
+    it('reads description from stdin with --stdin', async () => {
```
doistbot, line 693: <big><kbd>🟡 P2</kbd></big>  This adds normal-path `--stdin` coverage, but we still don't exercise `--stdin` together with `--dry-run` for either create or update. That's the highest-value branch here, because both commands now read stdin before their dry-run early return; without a nearby test, a regression will hang or consume input even when the user asked for a preview. Please add a dry-run case that asserts `readStdin` and the API are not called.

src/commands/section/section.test.ts
```diff
@@ -220,6 +228,8 @@ describe('section create --json', () => {
         const parsed = JSON.parse(output)
         expect(parsed.id).toBe('sec-new')
         expect(parsed.name).toBe('New Section')
+        // Regression guard: `description` must survive the curated JSON projection.
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This "regression guard" is stubbing a response shape the current pinned SDK can't produce: `setupApiMock()` lets `addSection` return `{ description }`, but the real section response currently strips that field. So this test stays green while `td section ... --json` still omits `description` in practice. Either move this check to the `formatJson()` unit seam in `src/lib/output.test.ts` (where it's clearly a pure projection test), or defer the command-level assertion until the SDK update lands.

doistbot (commented):
This cleanly implements section descriptions by mirroring the established project and task patterns in the CLI.

Few things worth tightening:
- **Agent skill docs**: The new `--description` and `--stdin` flags need to be added to `SKILL_CONTENT` in `src/lib/skills/content.ts` per the repository guidelines.
- **Argument mapping and types**: Empty string descriptions are currently dropped by a truthiness check, and the `Partial<SectionUpdateArgs>` SDK workaround could be narrowed (e.g., using `Omit`) to preserve type safety for other fields.
- **Network efficiency**: The update command fetches the existing section unconditionally, which creates a redundant blocking API call for description-only updates.
- **Test precision**: A `--stdin` paired with `--dry-run` test is needed to ensure previews don't accidentally hang, and the JSON "regression guard" should ideally be adjusted since it currently mocks an SDK response shape that hasn't shipped yet.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F396&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aa1e7ef50-6510-11f1-9ae4-a0394121210a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aa1e7ef50-6510-11f1-9ae4-a0394121210a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/section/index.ts
lmjabreu: Done. Added section `--description` and `--stdin` examples to SKILL_CONTENT and regenerated SKILL.md.

src/commands/section/create.ts
lmjabreu: Keeping as-is. Empty `--description` being a no-op is the intentional task/project contract; clearing is via `--stdin`. Section-specific clear semantics (backend NULL_CLEARS) are tracked as the SDK follow-up.

src/commands/section/update.ts
lmjabreu: Done. Replaced `Partial<SectionUpdateArgs>` with an explicit `{ name?; description? }` arg type; the SDK cast stays only at the call site.

src/commands/section/update.ts
lmjabreu: Fixed. Description-only updates skip the `getSection` fetch; the existing name is read only when renaming.

src/commands/section/section.test.ts
lmjabreu: Done. Added a `--stdin --dry-run` test asserting stdin is read, the API isn't called, and the preview prints.

src/commands/section/section.test.ts
lmjabreu: Done. Added a comment clarifying the guard simulates the post-SDK response shape and validates the CLI's curated projection.

lmjabreu:
@Doistbot /review

src/commands/section/update.ts
```diff
@@ -1,28 +1,64 @@
...
+    }
     const id = lenientIdRef(sectionId, 'section')
 
+    // SDK dependency: UpdateSectionArgs requires `name` and omits `description`,
+    // so the args type lists exactly the two fields we send. The cast at the
+    // call site bridges to the SDK signature until it models both.
+    const args: { name?: string; description?: string } = {}
+    if (options.name) args.name = options.name
+    if (options.stdin) {
+        args.description = await readStdin()
```
doistbot: <big><kbd>🟠 P1</kbd></big>  Empty stdin won't clear a section description with this implementation. `readStdin()` returns `''`, and this branch forwards that as `description: ''`, but section updates clear on `null`, not an empty string. That means `td section update id:sec-1 --stdin < /dev/null` will keep the existing description even though the CLI help now advertises it as a clear path. Normalize empty stdin to `null` here (and widen the local escape-hatch type to allow it), or remove the clear claim until the SDK/client can represent it correctly.

src/commands/section/update.ts
```diff
@@ -1,28 +1,64 @@
...
         console.log(formatJson(updated, 'section'))
         return
     }
 
-    const section = await api.getSection(id)
-    const updated = await api.updateSection(id, { name: options.name })
-    if (!isQuiet()) console.log(`Updated: ${section.name} → ${updated.name} (id:${id})`)
+    // Only fetch the existing section when renaming, to show "old → new".
+    // Description-only updates skip the extra blocking request.
+    const previousName = args.name ? (await api.getSection(id)).name : undefined
```
doistbot: <big><kbd>🟡 P2</kbd></big>  The `getSection` fetch is now correctly skipped for description-only updates, but it will still fire for name changes even when `--quiet` is active. You can completely avoid this network roundtrip when the output isn't going to be printed by adding an `!isQuiet()` check:

```typescript
const previousName = args.name && !isQuiet() ? (await api.getSection(id)).name : undefined
```

src/commands/section/section.test.ts
```diff
@@ -753,6 +864,66 @@ describe('section update', () => {
...
+            'Sprint backlog',
+        ])
+
+        expect(mockApi.updateSection).toHaveBeenCalledWith('sec-1', {
+            description: 'Sprint backlog',
+        })
+        expect(consoleSpy).toHaveBeenCalledWith('Updated: Planning (id:sec-1)')
+    })
+
+    it('clears description with empty stdin', async () => {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `section update` now has the same `--stdin`/dry-run interaction as create, but there’s still no `section update id:sec-1 --stdin --dry-run` case here. That leaves the highest-risk preview path untested: a regression could start consuming stdin or calling `updateSection()` during a dry run without this suite failing. Please add a sibling test that mocks `readStdin`, checks the preview output, and asserts the API isn’t called.

src/lib/skills/content.ts
```diff
@@ -231,7 +231,10 @@ td section list "Roadmap"
 td section list --search "Planning"
 td section list --search "Planning" --project "Roadmap"
 td section create --project "Roadmap" --name "In Progress"
+td section create --project "Roadmap" --name "QA" --description "Bugs to verify"
+td section create --project "Roadmap" --name "Imported" --stdin   # read the description from stdin
 td section update id:123 --name "Done"
+td section update id:123 --description "Sprint backlog"            # description-only update
```
doistbot, line 237: <big><kbd>🟡 P2</kbd></big>  `AGENTS.md` requires `SKILL_CONTENT` to stay in sync when command flags change, but `section update`'s new `--stdin` path still isn't documented here. Please add an update example that uses `--stdin` (ideally the empty-input clear case), then re-sync `skills/todoist-cli/SKILL.md`.

doistbot (commented):
This cleanly introduces section descriptions to the CLI, mirroring the task and project patterns while gracefully handling the temporary SDK limitations.

Few things worth tightening:
- Map empty `--stdin` input to `null` during updates so the description actually clears as advertised.
- Skip the `getSection` network roundtrip for name changes when the `--quiet` flag is active.
- Add a `--dry-run` test for the update `--stdin` path to ensure the API isn't accidentally called and input isn't consumed.
- Include an update example using `--stdin` in `SKILL_CONTENT` to keep the agent command reference in sync.

<details>

- **[P3] src/commands/section/section.test.ts:231**: This regression guard is still living at the command seam even though the pinned SDK/mock layer can't currently return `section.description`. That makes the test depend on an impossible response shape instead of the behavior this PR actually owns: the curated projection in `formatJson`/`SECTION_ESSENTIAL_FIELDS`. Please move this assertion to `src/lib/output.test.ts` (same for the update case below) and keep the command test focused on wiring.
- **[P3] src/commands/section/section.test.ts:872**: This mock for `getSection` is no longer needed. Since description-only updates now actively skip fetching the existing section, this endpoint is never called in this scenario. The same applies to the empty stdin test on line 895.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F396&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A7ba072a0-6514-11f1-8d40-9083ce84e7c7%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A7ba072a0-6514-11f1-8d40-9083ce84e7c7&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/section/update.ts
```diff
@@ -1,28 +1,65 @@
...
```
doistbot: <big><kbd>🟡 P2</kbd></big>  The `getSection` network fetch still executes when updating a name, even if `--quiet` is active. You can completely bypass this blocking request when output is suppressed by adding an `!isQuiet()` check:

```typescript
const previousName = args.name && !isQuiet() ? (await api.getSection(id)).name : undefined
```

src/lib/skills/content.ts
doistbot, line 237: <big><kbd>🟡 P2</kbd></big>  Per `AGENTS.md`, `SKILL_CONTENT` must stay in sync with command flags so agents can use them. `section update` now supports `--stdin`, but there is no updated example here. Please add a line demonstrating its use (e.g., for clearing a description) and regenerate `SKILL.md`.

src/commands/section/section.test.ts
```diff
@@ -753,6 +864,66 @@ describe('section update', () => {
...
+                'td',
+                'section',
+                'update',
+                'id:sec-1',
+                '--description',
+                'x',
+                '--stdin',
+            ]),
+        ).rejects.toHaveProperty('code', 'CONFLICTING_OPTIONS')
+    })
```
doistbot, line 926: <big><kbd>🟡 P2</kbd></big>  `section update` now supports the `--stdin` dry-run interaction, but the highest-risk preview path remains untested. Please add a sibling `td section update ... --stdin --dry-run` test that mocks `readStdin`, asserts the correct preview output, and verifies the API isn't called.

doistbot (commented):
This cleanly introduces section descriptions to the CLI and aligns nicely with our established project and task patterns.

Few things worth tightening:
- Skip the blocking `getSection` network fetch when updating a section name in `--quiet` mode.
- Add `--stdin` usage examples to `SKILL_CONTENT` for `section update` and regenerate `SKILL.md` to keep AI agents in sync.
- Include a test for the `--stdin --dry-run` update workflow to ensure the preview output is correct and the API is safely bypassed.

<details>

- **[P3] src/commands/section/section.test.ts:232**: Since package.json shows the SDK was bumped to `10.4.0` in this PR (adding native description support), this comment about simulating the "post-SDK response" because it "strips it today" is likely stale. The same applies for the update test comment on line 276.
- **[P3] src/commands/section/section.test.ts:872**: This `getSection` mock is unused and no longer needed since description-only updates actively bypass fetching the existing section. The same applies to the empty stdin test below on line 895.
- **[P3] src/commands/section/section.test.ts:207**: The new tests in this file are hand-building section payloads instead of starting from `fixtures.sections.*`, which is the repo’s standard for mock entities. Reusing fixtures here would keep these cases aligned with SDK shape changes and make the suite less brittle.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F396&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A60bed190-6585-11f1-99d1-e3da8dc215f6%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A60bed190-6585-11f1-99d1-e3da8dc215f6&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/section/update.ts
lmjabreu: Fixed. Already in 09f4eef with the 10.4.0 bump: empty stdin sends `description: null` (`piped === '' ? null : piped`), and a test asserts `updateSection` is called with `{ description: null }`.

src/commands/section/index.ts
lmjabreu: Done. Added a `section update --stdin` example (including the empty-clear case) in 6829ee7 and regenerated SKILL.md.

src/lib/skills/content.ts
lmjabreu: Done. Added the `section update --stdin` example in 6829ee7.

lmjabreu: Done. Added in 6829ee7.

src/commands/section/update.ts
lmjabreu: Fixed. The fetch now runs only for name updates with output on (`args.name && !isQuiet()`), in 6829ee7.

src/commands/section/update.ts
lmjabreu: Fixed. In 6829ee7: the fetch is skipped under `--quiet`.

src/commands/section/update.ts
lmjabreu: Fixed. In 6829ee7 (`args.name && !isQuiet()`).

src/commands/section/section.test.ts
lmjabreu: Done. Added the `section update --stdin --dry-run` test in 6829ee7; the create counterpart already existed.

src/commands/section/section.test.ts
lmjabreu: Done. Added in 6829ee7.

src/commands/section/section.test.ts
lmjabreu: Done. Added in 6829ee7.

src/commands/section/create.ts
lmjabreu: Keeping as-is. On create, empty and absent description are functionally identical (the section is new either way), and this matches task/project create. Clearing is an update concern and is covered there.

src/commands/section/update.ts
lmjabreu: Done. Superseded by the 10.4.0 bump: an explicit `{ name?, description? }` type replaced the `Partial`, with one documented cast bridging the SDK's `RequireAtLeastOne`.

src/commands/section/section.test.ts
lmjabreu: Done. Moot since the 10.4.0 bump: the SDK now ships `description` on section reads, so the mock matches the real response shape.

rfgamaral (approved):

PR #420, opened 2026-07-06
feat(apps): add --set-webhook-url to apps update

scottlovegrove:
## What

Adds `td apps update <ref> --set-webhook-url <url>` to change an app's webhook callback URL. Until now the app-management commands could only edit OAuth redirect URIs — the webhook was shown by `td apps view` but couldn't be written.

## Why

We needed to be able to update an app's webhook URL from the CLI. The webhook holds a single URL, so this is a straight swap — a "set" operation.

## How

Webhooks are a separate endpoint from the app record (`UpdateAppArgs` has no webhook field), so the flag reads the current webhook via `getAppWebhook` and writes through `updateAppWebhook`. Because `updateAppWebhook` requires the full event list, the command preserves the existing webhook's `events` and `version` and changes only the `callbackUrl`.

Behaviour:
- A webhook must already exist — creating one needs an event list a URL-only swap can't supply. If none is configured, the command errors with `NO_WEBHOOK`.
- The URL must be a public `https://` target; localhost and custom schemes are rejected (unlike OAuth redirect URIs).
- Setting the URL to the value already configured exits 0 with a message and makes no API call.
- Mutually exclusive with the OAuth-redirect flags; supports `--dry-run` and `--json` (the updated webhook object).

```
td apps update id:9909 --set-webhook-url https://example.com/webhook
```

## Tests

- New `apps update --set-webhook-url` block: happy-path swap (asserts preserved events/version), `--json`, `NO_WEBHOOK`, invalid URL, same-URL no-op, `--dry-run`, mutual exclusion.
- `validateWebhookUrl` unit cases in `helpers.test.ts`.
- Skill content updated and `SKILL.md` regenerated via `sync:skill`.

Full suite green (1710 tests), type-check + lint + format clean.


src/commands/apps/helpers.ts
```diff
@@ -22,6 +22,13 @@ function validateIntegrationUrl(url: string): boolean {
     }
 }
 
+// A webhook callback must be a public https:// endpoint. Unlike OAuth redirect
+// URIs, localhost and custom schemes are not valid webhook targets, so this
+// only accepts the plain https form.
+export function validateWebhookUrl(url: string): boolean {
+    return validateIntegrationUrl(url)
```
doistbot, line 38: <a href="#">P2</a>  `validateIntegrationUrl()` will still accept values like `https://foo.localhost/webhook`, because `foo.localhost` matches the generic `https://<host>` regex. `.localhost` is a reserved loopback-only TLD, so this lets a non-public callback URL through even though this flag is supposed to reject local targets. Please explicitly reject `localhost` and `*.localhost` before delegating to the generic validator, and add a test for that case.

src/commands/apps/update.ts
```diff
@@ -139,3 +154,58 @@ export async function updateApp(ref: string, options: UpdateAppOptions): Promise
...
+    }
+
+    const updated = await api.updateAppWebhook({
+        appId: app.id,
+        callbackUrl: url,
+        events: webhook.events,
+        version: webhook.version,
+    })
+
+    if (options.json) {
```
doistbot, line 204: <a href="#">P2</a>  This branch changes `td apps update --json` from the existing app-shaped payload to a bare webhook object (and the no-op branch above does the same). That gives one command two unrelated JSON contracts depending on which flag was used, which is a break from the surrounding update commands and forces scripts to add flag-specific parsers. Consider either splitting webhook edits into their own subcommand, or keeping `apps update --json` app-shaped and exposing webhook data under a separate key.

src/commands/apps/update.ts
```diff
@@ -139,3 +154,58 @@ export async function updateApp(ref: string, options: UpdateAppOptions): Promise
...
+
+// The webhook is a separate endpoint from the app record, and it holds a single
+// callback URL, so setting it is a straight swap. `updateAppWebhook` requires
+// the full event list, so we read the existing webhook and preserve its events
+// and version — a webhook must already exist for a URL-only swap to be possible.
+async function setWebhook(ref: string, url: string, options: UpdateAppOptions): Promise<void> {
+    if (!validateWebhookUrl(url)) throw invalidWebhookUrl(url)
+
+    const api = await getApi()
+    const app = await resolveAppRef(api, ref)
```
doistbot, line 166: <a href="#">P2</a>  This uses `resolveAppRef`, which does `getApps()` and then `getApp(matched.id)` for name-based refs. In this code path we only need `id` and `displayName` before calling `getAppWebhook`, so that second app fetch is just added latency. `td apps update "My App" --set-webhook-url ...` now does 3 sequential round-trips (`getApps` → `getApp` → `getAppWebhook`) before it can update. A lighter resolver that returns the matched list item here would avoid the extra call.

doistbot (commented):
This PR adds `--set-webhook-url` to `td apps update`, letting users swap an app's webhook callback URL via a separate API endpoint while preserving the existing events list and version.

Few things worth tightening:

- **`https://*.localhost` URLs slip through validation** — `validateIntegrationUrl` accepts `foo.localhost` as a generic HTTPS host, so explicitly reject `localhost` and `*.localhost` before delegating to the generic validator, and add a test for that case.
- **`--json` now has two unrelated payload shapes on the same command** (app-shaped for OAuth redirects vs. bare webhook object for `--set-webhook-url`). Consider either splitting webhook edits into their own subcommand or nesting webhook data under a separate key in an app-shaped payload so scripts don't need flag-specific parsers.
- **`resolveAppRef` causes an extra API round-trip** for name-based refs — it fetches the full app record when only `id` and `displayName` are needed before calling `getAppWebhook`. A lighter resolver returning the matched list item would avoid the redundant `getApp` call.
- **Add `NO_WEBHOOK` to the `ErrorCode` union** in `errors.ts` for consistency with the other error codes, and fill the missing `--json` no-op test case for the same-URL path (the OAuth redirect tests already model this pattern).

<details>

- **<a href="#">P3</a> src/lib/errors.ts:1**: The new `NO_WEBHOOK` error code (used in `update.ts:171`) is missing from the `ErrorCode` union in `errors.ts`. Every other error code thrown across the CLI is listed there for intellisense and discoverability — the type works because of the `| (string & {})` escape hatch, but omitting it breaks the convention. Add `| 'NO_WEBHOOK'` to the union, near the other state-error codes like `NO_CHANGES`, `NO_DUE_DATE`, etc.
- **<a href="#">P3</a> src/commands/apps/apps.test.ts:1129**: The `--json` branch of the same-URL no-op path (`update.ts:179-181`) has no test. The existing test at line 1129 only exercises the plain-text output path. The OAuth redirect tests already model the equivalent coverage pattern — `'remove no-op with --json outputs the unchanged app as JSON'` (near line 855) — so this is a straightforward gap to fill with the same `WEBHOOK` fixture already in scope.
- **<a href="#">P3</a> src/commands/apps/apps.test.ts:1148**: The same-URL no-op test only covers plain output. `setWebhook` has a dedicated `--json` branch for the no-op case (update.ts L179-181) whose comment calls out keeping `--json` parseable for scripts — a regression there (e.g. printing the human message instead of JSON) would silently break scripted callers. The adjacent `--remove-oauth-redirect` tests already model this exact case ("remove no-op with --json outputs the unchanged app as JSON"), so adding a webhook variant — assert `updateAppWebhook` not called and `consoleSpy` emits parseable JSON with the unchanged `callbackUrl` — is low-cost and in-pattern.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F420&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Acdb396d0-7941-11f1-8c0c-84cfb231b3db%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Acdb396d0-7941-11f1-8c0c-84cfb231b3db&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/apps/helpers.ts
scottlovegrove: Done in 5acbee2 — `validateWebhookUrl` now rejects `localhost` and `*.localhost` hosts before delegating to the generic validator, with a `https://foo.localhost/webhook` test case added.

src/commands/apps/update.ts
scottlovegrove: Not changing this. The two flags edit genuinely different resources — the OAuth-redirect flags mutate the app record, `--set-webhook-url` mutates the separate webhook. Returning the resource that was actually edited is the natural contract; a caller passing `--set-webhook-url` is performing a webhook operation and expects the webhook back. Echoing an app-shaped payload would either misrepresent an unchanged app record or nest the webhook under a key that the `apps update --json` essential-field filter strips. A separate subcommand was considered during planning and the flag-on-`update` surface was chosen deliberately.

src/commands/apps/update.ts
scottlovegrove: Not changing this. The webhook path shares `resolveAppRef` with the OAuth-redirect paths so all of `apps update` resolves refs uniformly and surfaces the same friendly `APP_NOT_FOUND` on a miss. The extra `getApp` only affects name-based refs (`id:N` and raw-numeric refs are a single fetch), and is a marginal latency cost against the value of a single, consistent resolver — duplicating the id/name resolution and 404 handling just to drop one call for the name case is not a worthwhile trade here.

henningmu (commented):
✅

PR #431, opened 2026-07-15
feat(apps): add --name and --description to apps update

scottlovegrove:
## What

Adds `td apps update <ref> --name <name>` and `--description <description>` so an app can be renamed and its description set from the CLI (mapping to the SDK's `displayName` / `description` fields).

While here, the artificial one-flag-per-invocation restriction is dropped: **every update flag now combines in a single command**. An invocation performs up to two API calls — an app-record patch (`updateApp`, carrying name/description/OAuth-redirect changes) followed by a webhook URL swap (`updateAppWebhook`, a separate endpoint).

## Behaviour

- `--name` maps to `displayName`; an empty/whitespace `--name` is rejected with `INVALID_OPTIONS`.
- `--description ""` clears the description.
- The **only** remaining mutual exclusion is `--add-oauth-redirect` + `--remove-oauth-redirect` (they read-modify-write the same field) → `CONFLICTING_OPTIONS`.
- A real OAuth-redirect removal still gates the **whole batch** behind `--yes`: without it, nothing is performed (plain output prints a combined "would update" preview; `--json` throws `CONFIRMATION_REQUIRED`).
- Execution order is record-patch-then-webhook. If the record patch succeeds and the webhook call then fails, the record change stays persisted (no rollback) — documented in-code.
- `--json` shape follows the surfaces touched: a lone record change emits the bare app object, a lone webhook change the bare webhook object, and touching both emits `{ "app": <app>, "webhook": <webhook> }`.
- `--dry-run` prints one combined preview of every pending change.

Existing single-flag invocations keep their error codes and JSON shapes. The human-facing success/preview wording unifies to `Updated <name> (id:N) — <clause>…`.

## Examples

```
td apps update id:9909 --name "My Renamed App"
td apps update id:9909 --description "Does a useful thing"
td apps update id:9909 --name "My App" --description "New blurb" --set-webhook-url https://example.com/webhook
```

## Notes

- Adds a small `projectEntity` export to `output.ts` so the embedded `app` in the combined JSON keeps the same essential-field projection as the lone-record path.
- Skill docs regenerated via `npm run sync:skill`.

## Testing

- `npm run type-check` — clean
- `npm test` — all passing (12 new apps tests covering the new + combined paths)
- `npm run check` — clean


src/lib/output.ts
```diff
@@ -378,6 +378,13 @@ function processItem<T extends object>(
     return base
 }
 
+// Return an entity projected to its essential fields (the same projection
+// `formatJson` applies) as an object, for callers that need to embed it inside
+// a larger JSON structure rather than emit it directly.
+export function projectEntity<T extends object>(data: T, type: EntityType): object {
```
doistbot: <a href="#">P2</a>  `projectEntity` duplicates the existing `processJsonItem` (line 359). Both are thin wrappers that call `processItem(data, type, false, false)` with identical behavior. The `src/lib/` catalog in CODEBASE.md explicitly lists `processJsonItem` as the "essential-field picker used by JSON callers that build composite payloads" — that's exactly the use case this PR needs. Use `processJsonItem` directly in `update.ts` instead of introducing a second identically-behaved export. If the name `projectEntity` is preferred, rename the existing `processJsonItem` and remove the old one rather than keeping both.

src/commands/apps/update.ts
```diff
@@ -32,180 +34,223 @@ function invalidWebhookUrl(url: string): CliError {
...
+
+    // Phase C — build the plan (no writes).
+    const patch: Parameters<typeof api.updateApp>[1] = {}
+    const changes: string[] = []
+    const noops: string[] = []
+    let removeIsMutation = false
+    let webhookWillChange = false
+
+    if (name !== undefined) {
+        patch.displayName = name
```
doistbot: <a href="#">P2</a>  `--name` always adds `displayName` to the patch, even when it equals `app.displayName` (and the description block has the same issue). Thus `td apps update <ref> --name "<current name>"` performs an avoidable `updateApp` request; combined with an unchanged webhook URL it still writes despite every requested value already matching. Compare each supplied value with the resolved app and treat equal values as no-ops before building the patch.

src/commands/apps/apps.test.ts
```diff
@@ -1195,24 +1199,304 @@ describe('apps update --set-webhook-url', () => {
...
+        ])
+
+        expect(mockApi.updateApp).toHaveBeenCalledWith('9909', { displayName: 'Renamed App' })
+        expect(mockApi.updateAppWebhook).toHaveBeenCalledWith({
+            appId: '9909',
+            callbackUrl: 'https://new.example.com/webhook',
+            events: WEBHOOK.events,
+            version: '1',
+        })
+        const parsed = JSON.parse(consoleSpy.mock.calls[0][0] as string)
```
doistbot, line 1417: <a href="#">P2</a>  The combined `{ app, webhook }` JSON test verifies structure and key values but doesn't check that `projectEntity` stripped non-essential fields. The existing `--add-oauth-redirect --json` test (line ~982) explicitly asserts `expect(parsed).not.toHaveProperty('iconSm')` and `expect(parsed).not.toHaveProperty('userId')`. Since `projectEntity` is a new export added specifically for this combined-output path, add the same assertions here: `expect(parsed.app).not.toHaveProperty('iconSm')` and `expect(parsed.app).not.toHaveProperty('userId')`.

src/commands/apps/apps.test.ts
```diff
@@ -1195,24 +1199,304 @@ describe('apps update --set-webhook-url', () => {
...
+        ])
+
+        expect(mockApi.updateApp).toHaveBeenCalledWith('9909', {
+            displayName: 'Renamed App',
+            oauthRedirectUri:
+                '["vscode://doist.todoist-vs-code/auth-complete","https://example.com/cb"]',
+        })
+    })
+
+    it('combines --name with --set-webhook-url and emits { app, webhook } JSON', async () => {
```
doistbot, line 1385: <a href="#">P2</a>  The combined record-patch-then-webhook sequence is new to this PR, and `update.ts` (Phase E comment) plus the PR description explicitly document the partial-failure contract: if `updateApp` succeeds and `updateAppWebhook` then throws, the record change stays persisted and the error propagates (no rollback). That contract is untested. The happy-path combined test right here uses the identical harness, so a failure variant is trivially scoped: mock `updateAppWebhook.mockRejectedValue(...)`, run `--name X --set-webhook-url Y --json`, and assert the rejection propagates while `updateApp` was still called. This guards against a future refactor reordering the calls or swallowing the webhook error — which would silently leave the app renamed but the webhook pointing at the stale URL while the command exits 0.

doistbot (commented):
This PR adds `--name` and `--description` to `td apps update` and lifts the one-flag-per-invocation restriction so all update flags combine in a single command, with up to two API calls (record patch then webhook swap).

<details>

- **<a href="#">P3</a> src/commands/apps/apps.test.ts:1281**: This test calls `captureConsole()` but doesn't capture the return value, so the human-readable success message is never asserted. The sibling `'sets the display name'` test (line 1258) captures `consoleSpy` and verifies the output contains `'Updated Todoist for VS Code'` and `'set name to …'`. The same pattern should apply here — verify `'Updated'` and `'set description'` in the output. The same gap exists in `'sets name and description together'` (line 1315) and `'combines --name with --add-oauth-redirect'` (line 1354).

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F431&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Add29c270-8070-11f1-8248-20d76d551e69%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Add29c270-8070-11f1-8248-20d76d551e69&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/output.ts
scottlovegrove: Done in 810fd7e — removed `projectEntity` and used the existing `processJsonItem` for the combined `{ app, webhook }` payload.

src/commands/apps/update.ts
scottlovegrove: Done in 810fd7e — `--name` / `--description` now compare against the resolved app and are treated as no-ops when unchanged (null and empty-string descriptions are considered equal), so an update with only matching values makes no `updateApp` request. The dry-run preview keys off the actual patch too.

src/commands/apps/apps.test.ts
scottlovegrove: Done in 810fd7e — added `expect(parsed.app).not.toHaveProperty('iconSm')` and `expect(parsed.app).not.toHaveProperty('userId')` to the combined-JSON test.

src/commands/apps/apps.test.ts
scottlovegrove: Done in 810fd7e — added a test that mocks `updateAppWebhook.mockRejectedValue(...)`, runs `--name X --set-webhook-url Y --json`, and asserts the rejection propagates while `updateApp` was still called.

PR #448, opened 2026-07-27
fix(skill): install gemini skill to Antigravity path

### LINKED ISSUE #423: td skill install gemini installs to obsolete path
`td skill install gemini` installs to .gemini/skills/todoist-cli/SKILL.md.

According to gemini itself:

does google antigravity pick up $HOME/.gemini/skills/todoist-cli/SKILL.md

The Short Answer
No, Google Antigravity will not automatically pick up your skill from that specific path.
Why It Doesn't Work
The directory ~/.gemini/skills/ is the standard global location natively used by the older Gemini CLI. With the launch of the Antigravity platform (which encompasses Antigravity 2.0, the IDE, and the Antigravity CLI), the ecosystem transitioned to a new configuration structure.
Antigravity natively looks for globally scoped skills in these directories instead:
•	~/.gemini/config/skills/
•	~/.gemini/antigravity/skills/
Because your todoist-cli skill is sitting in the legacy Gemini CLI directory, the Antigravity agent harness does not know it exists.

scottlovegrove:
Closes #423

## Problem

`td skill install gemini` wrote to `~/.gemini/skills/todoist-cli/SKILL.md`, the legacy Gemini CLI global skills directory. Gemini CLI has since been [transitioned to Antigravity CLI](https://developers.googleblog.com/an-important-update-transitioning-gemini-cli-to-antigravity-cli/) (it stopped serving Pro/Ultra/free users on 18 June 2026), and Antigravity resolves global skills from elsewhere — so the skill we installed was invisible to the agent.

Checked against the [Antigravity skills docs](https://antigravity.google/docs/skills) and [testing of every candidate path](https://atamel.dev/posts/2026/07-01_where_agy_agent_skills/):

| Global path | Antigravity | Antigravity CLI | Antigravity IDE |
| --- | --- | --- | --- |
| `~/.gemini/config/skills/` | ✅ | ✅ | ✅ |
| `~/.gemini/antigravity/skills/` | ✅ | ❌ | ✅ |
| `~/.gemini/antigravity-cli/skills/` | ❌ | ✅ | ✅ |
| `~/.gemini/skills/` (previous behaviour) | ❌ | ✅ | ✅ |

## Changes

- The `gemini` installer now targets `~/.gemini/config/skills/todoist-cli/SKILL.md`, the only global location all three Antigravity surfaces read. The agent key stays `gemini`, so `td skill install gemini` is unchanged.
- `createInstaller` gained an optional legacy global directory. Install and update write the new location and then delete the old file, pruning the emptied `todoist-cli/` and `skills/` directories.
- A skill that only exists at the legacy path counts as installed for update purposes, so existing installs migrate on their own — `postinstall` already runs `updateAllInstalledSkills`, so upgrading the CLI is enough. `install` and `update` print the removed legacy path.
- `uninstall` clears both locations and only errors when neither exists.
- `--local` is unchanged (`.gemini/skills/`) — it still serves enterprise Gemini Code Assist licences, which keep Gemini CLI support. Antigravity's workspace directory is `.agents/skills/`, already covered by `td skill install universal --local`; the README now says so.

## Testing

- New coverage in `skill.test.ts` for the Antigravity global path and the migration paths (install, update, uninstall from a legacy-only install, uninstall with neither present, and agents that have no legacy path), plus a legacy-only case in `update-installed.test.ts`.
- `npm test` (1736 passing), `npm run type-check`, `npm run check`, `npm run check:skill-sync` all pass.
- Manually verified against a sandboxed `$HOME` seeded with a legacy skill file: install writes `~/.gemini/config/skills/todoist-cli/SKILL.md`, removes `~/.gemini/skills/`, and uninstall cleans up afterwards.

### DIFF
--- a/src/lib/skills/create-installer.ts
+++ b/src/lib/skills/create-installer.ts
@@ -6,11 +6,13 @@ import { CliError } from '../errors.js'
 import { SKILL_COMPATIBILITY, SKILL_CONTENT, SKILL_DESCRIPTION, SKILL_NAME } from './content.js'
 import type { SkillInstaller } from './types.js'
 
-interface InstallerConfig {
+type InstallerConfig = {
     name: string
     description: string
     dirName: string
     globalDirName?: string
+    /** Directory an earlier version of the CLI installed the global skill to, cleaned up on install and update. */
+    legacyGlobalDirName?: string
 }
 
 export function generateSkillFile(): string {
@@ -28,6 +30,22 @@ metadata:
     return frontmatter + SKILL_CONTENT
 }
 
+async function removeEmptyDirs(startDir: string, levels: number): Promise<void> {
+    let dir = startDir
+    for (let level = 0; level < levels; level++) {
+        try {
+            const files = await readdir(dir)
+            if (files.length > 0) {
+                return
+            }
+            await rmdir(dir)
+        } catch {
+            return
+        }
+        dir = dirname(dir)
+    }
+}
+
 export function createInstaller(config: InstallerConfig): SkillInstaller {
     function getInstallPath(local: boolean): string {
         const base = local ? process.cwd() : homedir()
@@ -35,11 +53,50 @@ export function createInstaller(config: InstallerConfig): SkillInstaller {
         return join(base, dirName, 'skills', SKILL_NAME, 'SKILL.md')
     }
 
+    function getLegacyInstallPath(local: boolean): string | undefined {
+        if (local || !config.legacyGlobalDirName) {
+            return undefined
+        }
+        return join(homedir(), config.legacyGlobalDirName, 'skills', SKILL_NAME, 'SKILL.md')
+    }
+
+    async function hasLegacyInstall(local: boolean): Promise<boolean> {
+        const filepath = getLegacyInstallPath(local)
+        if (!filepath) {
+            return false
+        }
+        try {
+            await access(filepath)
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    async function removeLegacyInstall(local: boolean): Promise<void> {
+        const filepath = getLegacyInstallPath(local)
+        if (!filepath) {
+            return
+        }
+        try {
+            await unlink(filepath)
+        } catch {
+            return
+        }
+        // Prune the now-empty todoist-cli/ directory and, if nothing else lives
+        // there, the skills/ directory that contained it.
+        await removeEmptyDirs(dirname(filepath), 2)
+    }
+
     return {
         name: config.name,
         description: config.description,
 
         getInstallPath,
+
+        getLegacyInstallPath,
+
+        hasLegacyInstall,
 
         generateContent(): string {
             return generateSkillFile()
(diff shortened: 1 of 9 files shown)


doistbot (commented):
This PR moves the Gemini skill install path from the legacy `~/.gemini/skills/` to `~/.gemini/config/skills/`, the only Antigravity-resolved location that all three surfaces read, while cleaning up legacy files during install/update/uninstall so existing installs migrate automatically.

Few things worth tightening:
- Don't swallow all errors when deleting the legacy skill file — if the parent directory isn't writable, the unlink silently fails while `~/.gemini/skills/todoist-cli/SKILL.md` stays active for clients that still read it. Catch only the `ENOENT` race and propagate other unlink failures.
- In `removeEmptyDirs`, prefer calling `rmdir()` directly and stopping on `ENOTEMPTY` instead of `readdir()`-ing the directory first — it both checks and removes an empty dir without enumerating contents (which can be large for a shared `skills/` dir) and avoids an extra filesystem call.

<details>

- **<a href="#">P3</a> src/lib/skills/create-installer.ts:37**: `readdir()` loads every entry just to determine whether this directory is empty, including a potentially large shared `skills/` directory during install, update, and uninstall migrations. Call `rmdir()` directly and stop on `ENOTEMPTY` instead; it both checks and removes an empty directory without enumerating its contents or making an extra filesystem call.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F448&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A665e30f0-89de-11f1-99c9-bcac564dd556%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A665e30f0-89de-11f1-99c9-bcac564dd556&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/skills/create-installer.ts
```diff
@@ -28,19 +30,74 @@ metadata:
...
+    }
+
+    async function removeLegacyInstall(local: boolean): Promise<void> {
+        const filepath = getLegacyInstallPath(local)
+        if (!filepath) {
+            return
+        }
+        try {
+            await unlink(filepath)
+        } catch {
```
doistbot: <a href="#">P2</a>  Don't swallow all errors from deleting the legacy file. If the legacy skill's parent directory isn't writable, `td skill uninstall gemini` (and install/update migration) reports success even though `~/.gemini/skills/todoist-cli/SKILL.md` remains active for clients that still read it. Ignore only an `ENOENT` race, and propagate other unlink failures.

scottlovegrove: Done in 732d896 — legacy removal now ignores only ENOENT and rethrows everything else, so a failed unlink surfaces instead of reporting success. Added a test that makes the legacy directory read-only and asserts the update rejects.

Also applied the P3 note in the same commit: `removeEmptyDirs` calls `rmdir()` directly and stops on the first failure rather than `readdir()`-ing first.

PR #451, opened 2026-07-28
fix(auth): require explicit plaintext credential storage

### LINKED ISSUE #424: Headless Linux: OAuth can fall back to plaintext token when Secret Service has no login collection
## Summary

On a headless Linux VM, `td auth login` can complete OAuth successfully but silently fall back to persisting the resulting access token in plaintext when the Secret Service daemon is reachable but its default `login` collection cannot be created or prompted for.

The warning is emitted, but the default outcome is still a working write-capable credential stored in `~/.config/todoist-cli/config.json`.

## Reproduction

Environment:

- Todoist CLI: `1.76.1`
- Node: `v22.22.3`
- OS: Ubuntu 25.10, kernel `6.17.0-40-generic`
- Headless VM accessed over SSH
- `gnome-keyring-daemon` / `org.freedesktop.secrets` are running

1. Run `td auth login --no-browser-open --callback-port 8766` on the VM.
2. Use an SSH local port forward from a Mac so the browser callback reaches the VM loopback listener.
3. Complete OAuth in the Mac browser.
4. Observe successful OAuth login followed by:

   ```text
   Warning: system credential manager unavailable; access token saved as plaintext in ~/.config/todoist-cli/config.json
   ```

The underlying Secret Service failure is reproducible without touching a Todoist credential:

```text
secret-tool: ... does not exist at path "/org/freedesktop/secrets/collection/login"
```

The VM's keyring service logs also show that its graphical prompt cannot start in the headless session:

```text
couldn't initialize prompt: ... org.gnome.keyring.SystemPrompter exited with status 1
```

## Expected behavior

A headless/SSH environment should not unexpectedly leave a newly authorized write-capable OAuth token on disk in plaintext.

At minimum, the CLI should clearly identify this as a missing/unusable Secret Service collection and offer an actionable, non-secret-bearing remediation.

## Suggested directions

I understand that OS credential managers are a sensible default for desktop users. For headless Linux, could the CLI consider one or more of:

1. Require an explicit opt-in before writing an OAuth token to the plaintext fallback (for example, `--allow-plaintext-token`), rather than only emitting a warning after it has happened.
2. Detect the missing/unusable `login` collection specifically and explain that Secret Service exists but cannot create/unlock its default collection in the current session.
3. Offer a documented headless credential-store strategy or extension point (for example, a command-based secret provider), without requiring a particular third-party vault.

`TODOIST_API_TOKEN` is a useful escape hatch, but it moves the responsibility of safe runtime retrieval entirely to the caller and does not help users understand why a successful OAuth flow was persisted insecurely.

## Notes

- OAuth itself worked correctly; this is specifically a local credential-storage behavior.
- No OAuth codes, access tokens, client secrets, or authorization URLs are included here.

gnapse:
Closes https://github.com/Doist/todoist-cli/issues/424

## Summary

- Upgrade to `@doist/cli-core@1.1.0`, which owns the keyring storage policy.
- Default Todoist CLI credential writes to strict system storage: an unavailable credential manager fails the write without creating a plaintext config entry.
- Keep `--credential-store=plaintext` as the explicit opt-in for both OAuth and manual-token authentication; cli-core returns a warning for every successful plaintext credential write.

## Usage

Default, system-only storage:

```sh
td auth login
td auth token "YOUR_TOKEN"
```

Explicit plaintext storage:

```sh
td auth login --credential-store=plaintext
td auth token "YOUR_TOKEN" --credential-store=plaintext
# Warning: credential stored as plaintext in ~/.config/todoist-cli/config.json
```

## Verification

- Clean dependency install: `npm ci`
- `npm run build`
- `npm run type-check`
- Full test suite: 66 files, 1,740 tests
- `npm run check`
- Verified `td auth login --help` and `td auth token --help` show the new flag

### DIFF
--- a/src/commands/auth/token.ts
+++ b/src/commands/auth/token.ts
@@ -1,9 +1,16 @@
 import chalk from 'chalk'
 import { createApiForToken } from '../../lib/api/core.js'
-import { createTodoistTokenStore, toTodoistAccount } from '../../lib/auth-store.js'
+import {
+    createTodoistTokenStore,
+    type CredentialStore,
+    toTodoistAccount,
+} from '../../lib/auth-store.js'
 import { logTokenStorageResult, promptHiddenInput } from './helpers.js'
 
-export async function loginWithToken(token?: string): Promise<void> {
+export async function loginWithToken(
+    token?: string,
+    options: { credentialStore?: CredentialStore } = {},
+): Promise<void> {
     if (!token) {
         token = await promptHiddenInput('API token: ')
         if (!token.trim()) {
@@ -18,7 +25,7 @@ export async function loginWithToken(token?: string): Promise<void> {
     const probeApi = createApiForToken(trimmed)
     const user = await probeApi.getUser()
 
-    const store = createTodoistTokenStore()
+    const store = createTodoistTokenStore({ credentialStore: options.credentialStore })
     await store.set(
         toTodoistAccount({ id: user.id, email: user.email, authMode: 'unknown' }),
         trimmed,
--- a/README.md
+++ b/README.md
@@ -73,12 +73,18 @@ This opens your browser to authenticate with Todoist. Once approved, the token i
 - Windows: Credential Manager
 - Linux: Secret Service/libsecret
 
-If secure storage is unavailable, the CLI warns and falls back to `~/.config/todoist-cli/config.json`. Existing plaintext tokens are migrated automatically the next time the CLI reads them successfully from the config file.
+Secure storage is required by default. If it is unavailable, login fails without writing a credential to disk. To explicitly use plaintext config-file storage, pass `--credential-store=plaintext`; the CLI always prints a warning when it writes a credential this way. Existing plaintext tokens are migrated automatically the next time the CLI reads them successfully from the config file.
 
 For a read-only OAuth token (scope `data:read`), run:
 
 ```bash
 td auth login --read-only
+```
+
+To explicitly store a credential in plaintext instead of the OS credential manager:
+
+```bash
+td auth login --credential-store=plaintext
 ```
 
 In read-only mode, commands that change Todoist data (create/update/delete/complete/move/archive, etc.) are blocked by the CLI.
@@ -100,6 +106,9 @@ Run `td auth login --help` for the full list. When a command fails for lack of a
 ```bash
 td auth token "your-token"
 ```
+
+Manual tokens use the system credential manager by default too. To explicitly store one
+in plaintext, add `--credential-store=plaintext`; the CLI prints a warning after the write.
 
 **Environment variable:**
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@doist/cli-core": "1.0.1",
+        "@doist/cli-core": "1.1.0",
         "@doist/todoist-sdk": "12.0.0",
         "@napi-rs/keyring": "1.3.0",
         "@pnpm/tabtab": "0.5.4",
@@ -148,9 +148,9 @@
       }
     },
     "node_modules/@doist/cli-core": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/@doist/cli-core/-/cli-core-1.0.1.tgz",
-      "integrity": "sha512-tJamXPJt3LSA1sQHfC9IHdZbODJ5pVklqXY7NjbVI5cNhd0vCfSNb/gAQEOUD0+05oV/LByc757H7H0byVtjag==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@doist/cli-core/-/cli-core-1.1.0.tgz",
+      "integrity": "sha512-lSKRM2RmGkq8KUHAoBbFCLltgaCEgjkoUeTmfoDmqB279H6Al5fE3DpTV1zdyvkJkSTQW1onMYp7ZQClzY1PYA==",
       "license": "MIT",
       "dependencies": {
         "chalk": "5.6.2",
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "CHANGELOG.md"
   ],
   "dependencies": {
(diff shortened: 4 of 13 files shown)


doistbot (commented):
This PR hardens credential storage by defaulting to system storage and requiring explicit `--credential-store=plaintext` opt-in with warnings, covering login, token, and status store paths.

Few things worth tightening:
- The `--credential-store` flag on `td auth token` has no test coverage — the existing token tests mock `createTodoistTokenStore` without capturing the factory argument, so nothing verifies the flag actually reaches the store. Consider asserting the factory is called with `{ credentialStore: 'plaintext' }` when the flag is passed, matching the coverage already present for the login command.

<details>

- **<a href="#">P3</a> src/lib/auth-store.ts:213**: `clear` and `getLastClearResult` (plus the `lastClearResult` variable at L130) are pass-through wrappers with no plaintext branch — unlike `set`/`setBundle`/`getLastStorageResult`, which need the override for the plaintext path. `clear` just delegates to `inner.clear` and copies `inner.getLastClearResult()` into a local variable, duplicating methods that were inherited from `inner` via the `Object.create(inner)` prototype chain before this PR (the file's own JSDoc documents that design: "every method ... is inherited from inner via the prototype chain"). The `lastClearResult = undefined` reset before the call is moot because the only consumers (`logout.ts` `onCleared`, `user/remove.ts` `onRemoved`) read `getLastClearResult()` inside success callbacks that never run after a thrown clear. Dropping the `clear` wrapper, the `getLastClearResult` getter, and `lastClearResult` restores prototype inheritance and removes ~7 lines of redundant shadowing.
- **<a href="#">P3</a> src/commands/auth/login.ts:44**: Avoid maintaining `credentialStore` separately from Commander’s option state. The existing `loginCommand` forward reference already reads `browserOpen` at write time; have the store callback read `loginCommand?.opts().credentialStore` (with the system default) and pass `parseCredentialStore` directly to `.option()`, as `auth token` does. This removes the duplicate mutable state and custom parser wrapper.
- **<a href="#">P3</a> src/lib/auth-store.ts:164**: `storePlaintextBundle` calls `userRecords.upsert()` (read+write) then `promoteDefaultIfNeeded()` which calls `getDefaultId()` (read) and, if no default exists, `setDefaultId()` (read+write). That's 3 `readConfig()` + 2 `writeConfig()` when the first write already landed. Since `upsertStoredUser` and `setDefaultUser` from `users.ts` are pure config transforms, they could be composed into a single read-modify-write cycle: read config once, apply both mutations, write once. The plaintext path is opt-in and login is infrequent, so this is low priority — but the fix is straightforward and would keep the write path I/O-minimal.
- **<a href="#">P3</a> src/commands/auth/login.test.ts:194**: These two tests only verify Commander's internal option parsing (`program.commands[0]?.opts().credentialStore`), not the actual credential storage routing behavior. They depend on Commander's internal command structure and duplicate what Commander already validates. The behavioral coverage for `--credential-store` routing lives in `auth-store-credential.test.ts`; these tests add maintenance cost without meaningful regression protection. Consider dropping them or replacing with a test that exercises the `store` object passed to `attachLoginCommand` to verify the correct storage backend is used.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F451&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad34e8f80-8ab9-11f1-897e-6bfb67407459%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad34e8f80-8ab9-11f1-897e-6bfb67407459&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/token.ts
```diff
@@ -18,7 +25,7 @@ export async function loginWithToken(token?: string): Promise<void> {
     const probeApi = createApiForToken(trimmed)
     const user = await probeApi.getUser()
 
-    const store = createTodoistTokenStore()
+    const store = createTodoistTokenStore({ credentialStore: options.credentialStore })
```
doistbot, line 28: <a href="#">P2</a>  The `--credential-store` option added to `td auth token` (index.ts) and threaded through `loginWithToken` here has no test. The existing token tests in `auth.test.ts` mock `createTodoistTokenStore: () => holder.store`, discarding the factory argument, so nothing verifies `options.credentialStore` actually reaches `createTodoistTokenStore`. A break in this wiring silently falls back to system storage with no test failing — the `td auth token "$TOKEN" --credential-store=plaintext` contract documented in README/SKILL.md would be broken. The login command's equivalent flag is covered in `login.test.ts`; the token path should match. Consider asserting the factory is called with `{ credentialStore: 'plaintext' }` when the flag is passed (the mock would need to capture the arg).

scottlovegrove (commented):
I wonder if this is something that should have been added to cli-core? This way all the other CLIs would have it available too.

gnapse:
> I wonder if this is something that should have been added to cli-core? This way all the other CLIs would have it available too.

Makes sense. Will do it, and request your review there. Once approved and released, will change this PR to be about bring in that update.

scottlovegrove (commented):
👍🏻

rmartins90:
✅

PR #454, opened 2026-07-30
feat(view): add --include-children to task view and project view

scottlovegrove:
Ports [Doist/todoist-mcp#543](https://github.com/Doist/todoist-mcp/pull/543) to the CLI.

`td task view` returned on `--json` before it ever looked up subtasks, so machine consumers — the skill, and any agent driving `td` — saw nothing but a raw `parentId`, while human output gave a bare `Subtasks: N active` with no ids to act on. `td project view` mentioned neither a parent nor sub-projects in any mode. The practical consequence is the one from the MCP issue: a dated parent can hide an undated subtask that no date filter will surface, and there was no cheap way to notice.

`--include-children` now lists the direct children of a task or project and flags each one that has children of its own, in human and `--json` output on both commands, and in `--ndjson` on `project view` (`task view` has no NDJSON mode today — that is a pre-existing gap, not one this flag introduces). Without the flag every output — and every round trip — is byte-identical to before.

```
$ td task view id:87 --include-children
Plan party

ID:       87
Priority: p2
Project:  Personal
Subtasks: 3 active
  id:88  Book venue
  id:89  Send invites  due: Aug 2
  id:90  Chase caterer  ▸
URL:      https://app.todoist.com/app/task/87
```

## How children are read

**Subtasks** — one `getTasks({ parentId, limit: 25 })` call, then one cheap `limit: 1` probe per child to set the nesting flag (the API carries no child count on a task). The cost tracks the size of the answer, not the size of the project: a leaf task costs exactly one extra call, which is the common case. I deliberately did not scan every task in the project and group by `parentId` in memory — that is unbounded in project size and sequentially paginated.

**Sub-projects** — projects cannot be filtered by parent server-side, so the hierarchy comes from a single `loadPersonalProjects()` fetch grouped by `parentId`, which gives exact nesting flags for free. Workspace projects nest under folders rather than under a parent project, so they report none with no lookup at all.

Shared logic lives in a new `src/lib/children.ts` (mirroring the MCP module) with the two renderers in `output.ts`; `buildProjectChildren` is kept pure so the caller can reuse the fetched list for the parent lookup and unit-test it without an API mock.

## Behaviour worth knowing

- The project `Parent:` line is shown **whether or not** the flag is passed, mirroring the parent line `task view` already prints. That is the one behaviour change outside the flag: viewing a sub-project now costs one `getProject` call, and it is free when `--include-children` already loaded the list.
- A failed lookup reports `childrenError` and leaves the view intact — it neither fails the command nor omits the count silently, since an absent count reads as "no children", the exact mistake the flag exists to prevent. A failed nesting probe costs only that child's flag; the listing survives.
- Only **active** subtasks and **non-archived** sub-projects are counted. The list caps at 25; past that `hasMoreChildren` is set and the output points at `td task list --parent id:<id> --all`.
- Accessible mode substitutes `(has subtasks)` / `(has sub-projects)` for the `▸` glyph rather than decorating it, following `formatProgressBar`.

## Testing

New `src/lib/children.test.ts` plus cases across `output.test.ts`, `task.test.ts` and `project.test.ts` — 35 new tests, including guards that the flag-off path makes no extra calls and that `--json` without the flag has no `childCount`. Full suite passes (1772 tests), plus `type-check`, `check` and `check:skill-sync`. Rendering was verified against the built `dist` output for every branch (nesting, accessible, truncated, partial-probe, failed, none).

Skill content is updated and `skills/todoist-cli/SKILL.md` regenerated.


src/commands/task/index.ts
```diff
@@ -65,6 +65,10 @@ Examples:
         .option('--json', 'Output as JSON')
         .option('--full', 'Include all fields in output')
         .option('--raw', 'Disable markdown rendering')
+        .option(
```
doistbot, line 68: <a href="#">P2</a>  `task view` still does not register an `--ndjson` option, and `viewTask` has no NDJSON branch. As a result, `td task view id:... --include-children --ndjson` is rejected as an unknown option despite the feature promising NDJSON output. Add the option and emit the parent task plus projected child records.

src/commands/project/view.ts
```diff
@@ -16,13 +30,62 @@ async function printDescription(description: string | undefined, raw: boolean):
...
+    }
+}
+
+/** The already-projected sub-project records for `--ndjson`, if any were requested. */
+function childNdjsonRecords(
+    children: ChildrenResult<ProjectChild> | undefined,
+    options: ViewOptions,
+): object[] {
+    if (!children) return []
+    const records = processChildrenJson(
```
doistbot: <a href="#">P2</a>  This extracts only `children`, discarding `childCount`, `hasMoreChildren`, and `childrenError` for both project NDJSON paths. Consequently, an NDJSON consumer cannot distinguish no children from a failed lookup or know that the 25 emitted children are truncated. Emit these fields with the parent record or in a metadata record (as the paginated NDJSON formatter does).

src/commands/project/view.ts
```diff
@@ -16,13 +30,62 @@ async function printDescription(description: string | undefined, raw: boolean):
...
+            ? { childCount: 0, children: [] }
+            : await resolveChildren(async () => {
+                  allPersonal = await loadPersonalProjects(api)
+                  return buildProjectChildren(project, allPersonal)
+              })
+    }
+
+    // Free when the children lookup already loaded the list; one cheap call otherwise.
+    let parentProject: Project | undefined
+    if (!isWorkspaceProject(project) && project.parentId) {
```
doistbot: <a href="#">P2</a>  `parentProject` is resolved here unconditionally, before the `--json`/`--ndjson` early returns, but it's only consumed by `printParent` in the two human-readable paths (lines 150 and 248). In `--json` or `--ndjson` mode (detailed or not) without `--include-children`, viewing a personal sub-project fires `api.getProject(parentId)` whose result is discarded — a new round trip that didn't exist before and breaks the flag-off byte-identical-round-trip guarantee for machine consumers. Guard the lookup with `!options.json && !options.ndjson`, or move it to just before the `printParent` calls.

src/commands/project/view.ts
```diff
@@ -16,13 +30,62 @@ async function printDescription(description: string | undefined, raw: boolean):
...
 ): Promise<void> {
     const api = await getApi()
     const project = await resolveProjectRef(api, ref)
 
+    let allPersonal: PersonalProject[] | undefined
+    let children: ChildrenResult<ProjectChild> | undefined
+    if (options.includeChildren) {
+        children = isWorkspaceProject(project)
+            ? { childCount: 0, children: [] }
+            : await resolveChildren(async () => {
```
doistbot: <a href="#">P2</a>  Don't await the paginated project-tree lookup before starting the baseline view request. In non-JSON modes, `getFullProject()` or `getTasks()` is independent of `loadPersonalProjects()`, but isn't issued until every project page has loaded. Keep the children lookup as a promise and await it only when rendering so `--include-children` doesn't add the hierarchy-fetch latency serially.

src/commands/task/view.ts
```diff
@@ -1,15 +1,37 @@
...
 
-export async function viewTask(ref: string, options: ViewOptions): Promise<void> {
+export async function viewTask(
+    ref: string,
+    options: ViewOptions & { includeChildren?: boolean },
+): Promise<void> {
     const api = await getApi()
     const task = await resolveTaskRef(api, ref)
 
+    const children = options.includeChildren
```
doistbot: <a href="#">P2</a>  This await serializes the direct-child query and all nesting probes before the independent `getProjects()` and potential parent `getTask()` calls below. For human output with `--include-children`, start the children lookup as a promise and await it alongside those lookups; keep the current direct await only for the JSON early-return path.

doistbot (commented):
Adds an `--include-children` flag to `task view` and `project view` that lists direct children and flags nested ones, across human, `--json`, and `--ndjson` output, with shared logic in a new `src/lib/children.ts` module.

<details>

- **<a href="#">P3</a> src/commands/project/view.ts:60**: The `as object[] | undefined` cast is forced by `processChildrenJson` returning `Record<string, unknown>`. Since the `children` property is always `object[] | undefined` when it exists, the output helper in `src/lib/output.ts` could return a more specific shape — e.g. `{ childCount?: number; children?: object[]; hasMoreChildren?: boolean; childrenError?: string }` — eliminating the cast here without breaking the spread usage elsewhere.
- **<a href="#">P3</a> src/lib/output.ts:126**: The `due:` rendering in task child rows is uncovered. Every child fixture in the `formatChildrenBlock` and `formatTaskView` tests uses `fixtures.tasks.child`, which has `due: null`, so the `if (task.due)` branch in `formatChildRow` is never exercised — even though it's featured in the PR example output (`id:89 Send invites due: Aug 2`). Add a child with a non-null `due` to one `formatChildrenBlock` case to cover it.
- **<a href="#">P3</a> src/commands/project/project.test.ts:2139**: This test guards the no-flag `Parent:` line and the single `getProject` call, but not that the full project list is left unloaded on that path. The PR states viewing a sub-project costs one `getProject` call without `--include-children`; adding `expect(mockApi.getProjects).not.toHaveBeenCalled()` here would protect that contract, in the same spirit as the `task view` --json-untouched guard. Without it, a refactor that loads `loadPersonalProjects()` on every view would pass silently.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F454&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aea303b20-8c14-11f1-8d7f-58f3b27a51e4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aea303b20-8c14-11f1-8d7f-58f3b27a51e4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/project/view.ts
scottlovegrove: Done in 2349f36 — the project record in --ndjson now carries childCount, hasMoreChildren and childrenError (the children stay as their own lines). Covered by tests for the truncated, failed and no-flag cases.

src/commands/project/view.ts
scottlovegrove: Done in 2349f36 — the parent lookup moved to just before each printParent call, so --json and --ndjson no longer fire it. Added a test asserting getProject is called exactly once in both machine modes.

src/commands/project/view.ts
scottlovegrove: Done in 2349f36 — the children lookup is now started rather than awaited, and awaited alongside getFullProject / getTasks via Promise.all.

src/commands/task/view.ts
scottlovegrove: Done in 2349f36 — same treatment: the lookup starts immediately and is awaited alongside getProjects and the parent getTask.

src/commands/task/index.ts
scottlovegrove: Not changing this: `task view` has never had `--ndjson` — that is a pre-existing gap in the command's output modes, not something this flag introduces, and adding a whole output mode to `task view` belongs in its own change. The PR description was wrong to say NDJSON across the board; corrected to name `project view` specifically.

PR #456, opened 2026-07-31
feat(task): add explicit --first-due

### LINKED ISSUE #342: Recurring task created via `due_string` with a `starting` clause never advances on completion
## Summary

A task created with a `due_string` that contains a `starting` clause has the **entire string stored verbatim as its recurrence rule**. Completing the task then never advances it — every completion re-pins the same date. The identical recurrence *without* `starting` advances correctly.

Todoist's quick-add parser (the app's "Add task" box, and `td task quickadd`) handles the same text correctly: it consumes `starting <date>` into the due date and stores a clean recurrence rule (`every! 2 weeks`). So Todoist has two natural-language date parsers — the `due_string` field parser and the quick-add parser — and they disagree on `starting`.

This is **server-side** — it reproduces with raw REST, no `td` involved (see below). Filing it here because `td task add --due` is likely a common way to hit it. `td task quickadd` is **not** affected.

## Impact

Silent. A recurring task created via the API looks completely normal; the breakage only surfaces the first time you complete it — the task doesn't move, and never will. Anything that creates recurring tasks programmatically — `td task add --due`, the SDK, direct REST, bulk imports/migrations — is exposed.

## Reproduce — `td` CLI (v1.63.0)

```console
$ td task add "repro-cli" --due "every! 2 weeks starting 2026-05-17"
$ td task view "repro-cli" --json | jq -c .due
{"isRecurring":true,"string":"every! 2 weeks starting 2026-05-17","date":"2026-05-17","timezone":null,"lang":"en"}

$ td task complete "repro-cli"
$ td task view "repro-cli" --json | jq -c .due
{"isRecurring":true,"string":"every! 2 weeks starting 2026-05-17","date":"2026-05-17","timezone":null,"lang":"en"}
#                                                                  ^^^^^^^^^^^^^^^^^^ date UNCHANGED — did not advance
```

Control — `td task quickadd` parses the identical text correctly:

```console
$ td task quickadd "repro-qa every! 2 weeks starting 2026-05-17"
$ td task view "repro-qa" --json | jq -c .due
{"isRecurring":true,"string":"every! 2 weeks","date":"2026-05-17","timezone":null,"lang":"en"}
#                            ^^^^^^^^^^^^^^^^ "starting" consumed into the date; clean recurrence rule stored
```

That task advances correctly on completion.

## Reproduce — raw REST (no `td`; isolates it to the API)

```bash
TOKEN=...            # a Todoist API token
TODAY=$(date +%F)

# CASE 1 — due_string WITH a "starting" clause
ID=$(curl -s -X POST https://api.todoist.com/api/v1/tasks \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d "{\"content\":\"repro-A\",\"due_string\":\"every! 2 weeks starting $TODAY\"}" | jq -r .id)
curl -s -X POST "https://api.todoist.com/api/v1/tasks/$ID/close" -H "Authorization: Bearer $TOKEN"
curl -s "https://api.todoist.com/api/v1/tasks/$ID" -H "Authorization: Bearer $TOKEN" | jq -c .due
#   => due date is UNCHANGED after completion
curl -s -X DELETE "https://api.todoist.com/api/v1/tasks/$ID" -H "Authorization: Bearer $TOKEN"

# CASE 2 — identical recurrence WITHOUT "starting" (control)
ID=$(curl -s -X POST https://api.todoist.com/api/v1/tasks \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"content":"repro-B","due_string":"every! 2 weeks"}' | jq -r .id)
curl -s -X POST "https://api.todoist.com/api/v1/tasks/$ID/close" -H "Authorization: Bearer $TOKEN"
curl -s "https://api.todoist.com/api/v1/tasks/$ID" -H "Authorization: Bearer $TOKEN" | jq -c .due
#   => due date advanced by 2 weeks
curl -s -X DELETE "https://api.todoist.com/api/v1/tasks/$ID" -H "Authorization: Bearer $TOKEN"
```

Observed (run 2026-05-17):

```
CASE 1  due_string sent: "every! 2 weeks starting 2026-05-17"
  POST /tasks       -> stored due.string = "every! 2 weeks starting 2026-05-17"   due.date = 2026-05-17
  POST .../close    -> HTTP 204
  GET  /tasks/{id}  -> due.string = "every! 2 weeks starting 2026-05-17"   due.date = 2026-05-17   << UNCHANGED

CASE 2  due_string sent: "every! 2 weeks"
  POST /tasks       -> stored due.string = "every! 2 weeks"                due.date = 2026-05-17
  POST .../close    -> HTTP 204
  GET  /tasks/{id}  -> due.string = "every! 2 weeks"                       due.date = 2026-05-31   << advanced +2 weeks
```

Same endpoint, same auth, same completion call — the only variable is the `starting` clause.

## Where the fault is

- `td task add --due "X"` sends `X` verbatim as the task's `due_string`. Confirmed: the created task's `due.string` comes back byte-identical to the input — the CLI and SDK do not transform it.
- The raw-REST repro reproduces the bug with `td` entirely out of the loop, so the fault is **server-side**: the `due_string` parser on `POST /api/v1/tasks` stores a `starting` clause *inside* the recurrence rule instead of resolving it into the due date.
- The quick-add parser (the app's Add-task box, `td task quickadd`) resolves `starting` correctly. The two parsers should agree.

## Expected

`POST /api/v1/tasks` with `due_string: "every! 2 weeks starting <date>"` should behave like quick-add: set the first due date to `<date>`, store the recurrence rule as `every! 2 weeks`, and advance normally on completion.

## Workaround

Create the task with a bare recurrence, then position the first occurrence separately:

```bash
td task add "task" --due "every! 2 weeks"    # created due today
td task reschedule "task" 2026-06-20          # moves the first occurrence; recurrence preserved
```

`td task reschedule` keeps the recurrence rule intact, and the task advances correctly on completion thereafter.

## Versions

- `td` 1.63.0 (`@doist/todoist-cli`)
- Endpoint: `POST https://api.todoist.com/api/v1/tasks`
- Reproduces equally via direct REST with no `td` in the path (above)
- macOS 26.5

gnapse:
Adds an explicit --first-due option for task add and update, avoiding implicit parsing of natural-language due strings.

Closes #342

### DIFF
--- a/src/commands/task/add.ts
+++ b/src/commands/task/add.ts
@@ -1,4 +1,4 @@
-import { getApi } from '../../lib/api/core.js'
+import { getApi, rescheduleTask as rescheduleTaskSync } from '../../lib/api/core.js'
 import { resolveAssigneeId } from '../../lib/collaborators.js'
 import { CliError } from '../../lib/errors.js'
 import { isQuiet } from '../../lib/global-args.js'
@@ -13,7 +13,7 @@ import {
 } from '../../lib/refs.js'
 import { readStdin } from '../../lib/stdin.js'
 import { parsePriority } from '../../lib/task-list.js'
-import { applyDuration, type DurationArgs } from './helpers.js'
+import { applyDuration, type DurationArgs, validateFirstDueDate } from './helpers.js'
 
 export interface AddOptions {
     content: string
@@ -32,18 +32,18 @@ export interface AddOptions {
     order?: number
     json?: boolean
     dryRun?: boolean
+    firstDue?: string
 }
 
 export async function addTask(options: AddOptions): Promise<void> {
+    const firstDue = validateFirstDueDate(options.firstDue, options.due)
     const api = await getApi()
 
     const args: Parameters<typeof api.addTask>[0] = {
         content: options.content,
     }
 
-    if (options.due) {
-        args.dueString = options.due
-    }
+    if (options.due) args.dueString = options.due
 
     if (options.deadline) {
         args.deadlineDate = options.deadline
@@ -141,6 +141,7 @@ export async function addTask(options: AddOptions): Promise<void> {
             Content: args.content,
             Description: args.description,
             Due: args.dueString,
+            'First occurrence': firstDue,
             Deadline: args.deadlineDate ?? undefined,
             Priority: options.priority,
             Project: options.project,
@@ -155,7 +156,29 @@ export async function addTask(options: AddOptions): Promise<void> {
         return
     }
 
-    const task = await api.addTask(args)
+    let task = await api.addTask(args)
+
+    if (firstDue) {
+        if (!task.due) {
+            throw new CliError(
+                'FIRST_DUE_NOT_APPLIED',
+                `Task "${task.id}" was created without a due date, so its first occurrence could not be set.`,
+            )
+        }
+        try {
+            await rescheduleTaskSync(task.id, firstDue, task.due)
+        } catch (error) {
+            throw new CliError(
+                'FIRST_DUE_NOT_APPLIED',
+                `Task "${task.id}" was created, but its first occurrence could not be set.`,
+                [
+                    `Recover with: td task reschedule id:${task.id} ${firstDue}`,
+                    `Underlying error: ${(error as Error).message}`,
+                ],
+            )
+        }
+        task = await api.getTask(task.id)
+    }
 
     if (options.json) {
         console.log(formatJson(task, 'task'))
--- a/src/commands/task/task.test.ts
+++ b/src/commands/task/task.test.ts
@@ -706,6 +706,95 @@ describe('task add', () => {
         expect(consoleSpy).toHaveBeenCalledWith('Due: tomorrow')
     })
 
+    it('creates recurring task with --first-due that advances on completion', async () => {
+        const program = createProgram()
+        captureConsole()
+        const initialDue = {
+            date: '2026-07-31',
+            string: 'every! 2 weeks',
+            isRecurring: true,
+            timezone: null,
+            lang: 'en',
+        }
+
+        mockApi.addTask.mockResolvedValue({ id: 'task-new', content: 'Task', due: initialDue })
+        mockApi.getTask.mockResolvedValue({
+            id: 'task-new',
+            content: 'Task',
+            due: { ...initialDue, date: '2026-05-17' },
+        })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'task',
+            'add',
+            'Task',
+            '--due',
+            'every! 2 weeks',
+            '--first-due',
+            '2026-05-17',
+        ])
+
+        expect(mockApi.addTask).toHaveBeenCalledWith(
+            expect.objectContaining({ dueString: 'every! 2 weeks' }),
+        )
+        expect(mockRescheduleTask).toHaveBeenCalledWith('task-new', '2026-05-17', initialDue)
+        expect(mockApi.getTask).toHaveBeenCalledWith('task-new')
+    })
+
+    it('passes due strings through unchanged without --first-due', async () => {
+        const program = createProgram()
+        captureConsole()
+
+        mockApi.addTask.mockResolvedValue({ id: 'task-new', content: 'Task', due: null })
+
+        await program.parseAsync([
+            'node',
+            'td',
+            'task',
+            'add',
+            'Task',
+            '--due',
+            'every! 2 weeks starting 2026-05-17',
+        ])
+
+        expect(mockApi.addTask).toHaveBeenCalledWith(
+            expect.objectContaining({ dueString: 'every! 2 weeks starting 2026-05-17' }),
+        )
+        expect(mockRescheduleTask).not.toHaveBeenCalled()
+    })
+
+    it('rejects --first-due without --due before creating a task', async () => {
+        const program = createProgram()
+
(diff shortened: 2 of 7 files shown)


src/commands/task/add.ts
```diff
@@ -155,7 +156,29 @@ export async function addTask(options: AddOptions): Promise<void> {
         return
     }
 
-    const task = await api.addTask(args)
+    let task = await api.addTask(args)
+
+    if (firstDue) {
```
doistbot, line 161: <a href="#">P2</a>  The `firstDue` reschedule + error-handling + `getTask` block (L161–L181) is duplicated almost identically in `update.ts` (L112–L132). The only differences are the variable name (`task` vs `updated`) and the verb in the error messages (`created`/`updated`). Extract into a shared helper in `task/helpers.ts` — e.g. `applyFirstDue(api, taskId, firstDue, due, verb)` — alongside the existing `validateFirstDueDate` and `applyDuration` helpers that already serve both commands.

src/commands/task/update.ts
```diff
@@ -104,7 +107,29 @@ export async function updateTask(ref: string, options: UpdateOptions): Promise<v
...
+            throw new CliError(
+                'FIRST_DUE_NOT_APPLIED',
+                `Task "${updated.id}" was updated, but its first occurrence could not be set.`,
+                [
+                    `Recover with: td task reschedule id:${updated.id} ${firstDue}`,
+                    `Underlying error: ${(error as Error).message}`,
+                ],
+            )
+        }
+        updated = await api.getTask(updated.id)
```
doistbot, line 131: <a href="#">P2</a>  This extra `getTask` call is only needed for `--json`: the normal update output reads `updated.content`, and quiet mode emits nothing. Every `task update --first-due` without `--json` now performs an unnecessary sequential API request. Fetch the task only when serializing JSON.

src/commands/task/task.test.ts
```diff
@@ -706,6 +706,95 @@ describe('task add', () => {
         expect(consoleSpy).toHaveBeenCalledWith('Due: tomorrow')
     })
 
+    it('creates recurring task with --first-due that advances on completion', async () => {
```
doistbot, line 709: <a href="#">P2</a>  The error-recovery path when `rescheduleTaskSync` fails is untested for both `task add` (L168–179 in add.ts) and `task update` (L119–129 in update.ts). When the reschedule call throws, the CLI wraps it in a `CliError` with a recovery hint (`td task reschedule id:… <date>`) and the underlying error message. This is user-visible behavior — the recovery command format is effectively a contract. A test that makes `mockRescheduleTask.mockRejectedValueOnce(new Error('API error'))` and asserts the thrown error message and hints would protect that path.

doistbot (commented):
This PR adds an explicit `--first-due` flag to `task add` and `task update`, leveraging the existing `rescheduleTask` infrastructure with thorough validation and recovery guidance.

Few things worth tightening:

- Extract the shared first-due reschedule + error-handling + `getTask` block into a helper in `task/helpers.ts` — it's duplicated nearly verbatim between `add.ts` and `update.ts`, differing only in a variable name and verb.
- Gate the extra `getTask` call behind `--json` in `update.ts`, since the normal and quiet output paths don't need it — every non-JSON `task update --first-due` currently triggers an unnecessary sequential API request.
- Add tests covering the `rescheduleTaskSync` failure path in both commands to protect the recovery-command format and error wrapping behavior.

<details>

- **<a href="#">P3</a> src/commands/task/add.ts:1**: The import `rescheduleTask as rescheduleTaskSync` uses the `Sync` suffix for an `async` function, which is misleading. There is no local name collision in this file (the command function is `addTask`), so the alias is unnecessary — import `rescheduleTask` directly.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F456&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A64a4a8c0-8d0d-11f1-89be-eb7982d9a23a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A64a4a8c0-8d0d-11f1-89be-eb7982d9a23a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

gnapse:
No longer needed, see https://github.com/Doist/todoist-cli/issues/342#issuecomment-5146302011

PR #463, opened 2026-08-06
feat(filter): expose view options (board/list, group by, sort)

### LINKED ISSUE #462: feat: expose view options (board/list, group by, sort) — SDK supports view_options_set but the CLI never calls it
## Problem

Filters and projects can be created and updated from the CLI, but their **presentation** cannot. `td filter create` accepts only `--name`, `--query`, `--color` and `--favorite`, so there is no way to set the view mode (list/board/calendar), grouping, or sorting that the Todoist apps expose through the *View* menu.

This matters for any shared-project workflow. A concrete example: a household/team overview filter

```
(p1 | p2 | !no deadline)
```

is only useful rendered as a **board grouped by assignee** — as a flat list it is just 37 unordered tasks. The query is scriptable today; the view that makes it readable is not.

## The plumbing already exists

`@doist/todoist-sdk@13.0.0` (already a dependency) fully types the command:

```ts
// dist/types/types/sync/commands/view-options.d.ts
export type ViewOptionsSetArgs = {
    viewType: ViewType
    objectId?: string
    groupedBy?: GroupedBy | null
    filteredBy?: string | null
    viewMode?: ViewMode
    showCompletedTasks?: boolean
    sortedBy?: SortedBy | null
    sortOrder?: SortOrder | null
}
```

and `view_options_set` is present in `SyncCommandsMap`. The CLI already drives the Sync command interface in this exact shape — `src/lib/api/filters.ts` does:

```ts
await api.sync({ commands: [createCommand('filter_add', { ... }, tempId)] })
```

So this looks like a thin wiring gap rather than new capability.

## Current workaround

Callers have to bypass `td` and hand-roll the Sync request:

```bash
curl https://api.todoist.com/api/v1/sync \
  -H "Authorization: Bearer $TOKEN" \
  -d commands='[{"type":"view_options_set","uuid":"<uuid>","args":{
      "view_type":"FILTER","object_id":"<filter_id>",
      "view_mode":"BOARD","grouped_by":"ASSIGNEE",
      "sorted_by":"PRIORITY","sort_order":"DESC"}}]'
```

### Casing gotcha worth surfacing in the CLI

The enums must be **UPPERCASE**. The public API reference at `developer.todoist.com/api/v1` documents this command with lowercase values in its request example (`"view_mode": "board"`, `"grouped_by": "assignee"`, `"view_type": "filter"`), which fails:

```json
{"error": "Invalid argument value", "error_code": 20,
 "error_extra": {"expected": "view_mode: Input should be 'LIST', 'BOARD' or 'CALENDAR'; ..."}}
```

The SDK constants are the correct, uppercase source of truth and are re-exported from the package root, so flag validation could import them directly rather than restating the values:

```js
const { VIEW_MODES, GROUPED_BY_OPTIONS } = require('@doist/todoist-sdk')
// [ 'LIST', 'BOARD', 'CALENDAR' ]
// [ 'ASSIGNEE', 'ADDED_DATE', 'DUE_DATE', 'DEADLINE', 'LABEL', 'PRIORITY', 'PROJECT', 'WORKSPACE' ]
```

A CLI flag layer would hide the docs/API discrepancy from users entirely — arguably the strongest argument for exposing it here.

## Proposed surface

**Option A — flags on the existing commands** (smallest diff, mirrors `--color`/`--favorite`):

```bash
td filter create --name "Team" --query "(p1 | p2)" \
    --view-mode board --group-by assignee --sort-by priority --sort-order desc
td filter update "Team" --view-mode list
td project update "Roadmap" --view-mode board --group-by assignee
```

**Option B — a dedicated subcommand**, since view options also apply to labels and to the built-in `today` / `upcoming` views, which have no object to hang flags off:

```bash
td view-options set --filter "Team" --mode board --group-by assignee
td view-options set --today --group-by priority
td view-options show --filter "Team"
```

Option A covers the common cases with less surface; Option B is the only one that reaches `TODAY`/`UPCOMING`. Lowercase flag values accepted and normalised to the SDK's uppercase enums either way.

I have no stake in which shape you prefer — happy to open a PR with tests for whichever you'd take, or to leave it if this is intentionally out of scope.

## Environment

- `td` 3.1.3 (also checked 3.1.4 — not present)
- `@doist/todoist-sdk` 13.0.0
- macOS

tardio-clockero:
Implements **Option A** from #462 — flags on the existing commands, since that was the smallest surface that covers the common cases. Happy to reshape into a dedicated subcommand (Option B) if you'd rather reach `TODAY`/`UPCOMING` too; the parsing and API layers are already separate from the filter commands, so moving them is cheap.

## What this adds

```bash
td filter create --name "Team" --query "(p1 | p2 | !no deadline)" \
    --view-mode board --group-by assignee --sort-by priority --sort-order desc

td filter update "Team" --view-mode list
```

Four new flags on `filter create` and `filter update`: `--view-mode`, `--group-by`, `--sort-by`, `--sort-order`.

## Notes on the implementation

- **Wiring, not new capability.** `@doist/todoist-sdk` already types `view_options_set` and re-exports `VIEW_MODES` / `GROUPED_BY_OPTIONS` / `SORTED_BY_OPTIONS` / `SORT_ORDERS`. Allowed values are derived from those constants rather than restated, so the CLI can't drift from the SDK.
- **Casing is handled for the user.** The Sync API requires uppercase enums and rejects anything else with `error_code: 20`, while the API reference documents lowercase. Flags accept lowercase kebab-case and normalise (`due-date` → `DUE_DATE`). Invalid values fail before any request with the allowed list echoed back.
- **`update` with only view flags no longer trips `NO_CHANGES`**, and skips the `filter_update` call when nothing about the filter itself changed — one command, one request.
- `src/lib/api/view-options.ts` takes a `viewType`/`objectId` rather than being filter-specific, so `project`/`label` can reuse it.
- View options are stored per user; the doc comment says so, since it is easy to assume they travel with a shared project.

## Verification

- `npm run check`, `npm run type-check`, `npm run build` — clean
- `npm test` — 1784 passing (5 added)
- Exercised end-to-end against the live API: after `filter create ... --view-mode board --group-by assignee --sort-by priority --sort-order desc`, a `view_options` sync read returns

  ```json
  {"view_type": "FILTER", "object_id": "...", "view_mode": "BOARD",
   "grouped_by": "ASSIGNEE", "sorted_by": "PRIORITY", "sort_order": "DESC"}
  ```

- Dry-run and validation output:

  ```
  $ td filter create --name "Team" --query today --view-mode board --group-by assignee --dry-run
  [dry-run] Would create filter:
    Name: Team
    Query: today
    View: viewMode=board, groupedBy=assignee

  $ td filter create --name X --query today --group-by colour
  Error: INVALID_OPTIONS
  Invalid value "colour" for --group-by.
    - Allowed: assignee, added-date, due-date, deadline, label, priority, project, workspace
  ```

Closes #462

### DIFF
--- a/src/commands/filter/filter.test.ts
+++ b/src/commands/filter/filter.test.ts
@@ -12,7 +12,12 @@ vi.mock('../../lib/api/filters.js', () => ({
     deleteFilter: vi.fn(),
 }))
 
+vi.mock('../../lib/api/view-options.js', () => ({
+    setViewOptions: vi.fn(),
+}))
+
 import { addFilter, deleteFilter, fetchFilters, updateFilter } from '../../lib/api/filters.js'
+import { setViewOptions } from '../../lib/api/view-options.js'
 import { setupApiMock } from '../../test-support/api-mock.js'
 import { makeFilter } from '../../test-support/fixtures.js'
 import { type MockApi } from '../../test-support/mock-api.js'
@@ -22,6 +27,7 @@ const mockFetchFilters = vi.mocked(fetchFilters)
 const mockAddFilter = vi.mocked(addFilter)
 const mockUpdateFilter = vi.mocked(updateFilter)
 const mockDeleteFilter = vi.mocked(deleteFilter)
+const mockSetViewOptions = vi.mocked(setViewOptions)
 
 function createProgram() {
     return createTestProgram(registerFilterCommand)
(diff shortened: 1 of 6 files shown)


src/commands/filter/index.ts
```diff
@@ -28,6 +28,16 @@ export function registerFilterCommand(program: Command): void {
         .option('--query <query>', 'Filter query (required, e.g., "today | overdue")')
         .option('--color <color>', 'Filter color')
         .option('--favorite', 'Mark as favorite')
+        .option('--view-mode <mode>', 'Layout: list, board or calendar')
```
scottlovegrove, line 31: [P2] These new view-option flags need adding to `SKILL_CONTENT` in `src/lib/skills/content.ts`, then regenerate `skills/todoist-cli/SKILL.md` with `npm run sync:skill`. The installed agent skill is the command reference, and right now it documents `filter create` / `filter update` with no way to discover `--view-mode`, `--group-by`, `--sort-by` or `--sort-order`.

scottlovegrove, line 31: [P3] Worth centralising the registration of these four flags. The same definitions are duplicated for `create` and `update`, and the accepted values are separately derived from the SDK arrays in `view-options.ts`, so a future change to the options can easily leave one help block stale. Extracting a shared registration helper and deriving the choice descriptions from the same option metadata would keep them in step.

src/commands/filter/update.ts
```diff
@@ -14,14 +20,15 @@ export interface UpdateOptions {
 
 export async function updateFilterCmd(nameOrId: string, options: UpdateOptions): Promise<void> {
     const filter = await resolveFilterRef(nameOrId)
+    const viewOptions = parseViewOptionFlags(options)
```
scottlovegrove, line 23: [P2] Parse the view flags before resolving the filter. An invalid value currently triggers `resolveFilterRef`, which fetches the entire filter list, before failing locally — e.g. `td filter update Work --group-by colour` performs a Sync request that was never needed.

src/commands/filter/update.ts
```diff
@@ -32,11 +39,18 @@ export async function updateFilterCmd(nameOrId: string, options: UpdateOptions):
             Query: args.query,
             Color: args.color,
             Favorite: args.isFavorite !== undefined ? String(args.isFavorite) : undefined,
+            View: viewOptions ? describeViewOptions(viewOptions) : undefined,
         })
         return
     }
 
-    await updateFilter(filter.id, args)
+    if (Object.keys(args).length > 0) {
```
scottlovegrove, line 47: [P2] When both filter fields and view flags are supplied this fires two independent Sync requests serially, adding a full network round trip to every combined update. Batching `filter_update` and `view_options_set` into a single `api.sync` call (or at least running them concurrently) would avoid that.

src/commands/filter/filter.test.ts
```diff
@@ -821,3 +827,123 @@ describe('filter --dry-run', () => {
         expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining('Would update filter'))
     })
 })
+
+describe('filter view options', () => {
```
scottlovegrove, line 831: [P2] Could we add dry-run coverage with a view flag, for both create and update, asserting `setViewOptions` is not called? The existing dry-run tests only exercise the filter fields, so a regression that persisted view settings during `--dry-run` would pass while quietly changing the user's layout.

scottlovegrove (changes requested):
Thanks for creating the PR, just a few small things to take care of

PR #472, opened 2026-08-10
fix: support sections in filter views

### LINKED ISSUE #274: Todoist view filter doesn't show tasks when filter has sections
if you type `td filter view [ref]` for a filter that has sections in the query no tasks are returned but if you pick a filter that has no sections it works

gnapse:
## Summary

- split saved comma-separated filters into API-supported section queries
- fetch sections concurrently and preserve section order in human, JSON, and NDJSON output
- add regression coverage and update the agent command reference

Closes #274

## Why

Todoist saved filters use commas to display queries as separate lists. However, the public [Get Tasks By Filter API documentation](https://developer.todoist.com/api/v1/#tag/Tasks/operation/get_tasks_by_filter_api_v1_tasks_filter_get) states: “Multiple filters (using the comma `,` operator) are not supported.”

The CLI passed the full saved query to that endpoint, which returned a successful empty result. This change splits the saved query and calls the endpoint once per section.

## Behavior and deliberate simplifications

- Each non-empty section has an independent API request and cursor stream.
- Sections run concurrently through a worker pool capped at six active section streams. This limits request bursts while keeping common filters fully parallel.
- Pages within one section remain sequential because each page needs the previous cursor.
- Results preserve the saved section order even if responses complete out of order.
- Parsing is deliberately limited to finding unescaped commas. It is not a full Todoist query parser.
- A comma preceded by an odd number of backslashes is preserved as a literal. Escapes remain in the subquery sent to the API.
- Quotes and parentheses do not protect unescaped commas. Todoist documents an unescaped comma as the separate-list operator and backslash as the escape mechanism.
- Empty fragments, such as the middle fragment in `today,,tomorrow`, are ignored.
- Results are not deduplicated. The same task can intentionally appear in more than one section.
- If any section request fails, the whole command fails instead of returning output that looks complete.
- Empty multi-section results do not load unrelated project or collaborator data.
- `--limit` applies independently to each section.
- `--cursor` is rejected for multi-section filters because one cursor cannot represent several independent streams. Use `--all`, or query one section directly. Structured output includes `nextCursor` per section.
- Multi-section JSON uses `{ sections: [{ query, results, nextCursor }] }`. NDJSON emits one section object per line. Single-section output stays unchanged.

## Validation

- Reproduced with an isolated saved filter containing three comma-separated queries. Before the fix, the saved filter returned zero tasks while each component query returned one task.
- Re-ran the same saved filter with the local build. Human, JSON, and NDJSON output each returned three sections with one expected task per section.
- Removed the temporary filter, project, and tasks after validation.

## Tests

- `npm run type-check`
- `npm run check`
- `npm run check:skill-sync`
- `npm test` — 1,795 tests pass

### DIFF
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -259,6 +259,8 @@ td section delete id:123 --yes
 td section browse id:123
 ```
 
+Saved filters can contain multiple comma-separated queries. `td filter view` shows each query as a separate section and applies `--limit` to each section. Under `--json`, multi-section filters return `{ sections: [{ query, results, nextCursor }] }`; under `--ndjson`, each line is one section with the same fields. Because each section has its own pagination cursor, use `--all` instead of `--cursor` for multi-section filters.
+
 Shared labels can appear in `td label list` and `td label view`, but standard update and delete actions only work for labels with IDs. Use `td label rename-shared` and `td label remove-shared` for shared labels.
 
 ### Comments, Attachments, Notifications, And Reminders
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -259,7 +259,7 @@ td section delete id:123 --yes
 td section browse id:123
 ```
 
-Saved filters can contain multiple comma-separated queries. `td filter view` shows each query as a separate section and applies `--limit` to each section. Under `--json`, multi-section filters return `{ sections: [{ query, results, nextCursor }] }`; under `--ndjson`, each line is one section with the same fields. Because each section has its own pagination cursor, use `--all` instead of `--cursor` for multi-section filters.
+Saved filters can contain multiple comma-separated queries, each displayed as a separate filter section. `td filter view` preserves those sections and applies `--limit` to each one. Under `--json`, multi-section filters return `{ sections: [{ query, results, nextCursor }] }`; under `--ndjson`, each line is one section with the same fields. Because each section has its own pagination cursor, use `--all` instead of `--cursor` for multi-section filters.
 
 Shared labels can appear in `td label list` and `td label view`, but standard update and delete actions only work for labels with IDs. Use `td label rename-shared` and `td label remove-shared` for shared labels.
 
--- a/src/commands/filter/view.ts
+++ b/src/commands/filter/view.ts
@@ -22,6 +22,8 @@ interface FilterSection {
     nextCursor: string | null
 }
 
+const FILTER_SECTION_CONCURRENCY = 6
+
 /**
  * Split Todoist's multi-list filter syntax without reimplementing its query parser.
  * Todoist defines each unescaped comma as a list separator and uses backslashes
@@ -55,13 +57,7 @@ function formatFilterSectionsJson(
     full = false,
     showUrls = false,
 ): string {
-    return formatJson({
-        sections: sections.map((section) => ({
-            query: section.query,
-            results: section.results.map((task) => processJsonItem(task, 'task', full, showUrls)),
-            nextCursor: section.nextCursor,
-        })),
-    })
+    return formatJson({ sections: processFilterSections(sections, full, showUrls) })
 }
 
 function formatFilterSectionsNdjson(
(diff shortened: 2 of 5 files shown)


doistbot (commented):
This PR fixes support for Todoist's comma-separated saved filters by splitting them into individual API requests per section, with concurrent fetching and order-preserving output across human, JSON, and NDJSON formats.

Few things worth tightening:
- Skip the project/cache lookup in the multi-section path when all sections return no tasks — currently an empty multi-section filter still calls `getProjects()`, and if that unrelated request fails, the command errors instead of cleanly rendering the empty results.

<details>

- **<a href="#">P3</a> src/commands/filter/filter.test.ts:507**: Reuse `fixtures.tasks.basic` (with per-test overrides) for these new task responses instead of hand-building partial task objects. The shared fixture supplies a valid SDK task shape and keeps test data consistent; the same applies to the new literals in this test block.
- **<a href="#">P3</a> src/commands/filter/view.ts:79**: `formatFilterSectionsJson` and `formatFilterSectionsNdjson` share an identical `sections.map((section) => ({ query, results: section.results.map(task => processJsonItem(...)), nextCursor }))` body — only the outer wrapper (`formatJson({ sections })` vs `formatNdjson(...)`) differs. Extract the shared mapping into a small helper (e.g. `mapFilterSections(sections, full, showUrls)`) so the two callers don't drift apart if the section payload shape changes.
- **<a href="#">P3</a> src/commands/filter/view.ts:134**: Move this flattening below the JSON/NDJSON early returns. Structured output only reads `sections`, so this creates an unused second array of every task; with `--all` on a multi-section filter that can be a large allocation.
- **<a href="#">P3</a> src/commands/filter/view.ts:107**: `Promise.all` fans out all section requests concurrently with no bound. A filter with many comma-separated queries — say, 20+ — would fire that many concurrent API calls, each of which may paginate further. While uncommon in practice, adding a modest concurrency cap (e.g., 5–10) with a simple semaphore or a library like `p-limit` would prevent request bursts without meaningfully hurting latency for the common case of 2–5 sections.
- **<a href="#">P3</a> src/commands/filter/filter.test.ts:630**: No test covers the documented behavior that empty fragments like the middle fragment in `today,,tomorrow` are ignored. The exported `splitFilterQueries` function explicitly skips empty trimmed sections (lines 37-38 in view.ts), and the PR description calls this out as intentional. A naive refactor (e.g., using `query.split(',')`) would regress silently — users would see empty section headers in their output. Consider adding a unit test for `splitFilterQueries` covering empty fragments and/or the odd-backslash escape rule, since the function is exported and has documented edge cases not exercised by the integration tests.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F472&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5ec56320-9507-11f1-81d4-b8f677ff8f93%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5ec56320-9507-11f1-81d4-b8f677ff8f93&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/filter/view.ts
```diff
@@ -80,9 +173,11 @@ export async function showFilter(nameOrId: string, options: PaginatedViewOptions
     console.log('')
 
     if (tasks.length === 0) {
-        console.log('No tasks match this filter.')
-        console.log(formatNextCursorFooter(nextCursor))
-        return
+        if (!hasMultipleSections) {
```
doistbot, line 194: <a href="#">P2</a>  Don’t fall through to project fetching when every section is empty. Unlike the single-section path, a multi-section filter with no matches still calls `getProjects()` even though rendering section messages and cursors needs no project data; if that unrelated request fails, the command reports an error instead of the empty filter result. Skip project/cache loading when `tasks.length === 0` and render the empty sections directly.

gnapse: Fixed in 261cd98. Project and collaborator data now load only when at least one section contains tasks. Added a regression test that verifies an all-empty multi-section filter renders both empty sections without calling getProjects().

skills/todoist-cli/SKILL.md
```diff
@@ -259,6 +259,8 @@ td section delete id:123 --yes
 td section browse id:123
 ```
 
+Saved filters can contain multiple comma-separated queries. `td filter view` shows each query as a separate section and applies `--limit` to each section. Under `--json`, multi-section filters return `{ sections: [{ query, results, nextCursor }] }`; under `--ndjson`, each line is one section with the same fields. Because each section has its own pagination cursor, use `--all` instead of `--cursor` for multi-section filters.
```
scottlovegrove:
```suggestion
Saved filters can contain multiple comma-separated queries, each comma separation is a filter section. `td filter view` shows each query as a separate section and applies `--limit` to each section. Under `--json`, multi-section filters return `{ sections: [{ query, results, nextCursor }] }`; under `--ndjson`, each line is one section with the same fields. Because each section has its own pagination cursor, use `--all` instead of `--cursor` for multi-section filters.
```

scottlovegrove (approved):

PR #479, opened 2026-08-17
feat(filter): sort tasks the way the Todoist apps do

### LINKED ISSUE #473: feat: sorting via views is not working by default
Excited to use the new easier to use command for filters, but noticing there is no sorting capability. 

Right now I test it with this filter `##work & p4 & !subtask` this is my `Someday Work` it doesn't use sections at all but i noticed that it is set for default sorting but when you run the command `td filter list "someday work" --all` I do not see it sort in the same manner.

seems the default ordering is not followed. I changed my todoist filter to be sorted by date added asc and that seems to match what the CLI wants to do by default. Would be good if it followed what todoist does unless that is more of a frontend change and if so then maybe just passing a param for the sorting.

craigcarlyle:
## Summary
- `td filter view` now orders tasks the way the Todoist apps order them, instead of printing whatever order the API returned
- Reads the sorting saved on the filter's view with `api.getViewOptions()` and orders the results with `sortTasks`, both from `@doist/todoist-sdk@14.1.0`
- Falls back to Todoist's default hierarchy when the view has no saved sort, which is what "Manual (default)" means in the app
- Adds `--sort default|priority|date|deadline|date-added|name|project|assignee|workspace|none` and `--sort-order asc|desc`, resolved flag first, then saved view, then Todoist default
- `--sort none` returns the raw API order
- Sorts each section of a comma-separated filter on its own, and returns `--json` / `--ndjson` in the same order as the pretty output
- Prints the resolved sort in the view header, under `Query:` and `URL:`

## Why the order was wrong
- Todoist doesn't sort server-side. `GET /tasks/filter` returns storage order, roughly date added ascending, which is why the reporter found that setting their filter to "date added asc" made the app match the CLI
- Every client sorts locally, by `sorted_by` on the view and then by a [documented default](https://www.todoist.com/help/articles/default-sorting-order-for-todoist-tasks-mqmgerY7)
- The CLI read neither piece

## What stays in the CLI
@gnapse put the comparators and the view-options reader in the SDK (Doist/todoist-sdk-typescript#665 and #666), so this PR carries only what `sortTasks` asks its caller for:

- `buildProjectOrder`, which lays the sidebar out as the `projectOrder` map the SDK wants. It has no helper for this, so every consumer builds its own
- `queryUsesDates`, which decides between the priority-first and date-first hierarchies from the filter query
- The vocabulary `--sort` speaks, and the mapping from a saved view onto it

The ordering tests here cover that handoff. The comparators are tested in the SDK, so re-testing them through this wrapper would only break on his refactors.

## Not addressed
- `grouped_by` is still ignored, so a filter grouped by date or label renders flat. That overlaps with #462

## Test plan
- [x] 1861 tests pass, plus type-check, lint, format and SKILL.md sync, all against the published 14.1.0
- [x] Manual: a filter whose view is saved with `sorted_by: PRIORITY` renders p1 first
- [x] Manual: default-sorted filter with date queries leads with date, priority breaks the tie
- [x] Manual: `--sort none` reproduces the pre-fix order
- [x] Manual: `--sort assignee` across a workspace filter with 13 assignees sorts A-Z with unassigned last, and `--sort-order desc` reverses it

Closes #473

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   ],
   "dependencies": {
     "@doist/cli-core": "1.2.0",
-    "@doist/todoist-sdk": "14.0.2",
+    "@doist/todoist-sdk": "14.1.0",
     "@napi-rs/keyring": "1.3.0",
     "@pnpm/tabtab": "0.5.4",
     "chalk": "6.0.0",
--- a/src/commands/filter/view.ts
+++ b/src/commands/filter/view.ts
@@ -245,6 +245,7 @@ export async function showFilter(nameOrId: string, options: FilterViewOptions): 
     }
 
     let sections: FilterSection[]
+    const fetchLimit = sort.field === 'none' ? targetLimit : Number.MAX_SAFE_INTEGER
 
     try {
         // Section cursors are independent; pages within one section remain serial.
@@ -257,7 +258,7 @@ export async function showFilter(nameOrId: string, options: FilterViewOptions): 
                         cursor: cursor ?? undefined,
                         limit,
                     }),
-                { limit: targetLimit, startCursor: options.cursor },
+                { limit: fetchLimit, startCursor: options.cursor },
             )
             return { query, ...result }
         })
@@ -310,7 +311,10 @@ export async function showFilter(nameOrId: string, options: FilterViewOptions): 
                               projects: projectMap,
                           }) ?? task.responsibleUid)
                         : null,
-            }),
+            }).slice(0, targetLimit),
+            // An API cursor belongs to the API's order, not this local one.
+            // The complete result set is already loaded before sorting.
+            nextCursor: null,
         }))
     }
 
@@ -351,14 +355,7 @@ export async function showFilter(nameOrId: string, options: FilterViewOptions): 
     console.log(chalk.bold(`${filter.name}`))
     console.log(chalk.dim(`Query: ${filter.query}`))
     console.log(chalk.dim(`URL:   ${filterUrl(filter.id)}`))
-    const truncated = sections.some((section) => section.nextCursor)
-    const sortNotes = [
-        saved.unavailable ? 'saved view options unavailable' : null,
-        // The order can only be right across the tasks we actually hold. A task
-        // still behind a cursor can outrank everything on screen.
-        sorting && truncated ? `ordered over the ${tasks.length} tasks fetched, use --all` : null,
-    ].filter(Boolean)
-    const sortNote = sortNotes.length > 0 ? ` (${sortNotes.join('; ')})` : ''
+    const sortNote = saved.unavailable ? ' (saved view options unavailable)' : ''
     console.log(chalk.dim(`Sort:  ${formatTaskSort(sort)}${sortNote}`))
     console.log('')
 
--- a/src/commands/filter/view.ts
+++ b/src/commands/filter/view.ts
@@ -1,7 +1,15 @@
+import {
+    findViewOptions,
+    isWorkspaceProject,
+    type TodoistApi,
+    type ViewOptions as SavedViewOptions,
+} from '@doist/todoist-sdk'
 import chalk from 'chalk'
-import { getApi, type Project, type Task } from '../../lib/api/core.js'
+import { getAccountTimezone, getApi, type Project, type Task } from '../../lib/api/core.js'
+import { fetchWorkspaces } from '../../lib/api/workspaces.js'
 import { CollaboratorCache, formatAssignee } from '../../lib/collaborators.js'
 import { CliError } from '../../lib/errors.js'
+import { getLogger } from '../../lib/logger.js'
 import type { PaginatedViewOptions } from '../../lib/options.js'
 import {
     formatNextCursorFooter,
@@ -13,8 +21,29 @@ import {
     processJsonItem,
 } from '../../lib/output.js'
 import { LIMITS, paginate } from '../../lib/pagination.js'
+import { fetchProjects } from '../../lib/task-list.js'
+import {
+    buildProjectOrder,
+    defaultDirectionFor,
+    formatTaskSort,
+    parseTaskSortDirection,
+    parseTaskSortField,
+    queryUsesDates,
+    sortNeedsCollaborators,
+    sortNeedsProjects,
+    sortTasks,
+    type TaskSort,
+    type TaskSortDirection,
+    type TaskSortField,
+    taskSortFromViewOptions,
+} from '../../lib/task-sort.js'
 import { filterUrl } from '../../lib/urls.js'
 import { resolveFilterRef } from './helpers.js'
+
+export interface FilterViewOptions extends PaginatedViewOptions {
+    sort?: string
+    sortOrder?: string
+}
 
 interface FilterSection {
     query: string
@@ -98,9 +127,91 @@ async function mapFilterSections(
     return sections
 }
 
-export async function showFilter(nameOrId: string, options: PaginatedViewOptions): Promise<void> {
-    const filter = await resolveFilterRef(nameOrId)
+/**
+ * Flag beats saved view, saved view beats the Todoist default. `--sort-order`
+ * on its own re-points the direction of whichever field is already in play.
+ */
+function resolveSort(
+    filterId: string,
+    requested: { field?: TaskSortField; direction?: TaskSortDirection },
+    viewOptions: SavedViewOptions[],
+): TaskSort {
+    if (requested.field) {
+        return {
+            field: requested.field,
+            direction: requested.direction ?? defaultDirectionFor(requested.field),
+        }
+    }
+
+    const saved = taskSortFromViewOptions(
+        findViewOptions(viewOptions, {
+            viewTypes: ['FILTER', 'WORKSPACE_FILTER'],
+            objectId: filterId,
+        }),
+    )
+    return requested.direction ? { field: saved.field, direction: requested.direction } : saved
+}
+
+/**
+ * A view-options failure shouldn't cost you the list, the way a collaborator
+ * lookup failure doesn't cost you a task's name. It does change what the
+ * ordering means though, so the caller says so in the header rather than
+ * passing off the default hierarchy as the view's own sorting.
+ */
+async function loadViewOptions(
+    api: TodoistApi,
+): Promise<{ viewOptions: SavedViewOptions[]; unavailable: boolean }> {
+    try {
+        return { viewOptions: await api.getViewOptions(), unavailable: false }
+    } catch (error) {
+        getLogger().detail('failed to load saved view options', {
+            error: error instanceof Error ? error.message : String(error),
+        })
+        return { viewOptions: [], unavailable: true }
+    }
+}
+
+/**
+ * Workspace names order the workspace buckets the way the sidebar and
+ * `td project list` do. Only worth a request when more than one workspace is
+ * represented, since a single bucket has nothing to sort against.
+ */
+async function loadWorkspaceNames(
+    projects: Map<string, Project>,
+): Promise<Map<string, string> | undefined> {
+    const workspaceIds = new Set<string>()
+    for (const project of projects.values()) {
+        if (isWorkspaceProject(project)) workspaceIds.add(project.workspaceId)
+    }
+    if (workspaceIds.size < 2) return undefined
+
+    try {
+        const workspaces = await fetchWorkspaces()
+        return new Map(workspaces.map((workspace) => [workspace.id, workspace.name]))
+    } catch (error) {
+        getLogger().detail('failed to load workspaces for sort order', {
+            error: error instanceof Error ? error.message : String(error),
+        })
+        return undefined
+    }
+}
+
+export async function showFilter(nameOrId: string, options: FilterViewOptions): Promise<void> {
+    const requested = {
+        field: options.sort ? parseTaskSortField(options.sort) : undefined,
+        direction: options.sortOrder ? parseTaskSortDirection(options.sortOrder) : undefined,
+    }
     const api = await getApi()
+
+    // An explicit --sort makes the saved view options moot, so only pay for
+    // them when they can still change the order.
+    const [filter, saved] = await Promise.all([
+        resolveFilterRef(nameOrId),
+        requested.field
+            ? Promise.resolve({ viewOptions: [] as SavedViewOptions[], unavailable: false })
+            : loadViewOptions(api),
+    ])
+    const sort = resolveSort(filter.id, requested, saved.viewOptions)
 
     const targetLimit = options.all
         ? Number.MAX_SAFE_INTEGER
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -62,7 +62,7 @@
     "marked": "18.0.9",
     "marked-terminal-renderer": "2.2.0",
     "oauth4webapi": "3.8.7",
-    "open": "11.0.1"
+    "open": "11.0.0"
```
doistbot: <a href="#">P2</a>  This downgrades `open` from `11.0.1` to `11.0.0`, and the lockfile drags its transitive deps back with it (`powershell-utils` ^0.2.0→^0.1.0, `wsl-utils` ^1.0.0→^0.3.0, resolved `wsl-utils` 1.0.0→0.3.1). It's unrelated to the sorting feature and looks like a stale-base revert. Rebase onto current `main` so this PR doesn't ship a dependency downgrade.

src/commands/filter/view.ts
```diff
@@ -148,6 +227,40 @@ export async function showFilter(nameOrId: string, options: PaginatedViewOptions
...
+    // Do not deduplicate: Todoist can intentionally show one task in multiple lists.
+    const tasks = sections.flatMap((section) => section.results)
+    const isPretty = !options.json && !options.ndjson
+
+    // Rendering needs the projects and collaborators for every task; sorting
+    // needs them only for some fields. Fetch when either side asks, never when
+    // there is nothing to order or draw.
+    const projectMap =
+        tasks.length > 0 && (isPretty || sortNeedsProjects(sort.field))
+            ? await fetchProjects(api)
```
doistbot: <a href="#">P2</a>  `fetchProjects()` performs a single `getProjects()` request and drops `nextCursor`. When the API paginates projects, tasks in later project pages have no entries in `projectOrder` or `workspaceOrder`, so default, project, and workspace sorting cannot match the sidebar order. Fetch all project pages before building the order map.

src/commands/filter/view.ts
```diff
@@ -98,10 +125,62 @@ async function mapFilterSections(
...
+
+/**
+ * Saved view options are a display nicety: when they can't be read, the list
+ * should still render. Falls back to an empty set, which lands the caller on
+ * Todoist's default ordering, and leaves a breadcrumb behind `-vv`.
+ */
+async function loadViewOptions(api: TodoistApi): Promise<SavedViewOptions[]> {
+    try {
+        return await api.getViewOptions()
+    } catch (error) {
```
doistbot, line 166: <a href="#">P2</a>  Don't convert every `getViewOptions()` failure into an empty saved-sort list. A transient API/auth failure now makes a filter with a saved sort render successfully in the Todoist default order, which is indistinguishable from an intentional default sort unless the user enables `-vv`. Let the request error propagate; an actual empty response already handles the "no saved sort" case.

doistbot (commented):
`td filter view` now sorts tasks to match the Todoist apps, reading the saved view sort via the SDK and adding `--sort`/`--sort-order` flags, with graceful fallbacks and JSON/NDJSON parity.

Few things worth tightening:
- Rebase onto current `main` — this branch downgrades the `open` dependency (and its transitive deps) and reverts version metadata, both unrelated to the feature.
- `fetchProjects()` only fetches the first page; when projects are paginated, later projects won't appear in `projectOrder`/`workspaceOrder`, breaking default/project/workspace sorts. Fetch all pages before building the order map.
- Don't swallow `getViewOptions()` errors into an empty list — a transient API/auth failure silently renders a saved sort in default order. Let the error propagate; a genuinely empty response already handles the no-saved-sort case.
- `queryUsesDates` only strips the first word of named references, so a project like `#due date` leaves `date` behind and triggers the wrong hierarchy. Consume the full named-reference operand before applying the date pattern.
- `--sort none` still clones every section array through `sortTasks`; skip the sort block entirely when the field is `none` to avoid the unnecessary copy.

<details>

- **<a href="#">P3</a> src/lib/task-sort.ts:57**: This comment says `default` has no field, but `SORTED_BY_FIELD['default']` is `'MANUAL'` (via `FIELD_BY_SORTED_BY.MANUAL === 'default'`); only `none` is absent. A reader would expect `sortTasks` to pass `sortedBy: null` for the default sort when it actually passes `'MANUAL'`. Update the comment to reflect the actual mapping.
- **<a href="#">P3</a> src/commands/filter/filter.test.ts:1107**: The test name says the view options "cannot be read," but this mocks `getViewOptions` resolving to `[]` (a valid empty response), so `loadViewOptions`' catch branch — the logging + `return []` fallback — is never exercised. Use `mockApi.getViewOptions.mockRejectedValue(new Error('boom'))` to actually cover that path, or rename the test to describe the empty-response case.
- **<a href="#">P3</a> src/lib/task-sort.test.ts:266**: These assertions cover a branch that can't be reached through the CLI: `--sort`/`--sort-order` are registered with `withCaseInsensitiveChoices`, which sets Commander `choices`, so invalid values are rejected during parsing before `parseTaskSortField`/`parseTaskSortDirection` ever run. The `CliError` throws are dead code in the command path; either drop the defensive validation (and this test) or note that it's only guarding the exported helper, not user-facing behavior.
- **<a href="#">P3</a> src/lib/task-sort.test.ts:276**: `sortNeedsProjects` and `sortNeedsCollaborators` are one-line predicates (`field !== 'none'` and `field === 'assignee'`) whose observable behavior — whether `getProjects` / collaborators are fetched — is already verified by the integration tests in filter.test.ts ('fetches projects in JSON mode' and 'skips the project fetch when nothing is sorted'). These unit tests re-state the implementation without adding regression signal.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F479&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A62ba0ba0-9c07-11f1-8c3c-295c033a0124%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A62ba0ba0-9c07-11f1-8c3c-295c033a0124&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/task-sort.ts
```diff
@@ -0,0 +1,331 @@
...
+ * Names and free-text searches can contain date words ("#May launch",
+ * "search: due diligence"), so those operands are dropped before the query is
+ * inspected. A `search:` term runs to the next boolean operator or list comma.
+ */
+function stripNamedRefs(query: string): string {
+    return query
+        .replace(/"[^"]*"/g, ' ')
+        .replace(/'[^']*'/g, ' ')
+        .replace(/\bsearch\s*:[^&|(),]*/gi, ' ')
+        .replace(/[#@/]{1,2}[^\s&|()!,]+/g, ' ')
```
doistbot: <a href="#">P2</a>  This only removes the first word of a project/label reference. For example, a filter targeting a project named `#due date` leaves `date` in the query, so `queryUsesDates()` selects the date-first default hierarchy even though the filter contains no date condition. Consume the complete named-reference operand (through the next boolean/list delimiter, respecting escapes) before applying the date pattern.

src/commands/filter/view.ts
```diff
@@ -148,6 +227,40 @@ export async function showFilter(nameOrId: string, options: PaginatedViewOptions
...
+    const collaboratorCache = new CollaboratorCache()
+    if (tasks.length > 0 && (isPretty || sortNeedsCollaborators(sort.field))) {
+        await collaboratorCache.preload(api, tasks, projectMap)
+    }
+
+    // Each section is its own list in the Todoist apps, sorted on its own.
+    const projectOrder = buildProjectOrder(projectMap.values())
+    sections = sections.map((section) => ({
+        ...section,
+        results: sortTasks(section.results, sort, {
```
doistbot: <a href="#">P2</a>  `--sort none` still runs every section through `sortTasks`, whose `none` branch clones the entire result array. Thus `td filter view ... --all --sort none` retains a second array of references for every fetched task even though raw order needs no transformation. Skip this `sections.map` sorting block when `sort.field === 'none'` (and leave the fetched sections intact).

package.json
craigcarlyle: My fault, and worse than a stale rebase: a `reset --soft` while collapsing this from a two-PR stack captured `package.json`, `package-lock.json` and `CHANGELOG.md` from before 3.2.1 shipped, so the commit reverted the version and took `open` back to 11.0.0 with it.

Restored all three from `main` and re-applied just the SDK pin. The lockfile diff is down to the `@doist/todoist-sdk` version and its integrity hash.

src/commands/filter/view.ts
craigcarlyle: Correct, and it predates this PR: `fetchProjects` has always read one page. That was survivable when the map only supplied project names, and it stopped being survivable when sorting started reading it for `projectOrder` and `workspaceOrder`.

It pages to the end now. An account inside one page still costs one request; the second only happens when the response carries a cursor. `today` and the task-list views get the fix for free, since they share the helper.

src/commands/filter/view.ts
craigcarlyle: Half agreed, and I've taken the half I agree with. Failing the command on a transient sync error seems worse than the disease: you lose the whole list to fetch a display preference, and this codebase already leans the other way, with `children.ts` explicitly never throwing and an unresolvable collaborator degrading to a raw uid.

What was wrong is that the fallback was silent. The header now reads `Sort: Todoist default (saved view options unavailable)` when the read fails, so nothing passes the default hierarchy off as the view's own sorting, and the test asserts it.

src/lib/task-sort.ts
craigcarlyle: Good catch, that one bites. The pattern ended a name at whitespace, so `#due date` left `date` in the string and flipped a priority-first filter to date-first.

Names now run to the next operator, which is where Todoist ends them. Covered by three cases: `#due date` and `#due date & p1` stay priority-first, `#due date & today` is still date-first.

src/commands/filter/view.ts
craigcarlyle: Fixed, though the copy itself stays. An earlier review asked `sortTasks` to stop returning the caller's array, and that's still the right contract for a shared helper. What was wasteful was the command rebuilding every section to get it, so the whole sort block is skipped when the field is `none`.

src/lib/task-sort.ts
```diff
@@ -0,0 +1,337 @@
...
+        {
+            sortedBy: SORTED_BY_FIELD[sort.field] ?? null,
+            sortOrder: sort.direction === 'desc' ? 'DESC' : 'ASC',
+            defaultOrder: context.dateDriven ? 'DATE_FIRST' : 'PRIORITY_FIRST',
+        },
+        {
+            projectOrder: context.projectOrder,
+            workspaceOrder: context.workspaceOrder,
+            assigneeName: context.assigneeName,
+            timezone: localTimezone(),
```
gnapse: Could we use the Todoist account timezone here instead of the machine timezone? The SDK treats this value as the account timezone, and the two can differ in containers, remote sessions, or user settings. I reproduced an ordering reversal near midnight with the machine in Santiago and the account context in Los Angeles.

src/commands/filter/view.ts
```diff
@@ -148,6 +232,42 @@ export async function showFilter(nameOrId: string, options: PaginatedViewOptions
...
+    if (tasks.length > 0 && (isPretty || sortNeedsCollaborators(sort.field))) {
+        await collaboratorCache.preload(api, tasks, projectMap)
+    }
+
+    // Each section is its own list in the Todoist apps, sorted on its own.
+    if (sort.field !== 'none') {
+        const projectOrder = buildProjectOrder(projectMap.values())
+        sections = sections.map((section) => ({
+            ...section,
+            results: sortTasks(section.results, sort, {
```
gnapse, line 302: I think this can still return the wrong first page when a filter has more than `targetLimit` matches. We stop fetching in API storage order and sort that subset here, so a task after the cursor can never move into the first sorted page. Could we fetch the full section before sorting and apply the limit afterward, or reject cursor pagination when client sorting is active?

src/lib/task-sort.ts
```diff
@@ -0,0 +1,337 @@
...
+    return Intl.DateTimeFormat().resolvedOptions().timeZone || undefined
+}
+
+/**
+ * Tokens that make a filter query date-driven, which switches the default
+ * ordering from priority-first to date-first. Todoist parses the query for
+ * real; this is a keyword match over the English filter vocabulary, and the
+ * only thing riding on it is which of two default orders applies.
+ */
+const DATE_QUERY_PATTERN = new RegExp(
```
gnapse, line 319: I think this needs to account for the filter language. The backend parses a query using the account language when `lang` is not provided, so localized date terms can reach the SDK with `PRIORITY_FIRST`. Could we classify using the account language, or avoid duplicating this part of Filterist locally?

src/lib/task-sort.ts
```diff
@@ -0,0 +1,337 @@
...
+        const siblings = byParent.get(parentId) ?? []
+        siblings.push(project)
+        byParent.set(parentId, siblings)
+    }
+
+    for (const siblings of byParent.values()) {
+        siblings.sort(
+            (a, b) =>
+                compareText(folderKey(a), folderKey(b)) ||
+                compare(a.childOrder, b.childOrder) ||
```
gnapse: I don’t think these values represent the visible workspace order. `folderId` and workspace IDs are opaque, and workspace projects use `defaultOrder` rather than `childOrder`, so these ranks can differ from the sidebar. Could we build the context from the folder and workspace ordering data instead?

gnapse (changes requested):
I reviewed this with Codex against the SDK work in Doist/todoist-sdk-typescript#665 and #666. The SDK integration itself looks right, but I found four places where the context around it can still produce an order different from Todoist, so I left those inline. The PR is large, but I would not block on size alone since about half of the additions are tests and DoistBot already called out the review load.

_Review generated by Codex 5.6 Sol on my behalf_

src/lib/task-sort.ts
craigcarlyle: Fixed. `getAccountTimezone()` reads `tzInfo.timezone` off the account and hands it to `sortTasks`, cached per process and fetched in parallel with the projects, so it costs latency rather than a round trip. The machine clock is only a fallback now, and `--sort none` skips the lookup.

src/commands/filter/view.ts
craigcarlyle: Addressed. `--cursor` with an active sort is now a `CliError` pointing at `--all` or `--sort none`, and a truncated fetch says so in the header: `Sort: Todoist default (ordered over the 5 tasks fetched, use --all)`.

src/lib/task-sort.ts
craigcarlyle: Good call, I created Doist/todoist-sdk-typescript#669.

src/lib/task-sort.ts
craigcarlyle: Addressed. They now sort by `defaultOrder`, which already spans folders, so the folder key is gone. The workspaces themselves order by name via `fetchWorkspaces` (matching `td project list`).

gnapse:
I pushed a follow-up commit to close the pagination gap. Sorted filters now fetch the complete result set before applying `--limit`, so tasks on later API pages are included in the ordering.

_Drafted with Codex_

gnapse (approved):

PR #480, opened 2026-08-17
feat(sort): add task ordering that matches the Todoist apps

craigcarlyle:
## Summary
- Adds `src/lib/task-sort.ts`: Todoist's default ordering (priority, then date and time, then deadline, then project order, then task order within the project), the date-first variant that filters with date queries use, the named sorts, and sidebar project order
- Adds `src/lib/api/view-options.ts`: reads the sorting saved on a view from the `view_options` Sync resource
- Reads those options with a direct `/sync` POST instead of `api.sync()`, because the SDK schema types `object_id` as a required string while the API returns `null` for the Today and Upcoming rows, so a typed sync throws for anyone who has customised either view
- Reversing a sort flips only the primary criterion, and the tasks Todoist parks at the bottom of an ascending list (no date, no assignee) move to the top, per the [help article](https://www.todoist.com/help/articles/default-sorting-order-for-todoist-tasks-mqmgerY7)
- Keeps both modules independent of any command so `today`, `upcoming` and `label view` can pick them up later
- Nothing calls either module yet: `td filter view` wires them up in #479, the next PR in this stack

## Test plan
- [x] 32 new unit tests cover both default hierarchies, every named sort, direction reversal, project order and the date-query heuristic
- [x] Full suite passes on this branch alone (1827 tests)
- [x] Type-check, lint, format and SKILL.md sync clean
- [x] `view_options` payload confirmed against a live account, including the `null` `object_id` row that breaks the SDK

### DIFF
--- a/src/lib/api/view-options.ts
+++ b/src/lib/api/view-options.ts
@@ -1,0 +1,113 @@
+import type { GroupedBy, SortedBy, SortOrder, ViewMode, ViewType } from '@doist/todoist-sdk'
+import { getApiToken } from '../auth.js'
+import { getLogger } from '../logger.js'
+import { fetchTodoist } from '../usage-tracking.js'
+
+const SYNC_ENDPOINT = 'https://api.todoist.com/api/v1/sync'
+
+/**
+ * The presentation settings the Todoist apps store per view (project, label,
+ * filter, Today, Upcoming): list/board/calendar, grouping, and sorting.
+ *
+ * Named `SavedViewOptions` so it doesn't collide with `ViewOptions` in
+ * `src/lib/options.ts`, which is the CLI's own output-flag bag.
+ */
+export interface SavedViewOptions {
+    viewType: ViewType
+    /** `null` for the singleton views (Today, Upcoming) that have no object. */
+    objectId: string | null
+    sortedBy: SortedBy | null
+    sortOrder: SortOrder | null
+    groupedBy: GroupedBy | null
+    viewMode: ViewMode | null
+}
+
+interface RawViewOptions {
+    view_type?: unknown
+    object_id?: unknown
+    sorted_by?: unknown
+    sort_order?: unknown
+    grouped_by?: unknown
+    view_mode?: unknown
+    is_deleted?: unknown
+}
+
+function asString(value: unknown): string | null {
+    return typeof value === 'string' && value.length > 0 ? value : null
+}
+
+function parseViewOptions(raw: RawViewOptions): SavedViewOptions | null {
+    const viewType = asString(raw.view_type)
+    if (!viewType || raw.is_deleted === true) return null
+
+    return {
+        viewType: viewType as ViewType,
+        objectId: asString(raw.object_id),
+        sortedBy: asString(raw.sorted_by) as SortedBy | null,
+        sortOrder: asString(raw.sort_order) as SortOrder | null,
+        groupedBy: asString(raw.grouped_by) as GroupedBy | null,
+        viewMode: asString(raw.view_mode) as ViewMode | null,
+    }
+}
+
+/**
+ * Read every saved view option via the Sync API.
+ *
+ * This goes to `/sync` directly instead of through `api.sync()` because the
+ * SDK's response schema types `object_id` as a required string, while the API
+ * returns `null` for Today and Upcoming — so a typed sync that asks for
+ * `view_options` throws for anyone who has customised either view. Parsing the
+ * handful of fields we need here also means an unknown enum value can never
+ * take down a read-only command.
+ */
+export async function fetchViewOptions(): Promise<SavedViewOptions[]> {
+    const token = await getApiToken()
+    const response = await fetchTodoist(SYNC_ENDPOINT, {
+        method: 'POST',
+        headers: {
+            Authorization: `Bearer ${token}`,
+            'Content-Type': 'application/x-www-form-urlencoded',
+        },
+        body: new URLSearchParams({
+            sync_token: '*',
+            resource_types: JSON.stringify(['view_options']),
+        }),
+    })
+
+    if (!response.ok) {
+        throw new Error(`HTTP ${response.status} ${response.statusText}`)
+    }
+
+    const data = (await response.json()) as { view_options?: unknown }
+    if (!Array.isArray(data.view_options)) return []
+
+    return data.view_options
+        .map((entry) => parseViewOptions((entry ?? {}) as RawViewOptions))
+        .filter((entry): entry is SavedViewOptions => entry !== null)
+}
+
+/**
+ * Saved view options are a display nicety: when they can't be read, callers
+ * should still render the list rather than fail. Returns `[]` on any error and
+ * leaves a breadcrumb behind `-v`.
+ */
+export async function fetchViewOptionsSafely(): Promise<SavedViewOptions[]> {
+    try {
+        return await fetchViewOptions()
+    } catch (error) {
+        getLogger().detail('failed to load saved view options', {
+            error: error instanceof Error ? error.message : String(error),
+        })
+        return []
+    }
+}
+
+/** Find the saved options for one view, e.g. the FILTER view of a filter id. */
+export function findViewOptions(
+    viewOptions: SavedViewOptions[],
+    { viewTypes, objectId }: { viewTypes: ViewType[]; objectId: string },
+): SavedViewOptions | undefined {
+    return viewOptions.find(
+        (entry) => entry.objectId === objectId && viewTypes.includes(entry.viewType),
+    )
+}
--- a/src/lib/api/view-options.ts
+++ b/src/lib/api/view-options.ts
@@ -1,3 +1,10 @@
+import {
+    GROUPED_BY_OPTIONS,
+    SORT_ORDERS,
+    SORTED_BY_OPTIONS,
+    VIEW_MODES,
+    VIEW_TYPES,
+} from '@doist/todoist-sdk'
 import type { GroupedBy, SortedBy, SortOrder, ViewMode, ViewType } from '@doist/todoist-sdk'
 import { getApiToken } from '../auth.js'
 import { getLogger } from '../logger.js'
@@ -36,29 +43,44 @@ function asString(value: unknown): string | null {
     return typeof value === 'string' && value.length > 0 ? value : null
 }
 
+/**
+ * Read an enum value only if it is one the SDK still knows about.
+ *
+ * The Sync API can grow a new member before the SDK types catch up, and a
+ * value we don't recognise is better dropped than asserted into a type that
+ * then lies to everything downstream. An unknown `sorted_by` reads as "no
+ * saved sort", which lands the caller on Todoist's default ordering.
+ */
+function asMember<T extends string>(value: unknown, allowed: readonly T[]): T | null {
+    const candidate = asString(value)
+    return candidate && (allowed as readonly string[]).includes(candidate) ? (candidate as T) : null
+}
+
 function parseViewOptions(raw: RawViewOptions): SavedViewOptions | null {
-    const viewType = asString(raw.view_type)
+    // A view type we can't name is one no caller can ask for, so the row goes.
+    const viewType = asMember<ViewType>(raw.view_type, VIEW_TYPES)
     if (!viewType || raw.is_deleted === true) return null
 
     return {
-        viewType: viewType as ViewType,
+        viewType,
         objectId: asString(raw.object_id),
-        sortedBy: asString(raw.sorted_by) as SortedBy | null,
-        sortOrder: asString(raw.sort_order) as SortOrder | null,
-        groupedBy: asString(raw.grouped_by) as GroupedBy | null,
-        viewMode: asString(raw.view_mode) as ViewMode | null,
+        sortedBy: asMember<SortedBy>(raw.sorted_by, SORTED_BY_OPTIONS),
+        sortOrder: asMember<SortOrder>(raw.sort_order, SORT_ORDERS),
+        groupedBy: asMember<GroupedBy>(raw.grouped_by, GROUPED_BY_OPTIONS),
+        viewMode: asMember<ViewMode>(raw.view_mode, VIEW_MODES),
     }
 }
 
 /**
  * Read every saved view option via the Sync API.
  *
- * This goes to `/sync` directly instead of through `api.sync()` because the
- * SDK's response schema types `object_id` as a required string, while the API
- * returns `null` for Today and Upcoming — so a typed sync that asks for
- * `view_options` throws for anyone who has customised either view. Parsing the
- * handful of fields we need here also means an unknown enum value can never
- * take down a read-only command.
+ * This goes to `/sync` directly instead of through `api.sync()` for one
+ * reason: `ViewOptionsSchema` types `object_id` as a required string, while
+ * the API returns `null` for the singleton views (Today, Upcoming), so a typed
+ * sync asking for `view_options` throws for anyone who has customised either
+ * one. Making that single field nullable in the SDK is the whole fix. Once it
+ * ships, this reader can drop back to `api.sync()` or move into the SDK
+ * outright, and the enum handling below goes with it.
  */
 export async function fetchViewOptions(): Promise<SavedViewOptions[]> {
     const token = await getApiToken()
--- a/src/lib/api/view-options.ts
+++ b/src/lib/api/view-options.ts
@@ -89,7 +89,7 @@ export async function fetchViewOptions(): Promise<SavedViewOptions[]> {
 /**
  * Saved view options are a display nicety: when they can't be read, callers
  * should still render the list rather than fail. Returns `[]` on any error and
- * leaves a breadcrumb behind `-v`.
+ * leaves a breadcrumb behind `-vv`.
  */
 export async function fetchViewOptionsSafely(): Promise<SavedViewOptions[]> {
     try {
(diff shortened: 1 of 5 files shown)


doistbot (commented):
This PR adds two standalone modules — `task-sort.ts` for ordering tasks the way the Todoist apps do and `view-options.ts` for reading saved view options via a direct `/sync` call — with 32 new tests, ahead of the wiring in #479.

Few things worth tightening:

- The floating-due-time conversion appends `Z` unconditionally, ignoring `task.due.timezone`; timed tasks in different IANA zones will sort in the wrong order — pass the due timezone through and resolve unzoned times in that zone, with a cross-timezone test.
- `sortNeedsProjects` returns `false` for `priority`, `date`, `deadline`, `date-added`, and `name`, but all of these can tie-break on project order via `compareDefault` — either return `true` for everything except `none` or stop consulting `projectValue` in the shared fallback.
- The date-query heuristic in the filter string parser matches `search: today` / `search: due` as date-driven; strip or skip `search:` operands before applying the date pattern and add coverage.
- `view-options.ts` has no tests — `findViewOptions` is pure and exported, and the null-`object_id` / `is_deleted` filtering logic is exactly what the module exists for, so a small stubbed suite would add real regression protection.
- The assignee sort is only tested ascending; add a descending case (e.g. `['unassigned', 'zoe', 'ana']`) to guard the null-handling × reversal interaction.

<details>
<summary>Optional follow-up notes (8)</summary>

- **<a href="#">P3</a> src/lib/task-sort.ts:118**: The JSDoc says only `project`, `workspace`, and `default` sorts need the project list, but the function also returns `true` for `assignee`. That behavior is correct — assignee sorting needs the project map to resolve collaborator names (`CollaboratorCache.preload` -> `getUserName` looks up the project to find its workspace/shared status) — so the comment understates the contract. A caller wiring `sortNeedsProjects` per the docs (e.g. #479) could skip the project fetch for assignee sorts and silently leave every assignee unresolved. Update the comment to include `assignee`.
- **<a href="#">P3</a> src/lib/api/view-options.ts:108**: `SavedViewOptions.objectId` is `string | null` for singleton views like Today and Upcoming, but `findViewOptions` types `objectId` as a required `string`. Updating the parameter to `objectId?: string | null` (e.g. `{ viewTypes, objectId = null }: { viewTypes: ViewType[]; objectId?: string | null }`) will allow callers for Today and Upcoming views to query saved options without type errors.
- **<a href="#">P3</a> src/lib/task-sort.ts:262**: `timestampOf` hand-rolls ISO-8601 parsing (a timezone-suffix regex plus `Date.parse` normalization) that `date-fns/parseISO` already provides. The repo depends on date-fns directly and uses `parseISO` in `src/lib/dates.ts`, so this is a duplicate capability against the existing convention. `parseISO` handles date-only, offset, and `Z` strings, and for relative ordering it preserves the all-day-before-timed behavior this helper exists for; the invalid-input check becomes `isNaN(parseISO(value).getTime())`. Reusing it drops a custom parser and keeps date handling in one idiom.
- **<a href="#">P3</a> src/lib/api/view-options.ts:92**: The comment says the failure breadcrumb appears behind `-v`, but `getLogger().detail()` logs at DETAIL level 2, which needs `-vv`; `-v` (INFO) won't show it. Either switch to `getLogger().info()` to match the comment, or correct the comment to `-vv`.
- **<a href="#">P3</a> src/lib/task-sort.ts:364**: `sort.field === 'none'` returns the original `tasks` reference, but every other branch returns `[...tasks].sort(...)`. A caller that mutates the result (e.g. `.push()`, `.reverse()`) would accidentally mutate the input when the sort field is `'none'`. The existing "does not mutate the input list" test only covers `'default'`, so this gap is unguarded. Trivial fix: `return [...tasks]`.
- **<a href="#">P3</a> src/lib/task-sort.ts:368**: `sortTasks`' comparator re-parses date strings on every pairwise comparison. `compareDefault`/`comparePrimary` call `dueValue`/`deadlineValue`/`scheduleValue`, and each call runs a regex plus `Date.parse` in `timestampOf` — so the same task's dates are parsed many times over the ~n log n comparisons (a 300-task page is ~2,500 comparisons). Precompute a sort key once per task (decorate-sort-undecorate) and have the comparator read the cached values.
- **<a href="#">P3</a> src/lib/task-sort.test.ts:384**: The `formatTaskSort` test pins two exact display strings that are just a lookup over `FIELD_LABELS` in the same file — it duplicates the constants, covers only 2 of 10 fields, and breaks on wording changes rather than a real behavior regression. The `defaultDirectionFor` and `sortNeeds*` assertions in the same block are similar low-signal lookup tests (`defaultDirectionFor` is already exercised through `taskSortFromViewOptions`, and the interesting `sortNeedsProjects('assignee')` case isn't even checked). Consider keeping only the parse error-path tests, which have real signal.
- **<a href="#">P3</a> src/lib/task-sort.test.ts:251**: The 'workspace' sort field is a user-facing option (listed in `TASK_SORT_FIELDS`, mapped from `WORKSPACE` in `FIELD_BY_SORTED_BY`) but has zero test coverage — neither `workspaceValue` nor the `'workspace'` case in `comparePrimary` is exercised by any test. `buildProjectOrder` creates `workspaceIndex` but no `sortTasks` call consumes it. If the workspace sort path silently produces `NO_VALUE` for every task (e.g., because `sortNeedsProjects('workspace')` regressed to false), no test would fail.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F480&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Af04d3750-9a66-11f1-990b-a3bd1a009a85%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Af04d3750-9a66-11f1-990b-a3bd1a009a85&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/task-sort.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+/**
+ * Comparable instant for a due or deadline value. Date-only values become
+ * midnight UTC and floating datetimes are read as UTC, so an all-day task
+ * sorts ahead of a timed task on the same day.
+ */
+function timestampOf(value: string): number {
+    const hasZone = /(?:Z|[+-]\d{2}:?\d{2})$/.test(value)
+    const normalized = value.includes('T') ? (hasZone ? value : `${value}Z`) : `${value}T00:00:00Z`
```
doistbot: <a href="#">P2</a>  This appends `Z` to every floating due datetime, ignoring `task.due.timezone`. Timed Todoist tasks can carry an IANA timezone, so tasks such as `09:00` in `America/Los_Angeles` and `10:00` in `America/New_York` are ordered as 09:00 then 10:00 here even though the New York task occurs two hours earlier. Pass the due timezone into this conversion and resolve unzoned datetimes in that zone (while retaining explicit-offset handling), with a cross-timezone sorting test.

src/lib/task-sort.ts
```diff
@@ -0,0 +1,409 @@
...
+}
+
+export function formatTaskSort(sort: TaskSort): string {
+    return FIELD_LABELS[sort.field][sort.direction]
+}
+
+/** Sorting by project, workspace, or the default hierarchy needs the project list. */
+export function sortNeedsProjects(field: TaskSortField): boolean {
+    return (
+        field === 'default' || field === 'project' || field === 'workspace' || field === 'assignee'
```
doistbot: <a href="#">P2</a>  `sortNeedsProjects` misses several fields that still need the project list for tie-breaking. `sortTasks` falls back to `compareDefault` whenever the primary key ties, and its non-date-driven branch compares `projectValue` (L317). So `priority`, `date`, `deadline`, `date-added`, and `name` ties also depend on project order — but this helper returns `false` for all five. (`assignee` is already included for exactly this reason, even though its primary key is the assignee name.) Callers that trust this helper will skip the project fetch and those ties will silently fall back to `childOrder` instead of matching the Todoist apps. Return `true` for every field except `none`, or stop consulting `projectValue` in the shared fallback.

src/lib/task-sort.ts
```diff
@@ -0,0 +1,409 @@
...
+ */
+function stripNamedRefs(query: string): string {
+    return query
+        .replace(/"[^"]*"/g, ' ')
+        .replace(/'[^']*'/g, ' ')
+        .replace(/[#@/]{1,2}[^\s&|()!,]+/g, ' ')
+}
+
+export function queryUsesDates(query: string): boolean {
+    return DATE_QUERY_PATTERN.test(stripNamedRefs(query))
```
doistbot, line 418: <a href="#">P2</a>  Exclude `search:` operands before applying the date pattern. `search: today` or `search: due` is a text-search filter, but this marks it date-driven and sorts it date-first rather than priority-first. Strip or parse unquoted `search:` terms too, and add coverage for this case.

src/lib/api/view-options.ts
```diff
@@ -0,0 +1,113 @@
...
+    } catch (error) {
+        getLogger().detail('failed to load saved view options', {
+            error: error instanceof Error ? error.message : String(error),
+        })
+        return []
+    }
+}
+
+/** Find the saved options for one view, e.g. the FILTER view of a filter id. */
+export function findViewOptions(
```
doistbot, line 132: <a href="#">P2</a>  `view-options.ts` has zero test coverage. `findViewOptions` is pure and exported, so its match logic (objectId equality + viewType membership, including the non-matching/undefined cases) is cheap to test. The `parseViewOptions` behaviors this module was written for — keeping rows with `object_id: null` (Today/Upcoming) and dropping `is_deleted: true` rows — are the exact SDK/API quirk called out in the module docstring and have no regression protection. Add a small suite here: stub `getApiToken` and `fetchTodoist` (or `globalThis.fetch`, since `fetchTodoist` accepts a `fetchImpl`) for `fetchViewOptions`, and cover `findViewOptions` directly.

src/lib/task-sort.test.ts
```diff
@@ -0,0 +1,387 @@
...
+                sortTasks(
+                    tasks,
+                    { field: 'assignee', direction: 'asc' },
+                    {
+                        assigneeName: (task) =>
+                            task.responsibleUid ? names[task.responsibleUid] : null,
+                    },
+                ),
+            ),
+        ).toEqual(['ana', 'zoe', 'unassigned'])
```
doistbot, line 284: <a href="#">P2</a>  The assignee sort is only tested ascending. The code comment on `sortTasks` (task-sort.ts:359–361) and the PR description both call out that values parked at the end of an ascending list — including unassigned tasks — move to the top when reversed. The date field tests this reversal at line 140, but nothing verifies the interaction between `compareAssignee`'s null handling and the `* reverse` multiplier. If `compareAssignee`'s null special-casing changes, no test catches the regression. Add a descending assertion (e.g., `['unassigned', 'zoe', 'ana']`) alongside the existing ascending one.

src/lib/task-sort.ts
craigcarlyle: Fixed in 835a2ae, though not by the route suggested: `due.timezone` is null exactly when the datetime is floating, so it can't tell us which zone to resolve an unzoned time in.

The underlying problem was that the three shapes were being read in different frames. Date-only values became UTC midnight, floating datetimes were forced to UTC by the appended `Z`, and only offset datetimes carried a real instant. `date-fns/parseISO`, already the repo's date idiom in `src/lib/dates.ts`, reads date-only and floating values in the local zone and resolves offsets properly, which puts all three on one axis and drops the hand-rolled parser.

Covered by a cross-timezone test: noon in Tokyo now sorts between an all-day task and a floating 09:00.

src/lib/task-sort.ts
craigcarlyle: Agreed, and it was observable: `--json --sort name` ordered equal names differently from the pretty output, which fetches projects either way.

Took the first option in 835a2ae, `field !== 'none'`. The second one would have removed project order from the default hierarchy, where the help centre puts it as the fourth criterion. The filter tests in #479 now assert the fetch happens under a named sort and is skipped only for `--sort none`.

src/lib/task-sort.ts
craigcarlyle: Fixed in 835a2ae. `stripNamedRefs` now drops a `search:` operand up to the next boolean operator or list comma, alongside the quoted strings and `#`/`@` tokens it already stripped.

Tests cover `search: due diligence` and `search: today notes`, plus `search: invoice & today` so a real date term after the operand still counts.

src/lib/api/view-options.ts
craigcarlyle: Added in 835a2ae: `src/lib/api/view-options.test.ts` stubs `getApiToken` and `fetchTodoist`, then covers the `object_id: null` row, `is_deleted` filtering, the snake-to-camel mapping and the HTTP failure path.

`findViewOptions` is tested directly, including an id saved under a different view type and the singleton views that have no id at all. That last case needed a signature change: `objectId` now defaults to null, so a caller can look up Today or Upcoming without a type error.

src/lib/task-sort.test.ts
craigcarlyle: Added in 835a2ae, with the case you suggested: `['unassigned', 'zoe', 'ana']`.

Same commit also covers the `workspace` sort, which had no test at all, in both directions.

src/lib/api/view-options.ts
```diff
@@ -0,0 +1,117 @@
...
+    const viewType = asString(raw.view_type)
+    if (!viewType || raw.is_deleted === true) return null
+
+    return {
+        viewType: viewType as ViewType,
+        objectId: asString(raw.object_id),
+        sortedBy: asString(raw.sorted_by) as SortedBy | null,
+        sortOrder: asString(raw.sort_order) as SortOrder | null,
+        groupedBy: asString(raw.grouped_by) as GroupedBy | null,
+        viewMode: asString(raw.view_mode) as ViewMode | null,
```
gnapse: is it safe to forecefully typecast here? what if the raw value is not really valid for the type?

src/lib/api/view-options.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Read every saved view option via the Sync API.
+ *
+ * This goes to `/sync` directly instead of through `api.sync()` because the
+ * SDK's response schema types `object_id` as a required string, while the API
+ * returns `null` for Today and Upcoming — so a typed sync that asks for
+ * `view_options` throws for anyone who has customised either view. Parsing the
+ * handful of fields we need here also means an unknown enum value can never
+ * take down a read-only command.
```
gnapse: Could we improve the SDK? (not saying you should, or that it'd be a blocker, just asking if we could improve it so we do not need to bypass it here)

gnapse (commented):
The more I look at all this, the more I'm thinking that parts of this should go into the SDK. I may take the task to try that out.

src/lib/api/view-options.ts
scottlovegrove: Yes please, let's fix this in the SDK. We shouldn't really be adding any code that aims to work around any SDK issue.

src/lib/api/view-options.ts
craigcarlyle: Good catch, it wasn't safe. Fixed in 9a562b3.

Each field is now checked against the constant the SDK already exports (`VIEW_TYPES`, `SORTED_BY_OPTIONS`, `SORT_ORDERS`, `GROUPED_BY_OPTIONS`, `VIEW_MODES`) rather than asserted, which has the side benefit of the CLI no longer restating enums the SDK owns. An unrecognised `sorted_by` reads as "no saved sort", so the caller lands on Todoist's default ordering, and a row whose `view_type` we can't name is dropped because no caller can ask for it. Two tests cover both paths.

src/lib/api/view-options.ts
craigcarlyle: Yes, and it's a one-field change: `ViewOptionsSchema` types `object_id` as a required string, but the API returns `null` for Today and Upcoming, so `api.sync({ resourceTypes: ['view_options'] })` throws a `ZodError` for anyone who has customised either view. I hit it on my own account. Making that field nullable is the entire fix.

I've put that in the module docstring in 9a562b3 so whoever picks it up doesn't have to rediscover it, and noted that this reader can go back to `api.sync()` once it ships. Happy to send the SDK PR if you'd rather not carry it.

craigcarlyle:
@gnapse agreed on `view-options.ts`. It's a pure Sync API reader with no CLI in it, and it only exists because of the `object_id` schema gap, so the SDK is where it belongs. Take it if you want it and I'll delete this copy.

`task-sort.ts` I'd say is a closer call. The ordering it implements is app behaviour rather than API behaviour: the endpoints return storage order and each client applies the [documented hierarchy](https://www.todoist.com/help/articles/default-sorting-order-for-todoist-tasks-mqmgerY7) itself. That is an argument for the SDK (every client needs it, nobody should reimplement it) as much as against it (it encodes product decisions the API doesn't state). One thing to know if you do move it: it currently imports `CliError` for the flag parsing, and that would need to come out first.

Nothing here blocks on the outcome. Both modules are self-contained and the wiring in #479 touches them through a small surface, so lifting either one later is a delete plus an import change rather than a rewrite.

PR #482, opened 2026-08-18
feat(comment): notify collaborators when adding a comment

scottlovegrove:
Brings `td comment add` to parity with the MCP, which shipped this in Doist/todoist-mcp#580 for Doist/todoist-mcp#509.

**Stacked on #477** — based on that branch for SDK 14.0.1, which carries the `uidsToNotify` serialisation fix. Retarget to `main` once #477 merges.

## The problem

Comments posted with `td` notify nobody. `comment add` never sent `uidsToNotify`, so a teammate named in a comment found out only if they happened to open the task.

That also breaks the *next* comment, which is the part people notice and can't reproduce. Todoist's clients pick the recipients themselves and send them with each comment; the API notifies exactly who it is handed and derives nobody on its own. On a first comment the clients notify the assignee, the assigner and the creator; on a reply they notify the previous comment's participants, so threads keep flowing without everyone being re-tagged. A comment posted with an empty list therefore silences the comment that follows it — including one a human later writes in the app.

## What this does

`comment add` gains `--notify`, accepting names, emails, `id:xxx` or `"me"`, comma-separated in the same style as `--labels`. `--no-notify` posts in silence, matching the existing `--no-labels` negation.

Omitting `--notify` mirrors the clients: assignee, assigner and creator on a task's first comment, or the previous comment's participants on a reply, always excluding the author. The rules live in `src/lib/comment-recipients.ts`, ported from the MCP's equivalent.

Two things worth calling out in review:

- **Naming yourself is honoured**, not filtered out. Self-exclusion is only right when the recipients were *inferred* rather than asked for. I had this filtering unconditionally at first and caught it in live testing — `--notify me` silently did nothing.
- **The thread walk keeps only the newest comment** rather than using `paginate()`, which accumulates every result. Comments come back oldest-first with no reverse option, so reaching the last page is unavoidable; holding the whole history is not. Review flagged exactly this on the MCP version.

`resolveNotifyIds` is pure over an already-fetched collaborator list, so the `--notify` path fetches collaborators once and reuses them to render the names back — my own tests caught a double fetch here.

`@mentions` in the comment text are deliberately **not** parsed; the user names people with `--notify`. `SKILL_CONTENT` documents this so agents don't assume the text alone notifies.

Notification on *edit* is out of scope — `UpdateCommentArgs` is `{ content }` only.

## Surfacing

Who was notified is now visible in three places: the confirmation line after adding, a `Notified:` line in `comment view` (resolved only when there are recipients, so the common case costs no extra request), and `postedUid` / `uidsToNotify` in plain `--json` rather than only under `--full`.

## Verification

`npm run type-check`, `npm run check`, `npm run build`, `npm run check:skill-sync` and all 1815 tests pass (suite run three times to rule out flakiness).

Live against the real API on a throwaway task in a shared project, since deleted:

| case | result |
| --- | --- |
| `--dry-run` with `--notify` | previews the raw string unresolved, no API call |
| defaults, first comment on a solo task | no recipients |
| `--notify "Ada Lovelace"` | `Notified: Ada L.` |
| `--notify "me,ada@example.com"` | deduped to one |
| `--notify Ghost,Phantom` | one `ASSIGNEE_NOT_FOUND` naming both |
| reply, defaults | previous comment's participants, author excluded |
| `--no-notify` | silent, no recipient field sent |
| `comment view` / `--json` | recipients shown |

The paths involving a **second person** — a reply inheriting someone else's participants, and a first comment on a task assigned to someone else — are covered by unit tests rather than live calls, since verifying them for real means sending test notifications to an actual colleague.

### DIFF
--- a/src/commands/comment/add.ts
+++ b/src/commands/comment/add.ts
@@ -93,12 +93,6 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
     } else if (options.notify) {
         const project = await api.getProject(targetProjectId)
         notifyCollaborators = await fetchCollaboratorsForProject(api, project)
-        if (notifyCollaborators.length === 0) {
-            throw new CliError(
-                'NOT_SHARED',
-                `Cannot notify anyone on "${project.name}" — it is not shared with anybody.`,
-            )
-        }
         uidsToNotify = resolveNotifyIds({
             refs: options.notify.split(','),
             collaborators: notifyCollaborators,
--- a/src/commands/comment/add.ts
+++ b/src/commands/comment/add.ts
@@ -1,5 +1,12 @@
 import chalk from 'chalk'
-import { getApi } from '../../lib/api/core.js'
+import { getApi, getCurrentUserId } from '../../lib/api/core.js'
+import {
+    type CollaboratorInfo,
+    fetchCollaboratorsForProject,
+    formatUserShortName,
+    resolveNotifyIds,
+} from '../../lib/collaborators.js'
+import { getDefaultCommentRecipients } from '../../lib/comment-recipients.js'
 import { CliError } from '../../lib/errors.js'
 import { isQuiet } from '../../lib/global-args.js'
 import { openLocalFileAsBlob } from '../../lib/local-file.js'
@@ -13,6 +20,9 @@ interface AddOptions {
     file?: string
     fileName?: string
     project?: boolean
+    // Commander sets this to false for --no-notify, and to the raw
+    // comma-separated string for --notify.
+    notify?: string | false
     json?: boolean
     dryRun?: boolean
 }
@@ -51,6 +61,9 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
             'Target type': options.project ? 'project' : 'task',
             Content: content.length > 80 ? `${content.slice(0, 80)}...` : content,
             File: options.file,
+            // Printed unresolved: the dry-run deliberately runs before getApi(),
+            // so no lookup has happened at this point.
+            Notify: describeNotifyOption(options.notify),
         })
         return
     }
@@ -59,14 +72,41 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
 
     let targetArgs: { taskId: string } | { projectId: string }
     let targetName: string
+    let targetProjectId: string
     if (options.project) {
         const project = await resolveProjectRef(api, ref)
         targetArgs = { projectId: project.id }
         targetName = project.name
+        targetProjectId = project.id
     } else {
         const task = await resolveTaskRef(api, ref)
         targetArgs = { taskId: task.id }
         targetName = task.content
+        targetProjectId = task.projectId
+    }
+
+    // Held so the confirmation line can name people without a second fetch.
+    let notifyCollaborators: CollaboratorInfo[] | undefined
+    let uidsToNotify: string[]
+    if (options.notify === false) {
+        uidsToNotify = []
+    } else if (options.notify) {
+        const project = await api.getProject(targetProjectId)
+        notifyCollaborators = await fetchCollaboratorsForProject(api, project)
+        if (notifyCollaborators.length === 0) {
+            throw new CliError(
+                'NOT_SHARED',
+                `Cannot notify anyone on "${project.name}" — it is not shared with anybody.`,
+            )
+        }
+        uidsToNotify = resolveNotifyIds({
+            refs: options.notify.split(','),
+            collaborators: notifyCollaborators,
+            currentUserId: await getCurrentUserId(),
+            projectName: project.name,
+        })
+    } else {
+        uidsToNotify = await getDefaultCommentRecipients(api, targetArgs, await getCurrentUserId())
     }
 
     let attachment:
@@ -98,6 +138,7 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
         ...targetArgs,
         content,
         ...(attachment && { attachment }),
+        ...(uidsToNotify.length > 0 && { uidsToNotify }),
     })
 
     if (options.json) {
@@ -114,5 +155,41 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
     if (attachment) {
         console.log(chalk.dim(`Attached: ${attachment.fileName}`))
     }
+    if (uidsToNotify.length > 0) {
+        const names = await describeRecipients(
+            api,
+            uidsToNotify,
+            targetProjectId,
+            notifyCollaborators,
+        )
+        console.log(chalk.dim(`Notified: ${names}`))
+    }
     console.log(chalk.dim(`ID: ${comment.id}`))
+}
+
+function describeNotifyOption(notify: string | false | undefined): string | undefined {
+    if (notify === false) return '(nobody)'
+    if (notify) return notify
+    return undefined
+}
+
+/**
+ * Render recipients as short names, falling back to the raw ID for anyone the
+ * collaborator list does not cover — the same fallback `formatAssignee` makes.
+ */
+async function describeRecipients(
+    api: Awaited<ReturnType<typeof getApi>>,
+    userIds: string[],
+    projectId: string,
+    known: CollaboratorInfo[] | undefined,
+): Promise<string> {
+    try {
+        const collaborators =
+            known ?? (await fetchCollaboratorsForProject(api, await api.getProject(projectId)))
+        const names = new Map(collaborators.map((c) => [c.id, formatUserShortName(c.name)]))
+        return userIds.map((id) => names.get(id) ?? id).join(', ')
+    } catch {
+        // Naming people is a nicety; never fail a posted comment over it.
+        return userIds.join(', ')
+    }
 }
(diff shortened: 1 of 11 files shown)


src/commands/comment/add.ts
```diff
@@ -59,14 +72,41 @@ export async function addComment(ref: string, options: AddOptions): Promise<void
...
+
+    // Held so the confirmation line can name people without a second fetch.
+    let notifyCollaborators: CollaboratorInfo[] | undefined
+    let uidsToNotify: string[]
+    if (options.notify === false) {
+        uidsToNotify = []
+    } else if (options.notify) {
+        const project = await api.getProject(targetProjectId)
+        notifyCollaborators = await fetchCollaboratorsForProject(api, project)
+        if (notifyCollaborators.length === 0) {
```
doistbot: <a href="#">P2</a>  The `notifyCollaborators.length === 0` guard runs before `resolveNotifyIds`, so `--notify me` (or `--notify id:<self>`) on a task in a non-shared personal project throws `NOT_SHARED` instead of notifying you. `fetchCollaboratorsForProject` returns `[]` for a non-shared personal project, but `me` and `id:` refs never need the collaborator list — `resolveAssigneeId` resolves both *before* its own shared-project check for exactly this reason. Move the self/id short-circuit ahead of the guard (or let `resolveNotifyIds` resolve those refs first and only throw `NOT_SHARED` for name/email refs when the list is empty) so explicit self-notification works on personal projects.

src/lib/collaborators.ts
```diff
@@ -246,3 +256,81 @@ export async function resolveAssigneeId(
...
+/**
+ * Resolve a list of user references — names, emails, `id:xxx` or `me` — to the
+ * user IDs to notify about a comment.
+ *
+ * Pure over an already-fetched collaborator list, so the caller fetches once
+ * and can reuse the same list to render the names back. References are
+ * deduplicated, and every one that cannot be resolved is reported together so
+ * the caller fixes them all in one go rather than one per run.
+ */
+export function resolveNotifyIds({
```
doistbot, line 287: <a href="#">P2</a>  `resolveNotifyIds` re-implements the per-ref matching already in `resolveAssigneeId` in this same file: `me`, `id:xxx`, exact name, exact email, substring name, and the identical `AMBIGUOUS_ASSIGNEE` error (including the `slice(0, 5)` candidate list). Only the batching of unresolved refs and dedup differ. Extract the single-ref matching into a shared pure helper (e.g. `resolveCollaboratorRef(ref, collaborators)`), then have `resolveAssigneeId` fetch once and call it while `resolveNotifyIds` loops it and collects misses, so the two match rules cannot drift apart.

src/commands/comment/view.ts
```diff
@@ -46,9 +48,12 @@ export async function viewComment(commentId: string, options: ViewOptions): Prom
 
     console.log(chalk.bold('Comment'))
     console.log('')
-    console.log(`ID:      ${comment.id}`)
-    console.log(`Posted:  ${comment.postedAt}`)
-    if (url) console.log(`URL:     ${url}`)
+    console.log(`ID:       ${comment.id}`)
+    console.log(`Posted:   ${comment.postedAt}`)
+    if (comment.uidsToNotify?.length) {
+        console.log(`Notified: ${await describeNotified(api, comment)}`)
```
doistbot, line 54: <a href="#">P2</a>  The `Notified:` line added here (via `describeNotified`) has no test coverage. None of the `comment view` tests set `uidsToNotify`, so the new branch never runs: collaborator name resolution, the task-vs-project fallback (`comment.projectId ?? getTask(...)`), and the raw-ID fallback on lookup failure can all break silently. Add a `comment view` test with `uidsToNotify` set that asserts the rendered names, plus one where the collaborator lookup fails to assert the raw-ID fallback.

doistbot (commented):
This PR adds `--notify`/`--no-notify` to `td comment add` so comments can notify collaborators, bringing it to parity with the MCP's notification feature — including sensible defaults that mirror the Todoist clients' recipient inference logic.

<details>

- **<a href="#">P3</a> src/commands/comment/add.ts:180**: `describeRecipients` here and `describeNotified` in `src/commands/comment/view.ts` (line 81) duplicate the same id → short-name map plus raw-id fallback (`new Map(collaborators.map(c => [c.id, formatUserShortName(c.name)]))` then `userIds.map(id => names.get(id) ?? id).join(', ')`). Pull that mapping into one shared helper (e.g. `formatUserIds(collaborators, userIds)` in `src/lib/collaborators.ts`) and have both callers use it; each file keeps only its own fetch path.
- **<a href="#">P3</a> src/lib/collaborators.ts:168**: This recreates the workspace and shared-project pagination/mapping already implemented by `CollaboratorCache.fetchWorkspaceUsers` and `fetchProjectCollaborators`. Extract the endpoint-specific fetches into shared helpers (or expose a cache load/read API) and have both paths use them, so cursor handling and collaborator normalization have one implementation.
- **<a href="#">P3</a> src/commands/comment/add.ts:94**: When `--project` is set, `resolveProjectRef` already returned the full project (it's used for `targetName`/`targetProjectId` above), so this `api.getProject(targetProjectId)` is a redundant request. Hoist the resolved project out of the branch and reuse it here instead of refetching.
- **<a href="#">P3</a> CODEBASE.md:181**: Remove the dedicated `comment-recipients.ts` entry. This helper currently has one production caller (`comment/add.ts`), so listing it makes the structural map drift toward a file index rather than its intended catalog of broadly reusable helpers.
- **<a href="#">P3</a> src/lib/output.ts:323**: The plain-`--json` contract change (adding `postedUid`/`uidsToNotify` to `COMMENT_ESSENTIAL_FIELDS`) isn't pinned by any test. The existing `comment view --json` and `comment add --json` tests only assert `id`/`content`/`postedAt`, so dropping these two fields from the essential list would pass CI. Assert that both keys appear in non-`--full` JSON output.
- **<a href="#">P3</a> src/lib/comment-recipients.test.ts:43**: This fixture puts the newest comment first, so the test doesn't actually exercise the 'not the first page' behavior its name claims — it would still pass if `getLatestComment` returned `results[0]`. The within-page newest-not-first case is already covered by the integration test `defaults a reply...` in comment.test.ts (older comment listed first). Either reorder the fixture so the stale comment comes first, or drop this test as redundant.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F482&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac572a2e0-9af7-11f1-826e-d9a8d417e643%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac572a2e0-9af7-11f1-826e-d9a8d417e643&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/comment/add.ts
scottlovegrove: Fixed in 28254b0 — good catch, and the second instance of this same class of bug in this PR. `me` and `id:` refs now resolve inside `resolveNotifyIds` before anything consults the collaborator list, so they work on a personal project. `NOT_SHARED` is raised only for the name/email refs that genuinely need the list, and it names which ones they were rather than blaming the whole command. Verified live against a task in an unshared personal project: `--notify me` posts and notifies, `--notify Ana` gives `Cannot notify "Ana" — "<project>" is not shared with anybody.`

src/lib/collaborators.ts
scottlovegrove: Fixed in 28254b0. Extracted `matchCollaboratorRef(ref, collaborators)` — exact name, exact email, unique partial, and the ambiguity error with its candidate list — and both `resolveAssigneeId` and `resolveNotifyIds` now call it. `me` and `id:` stay with the callers, since neither needs a collaborator list and that short-circuit is exactly what the other thread was about.

src/commands/comment/view.ts
scottlovegrove: Fixed in 28254b0. Added a `the Notified line` block covering the rendered short names, the task route to the collaborator list (`getTask` → `getProject`), the direct project route for a project comment, the raw-ID fallback when `getProjectCollaborators` rejects, and that an empty recipient list neither prints the line nor looks anything up.

pawelgrimm (approved):
🚀

PR #486, opened 2026-08-19
feat(filter): add --description to filter create and update

craigcarlyle:
Adds `--description` to `filter create` and `filter update`.

- `--no-description` clears it, the same way `--no-favorite` works.
- Not shown in `filter list`. That is one line per filter and a description would wrap or truncate to nothing. It is already in `--json`.
- Blocked on Doist/todoist-sdk-typescript#668. Until that ships the compiler is not checking these args at all, so they look correct whether they are or not.

Backend: Doist/Todoist#29669


src/commands/filter/index.ts
```diff
@@ -29,6 +29,7 @@ export function registerFilterCommand(program: Command): void {
         .description('Create a filter')
         .option('--name <name>', 'Filter name (required)')
         .option('--query <query>', 'Filter query (required, e.g., "today | overdue")')
+        .option('--description <text>', 'Filter description')
```
doistbot, line 32: <a href="#">P2</a>  AGENTS.md requires `SKILL_CONTENT` in `src/lib/skills/content.ts` to be updated whenever flags are added in `src/commands/`. This PR adds `--description` to `filter create` and `--description`/`--no-description` to `filter update`, but the filter examples in `content.ts` (and the generated `skills/todoist-cli/SKILL.md`) still show `td filter create`/`td filter update` without those flags. Update `SKILL_CONTENT`, then run `npm run sync:skill`. Note the existing `check:skill-sync` CI check won't catch this because both files remain in sync with each other; the source-of-truth update is still required by the repo rule.

src/lib/api/filters.ts
```diff
@@ -44,6 +46,7 @@ export async function addFilter(args: AddFilterArgs): Promise<Filter> {
         id,
         name: args.name,
         query: args.query,
+        description: args.description ?? null,
```
doistbot, line 49: <a href="#">P2</a>  The added `description` won't appear in `filter create --json`. `createFilter` renders the result with `formatJson(filter, 'filter')`, which picks only `FILTER_ESSENTIAL_FIELDS` from `src/lib/output.ts`, and that list currently excludes `description`. So a filter created with `--description` is returned with the field, but the JSON output drops it. Add `'description'` to `FILTER_ESSENTIAL_FIELDS` (matching task/project/section, which all include it).

doistbot (commented):
Adds `--description` to `filter create` and `filter update`, with `--no-description` clearing it — a clean follow-through of the existing `--favorite`/`--no-favorite` pattern, including tests and dry-run coverage.

Few things worth tightening:
- Update `SKILL_CONTENT` in `src/lib/skills/content.ts` with the new flags (then `npm run sync:skill`) — AGENTS.md requires this for every flag change, and the current CI skill-sync check won't catch the drift since both files stay in sync.
- Add `description` to `FILTER_ESSENTIAL_FIELDS` in `src/lib/output.ts`, so `filter create --json` doesn't silently drop the field (task/project/section already include it).

<details>

- **<a href="#">P3</a> src/lib/api/filters.ts:76**: The `--no-description` test mocks `updateFilter` at the module boundary, so the actual wire serialization added here is never exercised. The important contract is that `description: null` reaches the SDK `filter_update` command (clears it) while an omitted key leaves it alone. A co-located `src/lib/api/filters.test.ts` test asserting `mockApi.sync` was called with a `filter_update` command containing `description: null` would catch a future change that filters out `null` or drops the field from `pickDefined`.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F486&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab5341570-a714-11f1-967b-b243e6bcb859%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab5341570-a714-11f1-967b-b243e6bcb859&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/filter/index.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-cli/commit/ff2628e

src/lib/api/filters.ts
craigcarlyle: Addressed in https://github.com/Doist/todoist-cli/commit/ff2628e

rmartins90:
✅

PR #492, opened 2026-08-25
fix: use the fetch the SDK pairs with its dispatcher

scottlovegrove:
Fixes a break that would otherwise land the moment the CLI picks up the new SDK. Depends on Doist/todoist-sdk-typescript#674, now merged and released as `@doist/todoist-sdk@14.2.0`, which this PR bumps to.

## The problem

The SDK builds a dispatcher for `HTTP_PROXY`/`HTTPS_PROXY` support and composes undici's `decompress` interceptor onto it. That dispatcher is only usable with the `fetch` it was paired with, because it decodes the response body itself — a `fetch` from a different undici build decodes it a second time.

`attachDispatcherIfNative` took the dispatcher from `getDefaultDispatcher()` and attached it to `globalThis.fetch`, which is backed by whatever undici ships inside the running Node release. That works today. From undici 7.29 onwards it does not: Node 26's built-in fetch reads `controller.rawHeaders`, still sees `content-encoding: gzip` on a body the interceptor already decompressed, and every API call fails with `TypeError: terminated` / `Error: incorrect header check`.

So the next SDK bump would have broken the CLI on Node 26 — a repeat of #318 — and nothing here would have caught it. Every test in the repo stubs fetch, so no test has ever seen a real compressed response.

## What changed

- `resolveRequestFetch` replaces `attachDispatcherIfNative`. It takes both halves from `getDefaultTransport()` and returns the fetch while attaching the dispatcher, so the two can only be chosen together. `createTrackedFetch` and `fetchTodoist` now default their fetch parameter to `undefined` rather than `globalThis.fetch`, which turns the old `fetchImpl !== globalThis.fetch` identity check into the question it was always standing in for: did the caller supply their own fetch? Explicit stubs still get no dispatcher.
- This fixes all three affected paths at once: the whole `TodoistApi` surface, the OAuth token exchange in `auth-provider.ts`, and the postinstall user fetch in `migrate-auth.ts`.
- `migrate-auth.ts` no longer pre-captures the global `fetch` binding. It worked by coincidence — the captured value was `=== globalThis.fetch` — and relying on that coincidence is what made the original bug subtle.
- **Verbose logging needed rescuing.** `patchGlobalFetch` only wraps `globalThis.fetch`, which API traffic no longer passes through, so `td --verbose` would have gone silent for every API request. The wrapper is now `withHttpLogging`, applied both to the global patch and to the paired fetch. The stale comment claiming the global patch is "the only way to get HTTP-level visibility for @doist/todoist-sdk" is updated.

## The test that would have caught this

`src/lib/usage-tracking.transport.test.ts` is the first CLI test to make a real HTTP request: a local `node:http` server serving gzip, deflate and brotli bodies over the real transport, with no SDK mock. It covers `createTrackedFetch`, `fetchTodoist`, a compressed binary body (the `viewAttachment` path) and the tracking headers.

Verified it has teeth — reverting `resolveRequestFetch` to `globalThis.fetch` and running on Node 26 gives:

```
× decodes a gzip-encoded response body
× decodes a deflate-encoded response body
× decodes a br-encoded response body
× decodes a compressed response through fetchTodoist
× reads a compressed binary body back byte for byte
TypeError: terminated
Caused by: Error: incorrect header check
```

The existing dispatcher tests are rewritten to assert the pairing rather than just the attachment, including that the global fetch is *not* called. New cases cover the Bun path (dispatcher with no paired fetch), the non-Node path, and a caller-supplied dispatcher being left alone.

## Review follow-ups

- **Verbose logging was untested** (P2) — fair, and it was the one behaviour this PR calls out as needing rescue. Covered now over the real paired transport, with the global patch undone so a log line can only come from the `withHttpLogging` wrap. Verified it fails when the wrap is dropped. First logger coverage in the repo.
- **Proxy env stubs leaking** (P3) — fixed with `vi.unstubAllEnvs()` in `afterAll`; `restoreMocks` does not undo `vi.stubEnv`.
- **Duplicated global-fetch stubbing** (P3) — folded into one `stubGlobalFetch()` helper.
- **Caching the wrapped paired fetch** (P3) — not doing it. One closure allocation per HTTP request is nothing against a network round trip, and a module-level or `WeakMap` cache would hold logger state across tests, where `resetLogger()` between cases is what keeps them honest. The cost is real but it buys a hazard rather than avoiding one.

## Verification

`npm test` (1878 tests), `npm run type-check` and `npm run check` all green on Node 24.18.0 and Node 26.4.0, against the **published** `@doist/todoist-sdk@14.2.0` — so with undici 7.29.0 actually resolved in the tree, which is the condition that breaks the old code.

Also checked end to end against the published package on both Node versions, with no mocks: a gzipped GET through `createTrackedFetch` succeeds, and the upload path (`fs.openAsBlob` → `customFetch` → paired undici fetch) sends real multipart framing with the file bytes and the tracking headers intact.

### DIFF
--- a/src/lib/logger.ts
+++ b/src/lib/logger.ts
@@ -150,19 +150,36 @@ export function initializeLogger(): void {
     const log = getLogger()
     log.initialize()
 
-    // When verbose, patch globalThis.fetch so SDK-internal HTTP calls are also logged.
-    // This is the only way to get HTTP-level visibility for @doist/todoist-sdk
-    // since it uses fetch() internally and doesn't expose hooks.
+    // When verbose, patch globalThis.fetch so HTTP calls made through it are
+    // logged. This covers everything that reaches the runtime's own fetch:
+    // the help centre client, and anything inside our dependencies that calls
+    // fetch directly. Todoist API traffic does not go through the global fetch
+    // — see `withHttpLogging`, which the transport layer applies itself.
     if (log.isEnabled()) {
         patchGlobalFetch()
     }
 }
 
+/** Whether HTTP logging should be applied to a fetch implementation. */
+export function isHttpLoggingEnabled(): boolean {
+    return getLogger().isEnabled()
+}
+
 /** Save original fetch and replace with logging wrapper. */
 function patchGlobalFetch(): void {
-    const originalFetch = globalThis.fetch
+    globalThis.fetch = withHttpLogging(globalThis.fetch)
+}
 
-    globalThis.fetch = async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+/**
+ * Wraps a fetch implementation with request and response logging.
+ *
+ * Used both to patch `globalThis.fetch` and, in the transport layer, to wrap
+ * the fetch the SDK pairs with its dispatcher. That one is undici's own fetch
+ * rather than the runtime's, so it never passes through the global patch and
+ * would otherwise be invisible to `--verbose`.
+ */
+export function withHttpLogging(originalFetch: typeof fetch): typeof fetch {
+    return async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
         const log = getLogger()
         const urlStr =
             typeof input === 'string' ? input : input instanceof URL ? input.toString() : input.url
--- a/src/lib/usage-tracking.ts
+++ b/src/lib/usage-tracking.ts
@@ -1,10 +1,7 @@
 import { randomUUID } from 'node:crypto'
-import {
-    type CustomFetch,
-    type CustomFetchResponse,
-    getDefaultDispatcher,
-} from '@doist/todoist-sdk'
+import { type CustomFetch, type CustomFetchResponse, getDefaultTransport } from '@doist/todoist-sdk'
 import packageJson from '../../package.json' with { type: 'json' }
+import { isHttpLoggingEnabled, withHttpLogging } from './logger.js'
 
 const CLI_NAME = 'todoist-cli'
 const CLI_VERSION = packageJson.version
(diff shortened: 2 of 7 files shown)


src/lib/usage-tracking.ts
```diff
@@ -76,21 +73,51 @@ function mergeTodoistHeaders(
...
+    // wrapped for logging by `initializeLogger` if verbose is on.
+    if (transport.fetch === undefined) return globalThis.fetch
+
+    // undici's fetch and the global fetch are the same call at runtime but
+    // carry different (undici vs DOM) types.
+    const pairedFetch = transport.fetch as unknown as typeof fetch
+
+    // The global patch in `initializeLogger` cannot see this fetch, so apply
+    // the same wrapper here or `--verbose` goes quiet for all API traffic.
+    return isHttpLoggingEnabled() ? withHttpLogging(pairedFetch) : pairedFetch
```
doistbot, line 120: <a href="#">P2</a>  The `isHttpLoggingEnabled() ? withHttpLogging(pairedFetch) : pairedFetch` branch is untested. This is the exact behavior the PR describes as needing rescue — without the `withHttpLogging` wrap, `td --verbose` goes silent for *all* API traffic, since API requests no longer pass through the global `patchGlobalFetch`. No existing test exercises this branch (there is no logger test, and neither the unit nor the transport tests enable logging), so a future regression that drops this wrap would ship silently. Add a test that enables the logger and asserts a request through `createTrackedFetch()`/`fetchTodoist` over the paired fetch still emits the `fetch ... => ...` log lines.

doistbot (commented):
This pairs the SDK’s dispatcher with the fetch it was built for so the next SDK bump won’t double-decode compressed responses on Node 26, and it re-wraps that fetch so `--verbose` still sees API traffic.

Few things worth tightening:
- The `withHttpLogging` wrap on the paired fetch is untested — add a case that enables the logger and asserts `createTrackedFetch()` / `fetchTodoist` still emit the `fetch ... => ...` lines, so a future drop of that wrap can’t ship silently.

<details>

- **<a href="#">P3</a> src/lib/usage-tracking.transport.test.ts:48**: Restore these environment stubs in `afterAll` with `vi.unstubAllEnvs()`. `restoreMocks` does not restore `vi.stubEnv`, so this suite leaves the worker's proxy variables unset and can alter later tests that use the runner's original proxy configuration.
- **<a href="#">P3</a> src/lib/usage-tracking.test.ts:271**: The `vi.spyOn(globalThis, 'fetch').mockImplementation(...)` block that captures `options` and returns a `Response('{}')` is duplicated verbatim in the two "uses the global fetch" tests (lines 271-280 and 296-305), and the same capture-and-respond shape also appears in `stubPairedTransport` and the "explicit command overrides" test. A small `stubGlobalFetch()` helper returning `{ getCaptured, spy }` would remove the copy-paste and keep the intent (capture the request options) in one place.
- **<a href="#">P3</a> src/lib/usage-tracking.ts:120**: `resolveRequestFetch` runs for every SDK request, so verbose mode creates a new `withHttpLogging(pairedFetch)` closure for every API call. Cache the wrapped paired fetch (for example, in a `WeakMap` keyed by `transport.fetch`, or alongside the resolved transport) so paginated or multi-request commands reuse it.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F492&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9ebee280-a06f-11f1-8ff8-376df977dbca%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9ebee280-a06f-11f1-8ff8-376df977dbca&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/usage-tracking.ts
scottlovegrove: Fair — that was the one behaviour I called out as needing rescue and then left uncovered.

Added a test in 0589702 to `usage-tracking.transport.test.ts`, so it runs over the real paired transport rather than a stub. It enables the logger, then restores `globalThis.fetch` immediately after `initializeLogger()`, which removes the alternative explanation: with the global patch undone, a `fetch GET … => 200` line can only have come from the wrapper applied to the paired fetch.

Verified it has teeth — replacing the branch with a bare `return pairedFetch` fails it on `expected … to match /fetch GET \\/json/`. This is also the first logger coverage in the repo.

src/lib/logger.ts
```diff
@@ -150,19 +150,36 @@ export function initializeLogger(): void {
...
-    globalThis.fetch = async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+/**
+ * Wraps a fetch implementation with request and response logging.
+ *
+ * Used both to patch `globalThis.fetch` and, in the transport layer, to wrap
+ * the fetch the SDK pairs with its dispatcher. That one is undici's own fetch
+ * rather than the runtime's, so it never passes through the global patch and
+ * would otherwise be invisible to `--verbose`.
+ */
+export function withHttpLogging(originalFetch: typeof fetch): typeof fetch {
```
craigcarlyle, line 181: There's no catch block here, should we have one? 🤔

craigcarlyle (commented):
✅

PR #494, opened 2026-08-26
feat!: build and parse Todoist URLs with the SDK

scottlovegrove:
## What

Takes `@doist/todoist-sdk` 15.0.0 and moves the CLI's Todoist URL handling onto it. The SDK is now the shared source of the URL rules (backed by `@doist/sdk-kmp`), so the CLI should not be assembling and matching links itself.

## Building

`lib/urls.ts` keeps every signature and hands each builder to the SDK. Two of these are links the SDK only gained in v15 — `getLabelUrl` and `getFilterUrl` — so the CLI no longer needs its own `entityUrl` at all.

`appInstallUrl` stays local; the SDK has no builder for install links.

## Parsing

`parseTodoistUrl` in `lib/refs.ts` delegates the entity part to the SDK, which extracts IDs properly instead of splitting the slug on its last hyphen. View links (`today`, `upcoming`) have no SDK equivalent and are still matched here.

The SDK recognises more kinds of link than the CLI has commands for, so two guards keep the existing behaviour:

- anything outside `task`, `project`, `label` and `filter` is still treated as unsupported;
- workspace-scoped links, such as a workspace filter, are still rejected. Without this the SDK parses them as filters and `td view <workspace filter URL>` would answer "Filter not found." instead of "Not a recognized Todoist URL".

## Breaking

`projectCommentUrl` now returns `/app/project/{id}#comment-{id}` rather than `/app/project/{id}/comments#comment-{id}`, to match the shared rules. Every other builder produces byte-identical output — checked against the previous implementation before the swap.

## Verification

1878 tests pass, `type-check`, `check` and `build` all clean.

### DIFF
--- a/src/lib/refs.ts
+++ b/src/lib/refs.ts
@@ -1,4 +1,9 @@
-import { type AppWithUserCount, TodoistApi, TodoistRequestError } from '@doist/todoist-sdk'
+import {
+    type AppWithUserCount,
+    parseTodoistUrl as parseSdkTodoistUrl,
+    TodoistApi,
+    TodoistRequestError,
+} from '@doist/todoist-sdk'
 import type { Project, Task } from './api/core.js'
 import {
     fetchWorkspaceFolders,
@@ -18,20 +23,21 @@ export interface ParsedTodoistUrl {
     id: string
 }
 
-const TODOIST_URL_PATTERN = new RegExp(
-    `^https?://app\\.todoist\\.com/app/(${URL_ENTITY_TYPES.join('|')})/([^?#]+)`,
-)
+function isUrlEntityType(value: string): value is UrlEntityType {
+    return (URL_ENTITY_TYPES as readonly string[]).includes(value)
+}
 
 export function parseTodoistUrl(url: string): ParsedTodoistUrl | null {
-    const match = url.match(TODOIST_URL_PATTERN)
-    if (!match) return null
+    const parsed = parseSdkTodoistUrl(url)
+    if (!parsed) return null
 
-    const entityType = match[1] as UrlEntityType
-    const slugAndId = match[2]
-    const lastHyphenIndex = slugAndId.lastIndexOf('-')
-    const id = lastHyphenIndex === -1 ? slugAndId : slugAndId.slice(lastHyphenIndex + 1)
+    // The SDK recognises more kinds of link than the CLI has commands for.
+    if (!isUrlEntityType(parsed.type)) return null
 
-    return { entityType, id }
+    // Workspace-scoped links (such as a workspace filter) have no CLI command.
+    if (parsed.workspaceId) return null
+
+    return { entityType: parsed.type, id: parsed.id }
 }
 
 const VIEW_TYPES = ['today', 'upcoming'] as const
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@doist/cli-core": "1.2.0",
-        "@doist/todoist-sdk": "14.2.0",
+        "@doist/todoist-sdk": "15.0.0",
         "@napi-rs/keyring": "1.3.0",
         "@pnpm/tabtab": "0.5.4",
         "chalk": "6.0.0",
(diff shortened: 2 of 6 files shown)


src/lib/refs.ts
```diff
@@ -18,20 +23,21 @@ export interface ParsedTodoistUrl {
...
+}
 
 export function parseTodoistUrl(url: string): ParsedTodoistUrl | null {
-    const match = url.match(TODOIST_URL_PATTERN)
-    if (!match) return null
+    const parsed = parseSdkTodoistUrl(url)
+    if (!parsed) return null
+
+    // The SDK recognises more kinds of link than the CLI has commands for.
+    if (!isUrlEntityType(parsed.type)) return null
```
doistbot, line 35: <a href="#">P2</a>  The new `isUrlEntityType` guard is untested. The SDK now recognizes types the CLI has no command for (e.g. `section`, since `getSectionUrl` is imported), and this guard is what keeps `parseTodoistUrl('https://app.todoist.com/app/section/...')` returning `null` instead of a `section` entity. Add a test asserting a section URL (or another SDK-only type) parses to `null`; without it, removing or weakening the guard would silently change `classifyTodoistUrl`/`lenientIdRef` behavior.

src/lib/refs.ts
```diff
@@ -18,20 +23,21 @@ export interface ParsedTodoistUrl {
...
 
-    const entityType = match[1] as UrlEntityType
-    const slugAndId = match[2]
-    const lastHyphenIndex = slugAndId.lastIndexOf('-')
-    const id = lastHyphenIndex === -1 ? slugAndId : slugAndId.slice(lastHyphenIndex + 1)
+    // Workspace-scoped links (such as a workspace filter) have no CLI command.
+    if (parsed.workspaceId) return null
```
doistbot, line 38: <a href="#">P2</a>  The workspace-scoped link rejection is untested. This guard is the only thing preventing `td view <workspace filter URL>` from being parsed as a filter and answering "Filter not found." instead of "Not a recognized Todoist URL" — a concrete user-visible contract called out in the PR description. Add a test that a workspace-scoped filter URL returns `null` from `parseTodoistUrl`.

doistbot (commented):
This swaps Todoist URL building and parsing onto `@doist/todoist-sdk` 15 while keeping CLI signatures and adding two guards so unsupported and workspace-scoped links still fail as before.

Few things worth tightening:
- The two guards that preserve existing parse behavior (`isUrlEntityType` and workspace-scoped rejection) have no tests; cover a section URL and a workspace filter URL so they keep returning `null` instead of leaking SDK-only types or a “Filter not found” path.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F494&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4442a400-a136-11f1-826d-307f55c205e4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4442a400-a136-11f1-826d-307f55c205e4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/refs.ts
scottlovegrove: Done in eb558d7 — added a case asserting a section URL and a bare workspace URL both parse to `null`, so the guard cannot be weakened without a failure.

src/lib/refs.ts
scottlovegrove: Done in eb558d7 — added a case asserting a workspace-scoped filter URL parses to `null`, pinning the "Not a recognized Todoist URL" behaviour rather than the "Filter not found." path.

rfgamaral (commented):
✅

craigcarlyle (commented):
✅

PR #495, opened 2026-08-26
feat: add ids-only output mode

gnapse:
Scripts and AI agents often need only IDs to pass into another command. Today they must parse human-readable output or request JSON and extract a field. `--ids-only` provides a small, stable, pipe-friendly output format for this use case.

## Summary

- add `--ids-only` to list commands with stable result IDs
- use `@doist/cli-core` 1.4.0 directly for ID output, global flag parsing, and output-mode validation
- keep stdout pipeable by sending pagination notices to stderr
- avoid unnecessary enrichment requests in IDs-only mode
- document the flag and update the bundled CLI skill

## Examples

```sh
# Print all task IDs, one per line
td task list --ids-only

# Print IDs for tasks in a specific project
td task list --project "Work" --ids-only

# Print all project IDs
td project list --ids-only

# Save IDs to a file; pagination notices remain on stderr
td task list --limit 50 --ids-only > task-ids.txt
```

Empty results produce no stdout. `--ids-only` is mutually exclusive with `--json` and `--ndjson`.


src/commands/workspace/users.ts
```diff
@@ -70,7 +73,16 @@ export async function listWorkspaceUsers(
     const users = allUsers.slice(0, targetLimit)
     const hasMore = allUsers.length > targetLimit || cursor !== undefined
 
-    if (options.json) {
+    if (outputMode === 'ids-only') {
+        outputIds(
+            users,
+            (user) => user.userId,
+            formatNextCursorNotice(hasMore ? (cursor ?? 'has-more') : null),
```
doistbot: <a href="#">P2</a>  This can print an IDs-only pagination notice after the final page. `cursor` is retained from an earlier response, so when role filtering requires a second request and that final response has no `nextCursor`, `hasMore` remains true even though all matching users were returned. Track/reset the continuation cursor based on the final response (or a separate `hasMore` flag) before passing a notice to `outputIds`.

src/lib/output.ts
```diff
@@ -599,8 +599,26 @@ export function formatPaginatedNdjson<T extends object>(
...
+    return chalk.dim('... more items exist. Use --all to fetch everything.')
+}
+
+export function outputIds<T>(
+    items: readonly T[],
+    getId: (item: T) => string,
+    paginationNotice = '',
+): void {
+    for (const item of items) {
+        console.log(getId(item))
```
doistbot: <a href="#">P2</a>  Respect stdout backpressure here. `--ids-only` supports `--all`, but this tight `console.log` loop cannot observe or await stream saturation, so piping a large list to a slow consumer can queue all remaining IDs in Node's stdout buffer. Make `outputIds` async and write buffered chunks while awaiting `drain`; await it from the command handlers.

doistbot (commented):
This adds `--ids-only` across list commands, with IDs on stdout and pagination notices on stderr so the output stays pipeable.

Few things worth tightening:
- In workspace user listing, a leftover cursor can emit an IDs-only “more items” notice after the last matching page; reset continuation from the final response.
- Make `outputIds` async and honor stdout backpressure so `--all` doesn’t flood a slow pipe.

<details>

- **<a href="#">P3</a> src/commands/notification/list.ts:47**: The `ids-only` branch reconstructs the exact same "showing X-Y of Z" pagination message (via `chalk.dim`) that the human-output branch still emits just below (lines 104-110). The two copies differ only in the leading `\n` and `nextOffset` vs. the inline `offset + limit`. Extract this once (e.g., a small local `const offsetNotice = hasMore ? \`... showing ${offset + 1}-${offset + notifications.length} of ${totalCount}. Use --offset ${offset + limit} to see more.\` : ''`) and reuse it in both the `outputIds` call and the trailing `if (hasMore)` block, so a future change to the message can't drift between the two paths.
- **<a href="#">P3</a> src/lib/output.test.ts:744**: This anchored assertion assumes `chalk.dim` emits no ANSI prefix, unlike the sibling `formatNextCursorFooter` test which uses `toContain` to survive colors. If chalk colors get enabled in the test environment (e.g. `FORCE_COLOR=1` in CI), `formatNextCursorNotice('some-cursor')` starts with `\x1b[2m` and `/^\.\.\. more items exist/` fails even though the function is correct. Prefer a color-safe check (e.g. `toContain('... more items exist')`) or assert the message doesn't start with `\n` after stripping ANSI.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F495&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A24e6a790-a169-11f1-8497-fa54d3ec216f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A24e6a790-a169-11f1-8497-fa54d3ec216f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/output.ts
gnapse: Good point. NDJSON output has the same long-standing behavior, so we will address stdout backpressure consistently in a separate follow-up PR rather than changing only IDs output here.

gnapse:
Addressed the three accepted review points in `2a5fba3`:

- reset workspace-user continuation state from the final response
- reuse the notification pagination notice
- make the pagination-notice test color-safe

Stdout backpressure remains deferred to a separate follow-up, as noted in its review thread.

scottlovegrove (commented):
Would it make sense to do the same change on `comms-cli`? If so, I wonder if it would be worth extracting the option/helpers to cli-core

gnapse:
> Would it make sense to do the same change on `comms-cli`? If so, I wonder if it would be worth extracting the option/helpers to cli-core

Makes sense, will do it.

scottlovegrove (approved):

src/lib/output.ts
gnapse: We explored this in a follow-up, including the existing NDJSON output. We have decided not to make this change for now. There are no reports of stdout backpressure causing a problem in practice, and the asynchronous output API would add broad complexity across command handlers and tests. We can revisit it if we get a real-world report or move to truly streaming large result sets.

PR #507, opened 2026-09-05
fix(text): slice and index name strings by code point, not code unit

### LINKED ISSUE #506: `formatUserShortName` abbreviates by UTF-16 code unit, so an emoji last name yields a lone surrogate
`formatUserShortName` in `src/lib/collaborators.ts` takes the last-name initial with `parts[parts.length - 1][0]`. Indexing a JS string that way returns one UTF-16 code unit, not one code point, so a last part that starts with an astral character (any emoji) gets split down the middle and the returned string carries an orphaned high surrogate.

A display name of `Omar | 🌴` returns `"Omar \ud83c."`.

That string is valid in JS and survives `--json`, so it reaches consumers intact. It is not valid anywhere that requires well-formed Unicode. Postgres `jsonb` rejects it with `invalid input syntax for type json` (detail: `Unicode low surrogate must follow a high surrogate`), which is how I found it: a service storing the output of `td completed list --full` stopped persisting for 29 hours, since one workspace member with a palm tree in their display name was inside its completed-tasks window.

Verified against the installed 5.2.1 bundle by calling the shipped function directly:

```
"Omar | 🌴"      -> "Omar \ud83c."   lone surrogate: true
"Ada Lovelace"   -> "Ada L."         lone surrogate: false
"Rui"            -> "Rui"            lone surrogate: false
"Yuki 🎌 Tanaka" -> "Yuki T."        lone surrogate: false
```

`Array.from(...)[0]` (or the spread, or `codePointAt`) takes a whole code point:

```js
const lastInitial = Array.from(parts[parts.length - 1])[0]
```

Same four inputs with that change: `Omar 🌴.`, `Ada L.`, `Rui`, `Yuki T.`. Only the broken case moves.

Worth checking whether the same indexing appears elsewhere. `src/lib/refs.ts:102` uses `charAt(0)` on an entity type, which is safe because those are ASCII, but a grep for `[0]` over user-supplied names would be the thing to run.

Happy to send a PR with the one-line change and a test over these four names if that is useful.

lmjabreu:
Closes #506.

## What was wrong

`formatUserShortName` took the last-name initial with `parts[parts.length - 1][0]`. Indexing a JS string that way returns one UTF-16 code unit rather than one code point, so a last part beginning with an astral character gets split in half. `Omar | 🌴` returned `"Omar \ud83c."`, carrying an orphaned high surrogate.

That string is valid in JS and survives `--json`, so it reaches consumers looking fine. It is not valid anywhere that requires well-formed Unicode: a lone surrogate is not a Unicode scalar value and has no UTF-8 representation at all. Postgres `jsonb` rejects it outright, which is how it surfaced. A service storing `td completed list --full` output stopped persisting for over a day because one workspace member has a palm tree in their display name.

## The sweep

Six sites share the defect, all now routed through a new `src/lib/text.ts`:

| site | what it cut |
| --- | --- |
| `lib/collaborators.ts` | `parts[last][0]`, the reported bug |
| `commands/comment/add.ts` | `content.slice(0, 80)` |
| `commands/comment/update.ts` (x2) | `slice(0, 80)` and `slice(0, 50)` |
| `commands/comment/delete.ts` | `slice(0, 50)` |
| `commands/activity.ts` | `firstLine.slice(0, maxLength - 3)` |

An earlier revision of this description claimed every other `slice(0, n)` in the tree was array indexing or ASCII. That was wrong. It came from a sweep matching the literal bounds `80` and `50`, which cannot see a computed one, and `activity.ts` was missed as a result. Thanks to @doistbot for catching it.

The sweep was redone matching computed bounds and excluding array indexing. Two further candidates came up and both are genuinely clear: the second `truncateContent` in `comment/helpers.ts` slices an array of lines rather than a string, and `completion.ts` cuts at `indexOf('=')`, which is never inside a surrogate pair. `completed/list.ts` and `reminder/helpers.ts` both slice ASCII dates.

Two deliberate details. `truncateForDisplay` counts and cuts in the same unit, because the old code tested `.length` in code units and changing only the slice would leave the two disagreeing on any string with an emoji. And in `activity.ts` the ellipsis counts toward `maxLength`, unlike the comment previews, so the helper gets a budget three short of it while the outer check stays in code units, since it decides only whether to cut and never where.

Code points, not grapheme clusters. A ZWJ sequence can still be split between its code points, but that produces an odd-looking **valid** string rather than an invalid one, so it is a cosmetic limit rather than this bug. Noted in the module comment for whoever needs `Intl.Segmenter`.

## Test plan

The risk is a helper that quietly does nothing, which looks identical to a working one. Every case has a control, and each guard was verified to fail against the original code rather than merely passing.

1. `npm test`
2. - [ ] Green: 73 files, 1905 tests.
3. `npx vitest run src/lib/collaborators.test.ts`
4. - [ ] The four names from #506: `Omar | 🌴` becomes `Omar 🌴.`, while `Ada L.`, `Rui` and `Yuki T.` stay byte-identical. Those three are the control, since the regression worth worrying about is ordinary names changing.
5. `npx vitest run src/lib/text.test.ts`
6. - [ ] An emoji straddling the cut survives whole, and the same content with the emoji clear of the boundary comes back `toBe(clear)`, not merely well-formed.
7. - [ ] 40 emoji (40 code points, 80 code units) are not truncated at a limit of 40.
8. `npx vitest run src/commands/activity.test.ts`
9. - [ ] The call-site guard: a real `td activity` run whose content puts an emoji across the 97th code point prints well-formed output. The helper being correct is not evidence the call site is, so this drives the command rather than the helper.
10. Mutation check.
11. - [ ] Reverting `activity.ts` to the code-unit slice, `firstCodePoint` to `text[0]`, or `truncateForDisplay` to the code-unit slice each turns the suite red.

`npm run check`, `type-check` and `build` are clean. The one `no-irregular-whitespace` warning is pre-existing in `src/commands/user/aliases.ts`, untouched here.

Follow-up filed as #508: the abbreviation also assumes `First Last`, so a display name carrying a status marker yields an initial from the status. Separate defect, not addressed here.

### DIFF
--- a/CODEBASE.md
+++ b/CODEBASE.md
@@ -180,6 +180,11 @@ New subcommand? Copy a sibling in the target group, wire it in that group's
 - **`pagination.ts`** — `paginate()`, `LIMITS` (tasks: 300, projects: 50, …)
 - **`completion.ts`** — `parseCompLine`, `getCompletions`,
   `withCaseInsensitiveChoices`, `withUnvalidatedChoices` (Commander tree-walker)
+- **`text.ts`** — `firstCodePoint`, `truncateForDisplay`. Code-point-safe
+  indexing and truncation. Reach for these rather than `name[0]` or
+  `text.slice(0, n)` on anything user-supplied: those cut by UTF-16 code unit
+  and split an emoji into a lone surrogate, which has no UTF-8 form and is
+  rejected by anything requiring well-formed text.
 - **`spinner.ts`** — `startEarlySpinner`, `LoadingSpinner` class
   (yocto-spinner wrapper)
 - **`markdown.ts`** — `preloadMarkdown`, markdown → terminal renderer
--- a/src/commands/activity.test.ts
+++ b/src/commands/activity.test.ts
@@ -59,6 +59,43 @@ describe('activity command', () => {
         expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining('Buy groceries'))
     })
 
+    it('does not split an emoji when truncating long activity content', async () => {
+        // The call-site guard for `truncateContent`. The default limit is 100 and
+        // the cut lands at 97, so this content puts a palm tree exactly across
+        // that boundary, which is where a code-unit slice used to bisect it.
+        const straddling = `${'x'.repeat(96)}\u{1F334} and more text past the limit`
+        const program = createProgram()
+
+        mockApi.getActivityLogs.mockResolvedValue({
+            results: [
+                {
+                    id: 'event-1',
+                    objectType: 'task',
+                    objectId: 'task-1',
+                    eventType: 'added',
+                    eventDate: new Date('2025-01-10T14:30:00Z'),
+                    parentProjectId: 'proj-1',
+                    parentItemId: null,
+                    initiatorId: 'user-1',
+                    extraData: { content: straddling },
+                },
+            ],
+            nextCursor: null,
+        })
+        mockApi.getProjects.mockResolvedValue({
+            results: [{ id: 'proj-1', name: 'Shopping' }],
+            nextCursor: null,
+        })
+
+        await program.parseAsync(['node', 'td', 'activity'])
+
+        const printed = consoleSpy.mock.calls.flat().join('\n')
+        expect(printed.isWellFormed()).toBe(true)
+        // It truncated (so the guard is exercised) and kept the emoji whole.
+        expect(printed).toContain('...')
+        expect(printed).toContain('\u{1F334}')
+    })
+
     it('shows "No activity found" when empty', async () => {
         const program = createProgram()
 
--- a/src/commands/activity.ts
+++ b/src/commands/activity.ts
@@ -14,6 +14,7 @@ import type { Pagination } from '../lib/options.js'
 import { formatPaginatedJson, formatPaginatedNdjson } from '../lib/output.js'
 import { paginate } from '../lib/pagination.js'
 import { extractId, isIdRef, resolveProjectId } from '../lib/refs.js'
+import { truncateForDisplay } from '../lib/text.js'
 
 interface ActivityOptions extends Pagination {
     since?: string
@@ -93,10 +94,10 @@ function getEventContent(event: ActivityEvent): string {
 
 function truncateContent(content: string, maxLength = 100): string {
     const firstLine = content.split('\n')[0]
-    if (firstLine.length <= maxLength) {
-        return firstLine
-    }
-    return `${firstLine.slice(0, maxLength - 3)}...`
+    // Unlike the comment previews, the ellipsis counts toward `maxLength` here,
+    // so the helper's budget is three short of it. The outer check stays in code
+    // units: it only decides WHETHER to cut, never where, so it cannot split a pair.
+    return firstLine.length <= maxLength ? firstLine : truncateForDisplay(firstLine, maxLength - 3)
 }
 
 function formatActivityRow(
--- a/src/commands/comment/add.ts
+++ b/src/commands/comment/add.ts
@@ -13,6 +13,7 @@ import { openLocalFileAsBlob } from '../../lib/local-file.js'
 import { formatJson, printDryRun } from '../../lib/output.js'
 import { resolveProjectRef, resolveTaskRef } from '../../lib/refs.js'
 import { readStdin } from '../../lib/stdin.js'
+import { truncateForDisplay } from '../../lib/text.js'
 
 interface AddOptions {
     content?: string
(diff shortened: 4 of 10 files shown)


doistbot (commented):
Fixes string slicing and indexing to operate on code points instead of UTF-16 code units, so astral characters like emoji no longer get split into lone surrogates in `formatUserShortName` and comment preview truncations; the shared helpers live in a new `src/lib/text.ts` with colocated tests. No inline issues were flagged.

Few things worth tightening:
- There's one more same-class truncation site in `src/commands/activity.ts` (`truncateContent` slices with `firstLine.slice(0, maxLength - 3)`), which the PR's sweep appears to have missed — worth folding into the new helper or at least noting as a follow-up.
- `src/lib/text.ts` is a new broadly-reusable lib module and should be registered in `CODEBASE.md` per the repo's documented update rule.
- The tests hand-roll a lone-surrogate check; `String.prototype.isWellFormed()` covers this out of the box and would keep the tests shorter.

<details>

- **<a href="#">P3</a> src/lib/text.ts:38**: Follow-up: the PR description says all other `slice(0, n)` hits are array indexing or ASCII, but `src/commands/activity.ts:94-99` (`truncateContent`) still slices user-provided activity content with `firstLine.slice(0, maxLength - 3)` — the same code-unit defect this PR fixes, reachable whenever an event's content has an emoji straddling the cut. Worth a follow-up that reuses `truncateForDisplay` there.
- **<a href="#">P3</a> src/lib/text.ts:25**: Per AGENTS.md, adding a broadly-reusable helper in `src/lib/` is a trigger to update CODEBASE.md's "don't reimplement" catalog and `src/` tree. `text.ts` (used by four command files plus collaborators.ts) isn't listed there yet.
- **<a href="#">P3</a> src/lib/text.test.ts:13**: This `LONE_SURROGATE` regex is duplicated verbatim in `src/lib/collaborators.test.ts:13` and reimplements `String.prototype.isWellFormed()`. The project targets ES2024 and runs on Node >= 24, so assertions like `expect(truncateForDisplay(...).isWellFormed()).toBe(true)` do the same check with the standard API and eliminate the duplicated constant in both test files.
- **<a href="#">P3</a> src/lib/text.ts:27**: `Array.from(text)[0]` builds an array of every code point in the string just to take the first one. Use `text.codePointAt(0)` with `String.fromCodePoint`, or `for (const ch of text) return ch` followed by `return ''`, to get the first code point without materializing the whole string.
- **<a href="#">P3</a> src/lib/text.ts:39**: `truncateForDisplay` always materializes the entire string as a code-point array, including the common already-fits path that used to be an O(1) `.length` check. Since UTF-16 length is always >= the code-point count, short-circuit with `if (text.length <= maxCodePoints) return text` before `Array.from`. For the over-limit path, iterating with `for...of` and stopping after `maxCodePoints` avoids building the full array for long inputs.
- **<a href="#">P3</a> src/lib/text.test.ts:56**: The comment says the emoji-clear-of-boundary case is "unaffected either way", but the assertion only checks for no lone surrogate — it would still pass if the input were mangled into a different valid string. Assert the stated invariant: `expect(truncateForDisplay(clear, 80)).toBe(clear)`. (76 code points fit under the limit, so this should hold.)

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F507&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5ac8dc70-a9d4-11f1-999f-a1ec204a8446%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5ac8dc70-a9d4-11f1-999f-a1ec204a8446&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (approved):

gnapse (approved):
Feedback about the PR description. Too lengthy and technical. I would've preferred it to be mostly a table comparing the before and after of what this does to certain names, to show how it is fixing the issue.

PR #511, opened 2026-09-07
docs: add extensions spec

scottlovegrove:
## Summary

Design spec for a `gh`-style extension system for `td`, following on from the internal discussion about plugins for the CLI. Draft for discussion, no code in this PR.

The proposal: an extension is an executable named `td-<name>` in a per-user directory; `td <name> …` spawns it with the remaining arguments passed through verbatim. Extensions install from a GitHub repo (git clone or prebuilt release binary) or a local path, and are managed with `td extension install|list|upgrade|remove|exec`. Because the contract is the process boundary rather than a JS API, extensions can be written in any language and get Todoist data by calling `td … --json`. The generic manager is proposed for `@doist/cli-core` so `tdc` and `tda` can adopt it too.

## What to look at

- The **Prior art** table for where this deliberately departs from `gh` (env contract, manifest descriptions, Node handling, checksum verification, trust warning ordering).
- **Environment contract** for what the child process gets, and why the token is deliberately not injected.
- **Phasing** for the proposed MVP versus later work.
- **Open questions** at the end: cli-core first or local first, an official-Doist marker, Windows as a phase 1 blocker, telemetry, and `extension` vs `plugin` naming.

### DIFF
--- a/docs/specs/extensions.md
+++ b/docs/specs/extensions.md
@@ -1,0 +1,345 @@
+# Extensions for the Todoist CLI
+
+## Status
+
+Draft — for discussion.
+
+## Summary
+
+Add a `gh`-style extension system to `td`. An extension is an executable named `td-<name>` that lives in a per-user directory. Running `td <name> …` executes it with the remaining arguments untouched. Extensions are installed from a GitHub repository (either a git clone or a prebuilt release binary) or from a local directory, and are managed with `td extension install|list|upgrade|remove|exec`.
+
+Because the contract between `td` and an extension is the process boundary (argv, environment, stdin/stdout/stderr, exit code) rather than a JavaScript API, extensions can be written in any language: a Node script, a shell script, a Go or Rust binary, a Python program. Anything that can be executed works. Extensions that need Todoist data do what a user would do: they call `td` itself (`td task list --json`, `td auth token view`), so they inherit authentication, account selection, and output conventions without a shared library.
+
+The dispatch and management logic is generic, so it belongs in `@doist/cli-core` where `tdc` and `tda` can adopt it with a one-line registration. `td` is the first host.
+
+## Motivation
+
+From the original proposal:
+
+> It would make it a lot easier to test new features on Todoist (remember `td goals`??). It would also allow people to add things to the CLI that we may not have thought of, or are not prepared to maintain ourselves.
+
+Concretely:
+
+1. **Prototyping.** A feature behind a server-side flag, or one that is still being designed, can ship as `Doist/td-goals` today and be used by the people testing it, without a release of `td` and without dead code in `main` when the experiment ends. If the feature graduates, the extension is either promoted into core or stays an extension.
+2. **Community.** People already script `td` in shell and Python. An extension mechanism gives those scripts a home (`td standup`, `td weekly-review`, `td export-obsidian`) and a discovery path, with no obligation on Doist to maintain them.
+3. **Consistency with the other Doist CLIs.** One implementation in cli-core means `tdc` and `tda` gain the same feature for free, with the same commands and the same on-disk layout.
+
+### Non-goals
+
+- In-process JavaScript plugins that register Commander commands directly. See "Alternatives considered".
+- A marketplace, ratings, or any hosted registry. Discovery is a GitHub topic, like `gh`.
+- Sandboxing. An extension runs with the user's privileges, exactly like any program the user installs. The spec is explicit about this in the install warning.
+
+## Prior art: how `gh` does it
+
+Everything below is modelled on `gh extension` (cli/cli `pkg/cmd/extension`), which has been stable since 2021 and is familiar to the audience most likely to write extensions. Where this spec departs from `gh` it says so and why. The main departures:
+
+| Area                        | `gh`                                       | This spec                                                                                                                                                    |
+| --------------------------- | ------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| Context passed to the child | Only `GH_EXTENSION=1` (added Aug 2026)     | `TD_EXTENSION`, `TD_PATH`, `TD_VERSION`, `TD_EXTENSION_NAME`, `TD_EXTENSION_DIR`, `TD_CONFIG_DIR`, and the resolved `--user`                                 |
+| Help text for extensions    | Fixed string `Extension <name>`            | Optional `td-extension.json` manifest supplies a description and a minimum host version                                                                      |
+| Node extensions             | No special handling; deps must be vendored | If the clone has a `package.json`, install runs `npm ci --omit=dev`; shebang `#!/usr/bin/env node` scripts run under the host's own `node` on every platform |
+| Release asset verification  | None                                       | Verify against a `checksums.txt` / `*.sha256` release asset when one exists; warn when it does not                                                           |
+| Argument completion         | None                                       | Optional, opt-in protocol (phase 3)                                                                                                                          |
+
+## Terminology
+
+The user-facing word is **extension**, not plugin. It matches `gh`, it avoids implying in-process loading, and `td ext` is a comfortable short form. The proposal thread uses "plugin"; this spec treats them as synonyms and standardises on "extension" in code and docs.
+
+## User experience
+
+### Running an extension
+
+```bash
+td goals                     # runs td-goals with no args
+td goals add "Ship v5" --by 2026-12-01
+td goals --help              # forwarded verbatim; td never interprets args after the name
+td --user scott@doist.com goals list   # global flags before the name are td's
+```
+
+Rules:
+
+- Everything after the extension name is passed to the extension verbatim. `td` does not parse, validate, or reorder it. That includes `--help`, `--json`, and `--user`.
+- Global `td` flags that appear **before** the extension name are consumed by `td` and translated into environment variables for the child (see "Environment contract"). `--user` is the important one.
+- The child's stdin, stdout, and stderr are the parent's. `td` adds nothing to either stream on success, so `td goals list --json | jq` works.
+- `td` exits with the extension's exit code. If the extension is killed by a signal, `td` exits `128 + signal` in the usual way.
+- Spinners are not started for extension dispatch; the extension owns its terminal.
+
+### Managing extensions
+
+```
+td extension install <source> [--pin <ref>] [--force]
+td extension list [--json]
+td extension upgrade [<name> | --all] [--force] [--dry-run]
+td extension remove <name>
+td extension exec <name> [args...]
+```
+
+`td ext` is an alias for `td extension`. All subcommands support `--json` where they produce data, following the repo's existing conventions (`formatJson`, `CliError` codes, `--dry-run` previews with the standard `[dry-run] Would …` block).
+
+#### `install <source>`
+
+`<source>` is one of:
+
+| Form          | Example                                        | Result                                                                            |
+| ------------- | ---------------------------------------------- | --------------------------------------------------------------------------------- |
+| `owner/repo`  | `Doist/td-goals`                               | GitHub. Binary release if the latest release has a matching asset, else git clone |
+| Full URL      | `https://github.com/scottlovegrove/td-standup` | Same as above; also allows non-github.com hosts via git                           |
+| `.` or a path | `td extension install .`                       | Local development install: a symlink to the directory                             |
+
+Repository names must start with `td-`. The command name is the repository name without the prefix.
+
+Install steps for a GitHub source:
+
+1. Validate the name: must match `^td-[a-z0-9][a-z0-9-]*$`, and the command name must not match a built-in command or alias. Refuse with `EXTENSION_NAME_RESERVED` and a hint pointing at `td extension exec` if the extension is already installed and a later `td` release added a core command with the same name.
+2. If an extension with this command name is already installed from a different owner, refuse with `EXTENSION_ALREADY_INSTALLED` (`--force` replaces it). Same owner: report already installed, suggest `upgrade`.
+3. Query `GET /repos/{owner}/{repo}/releases/latest`. If it has an asset whose name ends in `<platform>-<arch>[.exe]` for the current machine, treat it as a **binary extension**: download the asset to `<dir>/td-<name>[.exe]`, `chmod 0755`, verify the checksum when the release also carries `checksums.txt` or `<asset>.sha256`, and write `manifest.json`. Platform names use Node's `process.platform` / `process.arch` values (`linux-x64`, `darwin-arm64`, `win32-x64`) rather than Go's, since extension authors building with Node will already have those in hand. A release workflow template (phase 2) produces both spellings so a single repo can serve both `gh` and `td` if the author wants.
+4. Otherwise, confirm the repository exists and has a root file named `td-<name>`, then `git clone` it. This is a **script extension**. With `--pin <ref>`, resolve the ref, check it out, and record the SHA.
+5. If the clone contains `package.json`, run `npm ci --omit=dev` (falling back to `npm install --omit=dev` when there is no lockfile) inside it. `npm` is guaranteed to exist because `td` is itself installed with it. Failure here is a hard install failure with the npm output attached as hints.
+6. Print the trust warning (below) and the install location.
+
+`GH_TOKEN` / `GITHUB_TOKEN` are honoured for the GitHub API and asset downloads so that private repositories work. This matters for Doist-internal prototypes. Git clones use the user's own git credential setup.
+
+The trust warning is printed on every install and upgrade:
+
+```
+Installed td-goals from Doist/td-goals (v0.3.0) to ~/.local/share/todoist-cli/extensions/td-goals
+
+Extensions are not reviewed, signed, or endorsed by Doist. Installing one runs
+code from its publisher with your permissions and your Todoist credentials.
+Review the source before use.
+```
+
+#### `list`
+
+```
+NAME        SOURCE                       VERSION   KIND
+goals       Doist/td-goals               v0.3.0    binary
+standup     scottlovegrove/td-standup    a1b2c3d4  git (pinned)
+scratch     ~/code/td-scratch            —         local
+weekly      example/td-weekly            0.1.0     git · shadowed by built-in "weekly"
+```
+
+No network access. `--json` returns the manifest fields plus `path`, `kind`, `shadowed`, and `executable` (false when the file is missing or not executable, which is the usual state of a freshly cloned Go extension that has not been built).
+
+#### `upgrade`
+
+- Git extensions: `git pull --ff-only`; with `--force`, `git fetch` and `git reset --hard origin/HEAD`. Re-run the `npm ci` step if `package.json` changed.
+- Binary extensions: repeat the release lookup and download when the tag differs.
+- Pinned extensions are skipped with a note unless `--force`. Local extensions are always skipped.
+- `--all` checks every extension, fetching release metadata concurrently under a single spinner.
+- `--dry-run` reports what would change and exits 0.
+
+#### `remove`
+
+Deletes the extension directory and its state file. Accepts `name`, `td-name`, or `owner/td-name`. Requires no confirmation; the operation is reversible by reinstalling and the repo convention is `--yes` only for data-destroying commands.
+
+#### `exec`
+
+`td extension exec <name> [args...]` dispatches directly, bypassing the built-in command table. It is the escape hatch for the shadowed case and for scripting where an explicit form is preferable.
+
+### Help and discovery
+
+- `td --help` gains an `Extensions:` section listing installed extensions with the description from their manifest, or `Extension <name>` when there is none. This is where `gh` is weakest, and a one-line description is cheap for authors to provide.
+- `td <name> --help` is forwarded to the extension, which is expected to print its own usage.
+- `td extension --help` includes the trust warning and a pointer to the authoring guide.
+- Discovery uses the GitHub topic `td-extension`. `td extension search` (phase 2) wraps the GitHub search API over that topic, filters to `td-` prefixed repos, and marks installed ones.
+
+## On-disk layout
+
+```
+${XDG_DATA_HOME:-~/.local/share}/todoist-cli/extensions/
+├─ td-goals/
+│  ├─ td-goals              # executable (binary or script)
+│  └─ manifest.json         # written by td for binary installs
+├─ td-standup/              # git clone
+│  ├─ .git/
+│  ├─ td-standup
+│  ├─ td-extension.json     # optional, authored by the extension
+│  ├─ package.json          # optional; triggers npm ci on install/upgrade
+│  └─ node_modules/
+└─ td-scratch -> /home/scott/code/td-scratch   # local install (symlink)
+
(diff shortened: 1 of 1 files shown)


docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+| `TD_EXTENSION_NAME`                         | `goals`                              | The command name, so one executable can be installed under several names                                                                                                                                                  |
+| `TD_EXTENSION_DIR`                          | absolute path                        | The extension's own directory, for locating bundled assets                                                                                                                                                                |
+| `TD_PATH`                                   | absolute path to the `td` executable | Call the host without depending on `PATH`. Set to `process.argv[1]` resolved, with `process.execPath` available as `TD_NODE` for scripts that need to spawn Node explicitly                                               |
```
doistbot: <a href="#">P2</a>  TD_PATH is specified as `process.argv[1]` resolved, but argv[1] is the JS entry file (dist/index.js), not a launcher. On POSIX that works because dist/index.js has a node shebang and an exec bit. On Windows, an npm install creates `td.cmd` shims, so argv[1] is still the bare `.js` file — which CreateProcess can't execute. The documented calling patterns fail there: `"$td" task list …` (L232) under sh.exe, and `execFile(td, ['project', 'list', …])` (L239) both error with a shebang script path. Suggest the spec state how a Windows extension calls the host, e.g. document `spawn(process.env.TD_NODE, [TD_PATH, …])` in the authoring guide, or define TD_PATH as the shim/launcher when one exists.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+2. If an extension with this command name is already installed from a different owner, refuse with `EXTENSION_ALREADY_INSTALLED` (`--force` replaces it). Same owner: report already installed, suggest `upgrade`.
+3. Query `GET /repos/{owner}/{repo}/releases/latest`. If it has an asset whose name ends in `<platform>-<arch>[.exe]` for the current machine, treat it as a **binary extension**: download the asset to `<dir>/td-<name>[.exe]`, `chmod 0755`, verify the checksum when the release also carries `checksums.txt` or `<asset>.sha256`, and write `manifest.json`. Platform names use Node's `process.platform` / `process.arch` values (`linux-x64`, `darwin-arm64`, `win32-x64`) rather than Go's, since extension authors building with Node will already have those in hand. A release workflow template (phase 2) produces both spellings so a single repo can serve both `gh` and `td` if the author wants.
```
doistbot: <a href="#">P2</a>  Define `--pin` handling for binary installs here. This step always queries `releases/latest` and downloads its asset, so `td extension install owner/td-name --pin v1.2.3` silently installs the latest binary whenever one is available. Resolve the requested release tag for binary assets (or explicitly fall back to a pinned git checkout) so the pinned state described for upgrade is real.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+
+- `description` is shown in `td --help` and `td extension list`.
+- `requires.td` is a semver range. When the running `td` does not satisfy it, dispatch still happens, but a warning goes to stderr first. Refusing outright would make a `td` upgrade break a working extension for no good reason; the extension can enforce it itself using `TD_VERSION` if it must.
+- `completion` opts in to the argument-completion protocol (phase 3).
+
+For Node extensions the same fields may instead be supplied under a `"td"` key in `package.json`, so authors do not need two metadata files. `td-extension.json` wins when both exist.
+
+## Environment contract
+
+The child process inherits the parent's environment plus:
```
doistbot, line 216: <a href="#">P2</a>  This inheritance rule still exposes `TODOIST_API_TOKEN` to extensions when a user invokes `TODOIST_API_TOKEN=... td <extension>`, despite the following section saying the token is deliberately withheld to avoid passing it to descendants. Either sanitize that variable in the spawned environment, with a defined way for env-token users to make nested `td` calls, or document this as an explicit exception to the no-token contract.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+
+Each discovered extension whose name does not collide with a built-in command is registered as a Commander command with:
+
+- `.description()` from its manifest, or `Extension <name>`.
+- `.allowUnknownOption()`, `.passThroughOptions()`, and `.helpOption(false)`, so Commander touches nothing after the name.
+- An action that spawns the executable.
+
+Registering real commands, rather than catching the unknown-command error, is what `gh` does and it is the right call here too: it makes `--help` and name completion work with no special cases, and it keeps the existing dispatcher's "find the command token in argv" logic intact. A colliding name is skipped and surfaces as `shadowed` in `list`.
+
+The existing pre-parse `--user` stripping in `src/index.ts` needs one adjustment: it must only consider arguments before the extension name, since anything after it is opaque.
```
doistbot: <a href="#">P2</a>  Restricting only `stripUserFlag` is insufficient. `getRequestedUserRef()` currently parses the whole argv via `parseTdLocalFlags`, so `td goals --user extension-value` would still be treated as the host selector and exported as `TD_USER`, changing nested `td` calls even though this spec declares post-name arguments opaque. Scope global `--user` parsing itself to tokens before the discovered extension name.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+3. Query `GET /repos/{owner}/{repo}/releases/latest`. If it has an asset whose name ends in `<platform>-<arch>[.exe]` for the current machine, treat it as a **binary extension**: download the asset to `<dir>/td-<name>[.exe]`, `chmod 0755`, verify the checksum when the release also carries `checksums.txt` or `<asset>.sha256`, and write `manifest.json`. Platform names use Node's `process.platform` / `process.arch` values (`linux-x64`, `darwin-arm64`, `win32-x64`) rather than Go's, since extension authors building with Node will already have those in hand. A release workflow template (phase 2) produces both spellings so a single repo can serve both `gh` and `td` if the author wants.
+4. Otherwise, confirm the repository exists and has a root file named `td-<name>`, then `git clone` it. This is a **script extension**. With `--pin <ref>`, resolve the ref, check it out, and record the SHA.
+5. If the clone contains `package.json`, run `npm ci --omit=dev` (falling back to `npm install --omit=dev` when there is no lockfile) inside it. `npm` is guaranteed to exist because `td` is itself installed with it. Failure here is a hard install failure with the npm output attached as hints.
```
doistbot: <a href="#">P1</a>  `npm ci` executes lifecycle scripts from the untrusted repository, but the trust warning is only printed in the following step. A malicious extension can therefore run code before the user is told to review it. Print the warning before any step that can execute repository-controlled code (and document whether lifecycle scripts are intentionally allowed).

docs/specs/extensions.md
doistbot: <a href="#">P2</a>  This flow downloads only the binary and writes td-owned `manifest.json`. It never obtains the author’s `td-extension.json`, while the generated manifest schema has no `description` or `requires` fields. As a result, binary installs cannot provide the manifest description in help/list or the minimum-version warning. Persist those author metadata fields during binary installation (for example, by fetching the repository manifest), or define a metadata release asset.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+const projects = JSON.parse(stdout)
+```
+
+Because `td` already treats `--json` output and `CliError` JSON envelopes as a stable contract for AI agents, extensions get the same guarantee.
+
+## Dispatch design
+
+### Startup cost
+
+`td` starts by registering placeholders for its built-in commands without importing them. Extensions are discovered by a single `readdir` of the extensions directory, which costs well under a millisecond and returns nothing for the majority of users who have no extensions. Discovery runs on every invocation because it is needed for `--help`, for completion, and for the unknown-command path. Manifests are read lazily and only for `list`, `--help`, and the shadowing check.
```
doistbot: <a href="#">P2</a>  The stated lazy-read policy excludes dispatch, but dispatch must read `td-extension.json` to emit the `requires.td` warning specified above. Include loading the selected extension’s manifest before spawning in this design; otherwise the compatibility-warning contract cannot be implemented.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+2. If an extension with this command name is already installed from a different owner, refuse with `EXTENSION_ALREADY_INSTALLED` (`--force` replaces it). Same owner: report already installed, suggest `upgrade`.
+3. Query `GET /repos/{owner}/{repo}/releases/latest`. If it has an asset whose name ends in `<platform>-<arch>[.exe]` for the current machine, treat it as a **binary extension**: download the asset to `<dir>/td-<name>[.exe]`, `chmod 0755`, verify the checksum when the release also carries `checksums.txt` or `<asset>.sha256`, and write `manifest.json`. Platform names use Node's `process.platform` / `process.arch` values (`linux-x64`, `darwin-arm64`, `win32-x64`) rather than Go's, since extension authors building with Node will already have those in hand. A release workflow template (phase 2) produces both spellings so a single repo can serve both `gh` and `td` if the author wants.
+4. Otherwise, confirm the repository exists and has a root file named `td-<name>`, then `git clone` it. This is a **script extension**. With `--pin <ref>`, resolve the ref, check it out, and record the SHA.
```
doistbot: <a href="#">P2</a>  Avoid confirming repository existence and checking the root file before cloning. The clone already establishes whether the repository is reachable, and the expected executable can be validated locally afterward (ideally in a staging directory before moving it into place). The preflight adds an extra remote request to every script-extension install.

docs/specs/extensions.md
```diff
@@ -0,0 +1,345 @@
...
+```
+
+No network access. `--json` returns the manifest fields plus `path`, `kind`, `shadowed`, and `executable` (false when the file is missing or not executable, which is the usual state of a freshly cloned Go extension that has not been built).
+
+#### `upgrade`
+
+- Git extensions: `git pull --ff-only`; with `--force`, `git fetch` and `git reset --hard origin/HEAD`. Re-run the `npm ci` step if `package.json` changed.
+- Binary extensions: repeat the release lookup and download when the tag differs.
+- Pinned extensions are skipped with a note unless `--force`. Local extensions are always skipped.
+- `--all` checks every extension, fetching release metadata concurrently under a single spinner.
```
doistbot: <a href="#">P2</a>  Specify a bounded concurrency limit for `--all`. Installed extensions are user-controlled in number, so issuing all release lookups at once can create an unbounded request burst and trigger GitHub secondary rate limits. Use a small worker pool while retaining the single spinner.

doistbot (commented):
This PR adds a docs-only design spec for a `gh`-style extension system for `td` — installable `td-<name>` executables dispatched by name, managed via `td extension install|list|upgrade|remove|exec`, with a generic manager proposed for `@doist/cli-core`. The spec is well-researched and its departures from `gh` are justified, but a few flows need tightening before implementation.

Few things worth tightening:

- **Trust-before-execution:** `npm ci` runs repository-controlled lifecycle scripts before the trust warning is printed — move the warning ahead of any step that can execute untrusted code. Relatedly, the environment-inheritance rule still leaks `TODOIST_API_TOKEN` to extensions when the user sets it inline, contradicting the stated no-token contract; either sanitize it or document the exception.
- **Contract gaps:** `TD_PATH` points at `dist/index.js`, which isn't spawnable on Windows via npm shims — define the calling pattern for Windows authors. `--pin` on binary installs silently installs the latest release, so pinning isn't real. And global `--user` parsing should be scoped to tokens before the extension name, or post-name args (declared opaque) can still leak into `TD_USER`.
- **Spec completeness:** binary installs never obtain the author's `td-extension.json`, so `description` and `requires` can't be surfaced; and the lazy-read manifest policy needs to include loading the selected extension's manifest at dispatch for the `requires.td` warning to work.
- **Operational polish:** drop the pre-clone repository/root-file checks (the clone already validates reachability), and bound the concurrency of `--all` upgrades with a small worker pool to avoid GitHub secondary rate limits.

<details>

- **<a href="#">P3</a> docs/specs/extensions.md:88**: The name rules only cover GitHub sources: the `^td-[a-z0-9][a-z0-9-]*$` check and "repository names must start with `td-`" (L90) sit in the GitHub install steps, but local installs also produce a command name. Specify where a local install's name comes from (directory name? the `td-<name>` root executable?) and whether the same validation and `EXTENSION_NAME_INVALID` error apply — otherwise a local directory like `~/code/my-tool` has undefined behavior. The authoring guide (L292) implies the directory must be named `td-<name>`, so make that explicit here.
- **<a href="#">P3</a> docs/specs/extensions.md:98**: 'npm is guaranteed to exist because td is itself installed with it' isn't a guarantee: Node installs without npm are real (e.g. Debian/Ubuntu ship npm as a separate package, corepack-only setups, pnpm/yarn installs, direct dist/index.js runs). The npm ci step should treat a missing npm binary as a distinct failure mode with its own hint (how to install npm or vendor dependencies), instead of relying on the guarantee.
- **<a href="#">P3</a> docs/specs/extensions.md:213**: `TD_NODE` is promised here but is absent from the environment-contract table. Since the child otherwise only inherits the parent environment, extension authors cannot rely on it being set. Add it as an explicitly injected variable with its value, or remove this guarantee.
- **<a href="#">P3</a> docs/specs/extensions.md:304**: The Phase 1 test list doesn't mention checksum verification, even though install verifies `checksums.txt` / `<asset>.sha256` (line 96) and defines `EXTENSION_CHECKSUM_MISMATCH` (line 275). A tampered or truncated release asset is the main integrity boundary for binary installs; add a manager test asserting that a mismatched checksum refuses the install and surfaces that error.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F511&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A7c4ec470-aab7-11f1-9f22-e928bb2b2d5a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A7c4ec470-aab7-11f1-9f22-e928bb2b2d5a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. Added TD_NODE (process.execPath) to the env table, defined TD_PATH as the entry script, and rewrote both examples to use the portable spawn(TD_NODE, [TD_PATH, ...]) form with a note that the bare $TD_PATH shortcut is POSIX-only.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. With --pin <ref> the install queries releases/tags/{ref} instead of releases/latest; a pinned tag with no release falls through to a git checkout of the ref.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. Documented as an explicit exception: a user-set TODOIST_API_TOKEN is inherited like any other variable, since scrubbing it would break nested td calls for env-token users and they have already chosen to expose it to every process they start. The rule is that td does not inject secrets, not that it scrubs the environment.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. The dispatch section now says both getRequestedUserRef and stripUserFlag must only consider arguments before the extension name, with the explicit case that td goals --user alice reaches the extension untouched and does not set TD_USER. Also added --user scoping to the phase 1 test list.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. The trust warning is now step 3, before the release query, clone, or npm ci. Lifecycle scripts stay allowed, and the spec now says so and why (native deps; the trust decision is made before anything runs).

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. Binary installs now fetch td-extension.json from the repo at the release tag (optional, 404 is fine) and persist description and requires into manifest.json; the manifest example and the td-extension.json notes were updated to match.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. The lazy-read policy now includes reading the selected extension's manifest at dispatch, for the requires.td warning.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. Dropped the pre-clone repo and root-file checks. Script installs clone into a staging directory, validate the td-<name> executable locally, and move into place; a clone failure surfaces git's own error.

docs/specs/extensions.md
scottlovegrove: Done in 2b531fb. upgrade --all now uses a worker pool of four under the single spinner, with the rate-limit reasoning stated.

docs/specs/extensions.md
```diff
@@ -0,0 +1,356 @@
...
+standup     example/td-standup           a1b2c3d4  git (pinned)
+scratch     ~/code/td-scratch            —         local
+weekly      example/td-weekly            0.1.0     git · shadowed by built-in "weekly"
+```
+
+No network access. `--json` returns the manifest fields plus `path`, `kind`, `shadowed`, and `executable` (false when the file is missing or not executable, which is the usual state of a freshly cloned Go extension that has not been built).
+
+#### `upgrade`
+
+- Git extensions: `git pull --ff-only`; with `--force`, `git fetch` and `git reset --hard origin/HEAD`. Re-run the `npm ci` step if `package.json` changed.
```
gnapse: Should be when either `package.json` or `package-lock.json` change.

docs/specs/extensions.md
```diff
@@ -0,0 +1,356 @@
...
+└─ td-goals.json            # { pinned, checkedForUpdateAt, latestRelease }
+```
+
+On Windows the data root is `%LOCALAPPDATA%\todoist-cli`, and a local install writes a plain text file containing the target path instead of a symlink, as `gh` does, because symlinks need elevated rights there.
+
+Kind is inferred from the directory: `manifest.json` means binary; a `.git` directory means git; a symlink or path file means local. No central registry file, so a broken or hand-deleted extension cannot corrupt the others, and `rm -rf` of a directory is a valid uninstall.
+
+The config file is untouched. Nothing in this design needs a new config key, so `validateConfigForDoctor` and `KNOWN_CONFIG_KEYS` are unaffected.
+
+### `manifest.json` (written by `td`, binary extensions only)
```
gnapse: What if an extension package has a file named like this?

docs/specs/extensions.md
```diff
@@ -0,0 +1,356 @@
...
+
+- Git extensions: `git pull --ff-only`; with `--force`, `git fetch` and `git reset --hard origin/HEAD`. Re-run the `npm ci` step if `package.json` changed.
+- Binary extensions: repeat the release lookup and download when the tag differs.
+- Pinned extensions are skipped with a note unless `--force`. Local extensions are always skipped.
+- `--all` checks every extension, fetching release metadata through a small worker pool (four at a time) under a single spinner. Bounding it avoids a request burst against GitHub's secondary rate limits when many extensions are installed.
+- `--dry-run` reports what would change and exits 0.
+
+#### `remove`
+
+Deletes the extension directory and its state file. Accepts `name`, `td-name`, or `owner/td-name`. Requires no confirmation; the operation is reversible by reinstalling and the repo convention is `--yes` only for data-destroying commands.
```
gnapse: I would avoid promising it is reversible. Or add the --yes just in case. It may be an edge case, but what if an extension is local-only (e.g., generated for testing or development purposes first before publishing it)?

docs/specs/extensions.md
```diff
@@ -0,0 +1,356 @@
...
+
+### Registration
+
+Each discovered extension whose name does not collide with a built-in command is registered as a Commander command with:
+
+- `.description()` from its manifest, or `Extension <name>`.
+- `.allowUnknownOption()`, `.passThroughOptions()`, and `.helpOption(false)`, so Commander touches nothing after the name.
+- An action that spawns the executable.
+
+Registering real commands, rather than catching the unknown-command error, is what `gh` does and it is the right call here too: it makes `--help` and name completion work with no special cases, and it keeps the existing dispatcher's "find the command token in argv" logic intact. A colliding name is skipped and surfaces as `shadowed` in `list`.
```
gnapse: The intended argument handling makes sense. One implementation detail: our [current lazy-loader lookup](https://github.com/Doist/todoist-cli/blob/06e9d1257f1f6b70cef4d6f2b5c010d694388643/src/index.ts#L268-L279) scans for a built-in name anywhere in `argv`.

https://github.com/Doist/todoist-cli/blob/06e9d1257f1f6b70cef4d6f2b5c010d694388643/src/index.ts#L268-L279

Will the extension dispatch bypass that lookup? Otherwise, `td goals task list` would load the built-in `task` module and start its spinner. We should clarify that boundary here.

gnapse (approved):

docs/specs/extensions.md
scottlovegrove: Done in 7e669f7 — now re-runs npm ci when either package.json or package-lock.json changed.

docs/specs/extensions.md
scottlovegrove: Done in 7e669f7 — good point, a cloned repo can easily ship its own manifest.json. Renamed the td-written file to .td-manifest.json and reordered kind inference so local (symlink/path file) and git (.git dir) are checked before it; a clone can never be misread as binary whatever it contains.

docs/specs/extensions.md
scottlovegrove: Done in 7e669f7 — dropped the reversibility claim and spelled out per-kind behaviour: local installs delete only the symlink/path file, never the target directory; git clones with uncommitted changes refuse with EXTENSION_DIRTY unless --force; binary dirs are removed. Kept it without --yes since nothing user-authored is deleted without --force, matching both the repo convention and gh.

docs/specs/extensions.md
scottlovegrove: Done in 7e669f7 — you're right, the current find-anywhere lookup would load task and start the spinner for td goals task list. Added a 'Command token lookup' section: walk argv skipping global flags and their values, the first remaining token is the command, nothing after it is inspected; a built-in loads its module, an extension loads nothing and starts no spinner. Same boundary applied to the --user handling and the completion-server shortcut, and removed the earlier claim that the existing lookup stays intact.

rmartins90:
✅

PR #513, opened 2026-09-08
feat(filter): read a filter query in the account language

craigcarlyle:
Todoist reads a saved filter in whatever language your account is set to. `td filter view` matched English keywords, so a Spanish account with `hoy | vencidas` got priority-first where Todoist gives date-first.

- **What replaced the regex:** `isDateDrivenQuery` from the SDK, which shares its vocabulary with the web, Android and iOS lexers. The pattern and the helper that stripped project and label names ahead of it are both gone.
- **Where the language comes from:** `getUser`, which the timezone lookup already calls. The two share one request now, so this adds no round trip.
- **Requires:** `@doist/todoist-sdk` 15.2.0.
- **The test to look at:** `sin fecha`, a keyword under `es` and an unknown phrase under `en`, so each language gives it a different hierarchy. Delete `{ lang }` from the call and that test fails on its own.
- **Checked in the real CLI:** ran `td filter view` against my own account on an English query, still date-first.

Raised in Doist/todoist-sdk-typescript#669.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@doist/cli-core": "1.4.0",
-        "@doist/todoist-sdk": "15.1.0",
+        "@doist/todoist-sdk": "15.2.0",
         "@napi-rs/keyring": "2.0.0",
         "@pnpm/tabtab": "0.5.4",
         "chalk": "6.0.0",
@@ -228,9 +228,9 @@
       "license": "MIT"
     },
     "node_modules/@doist/todoist-sdk": {
-      "version": "15.1.0",
-      "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-15.1.0.tgz",
-      "integrity": "sha512-qKtCpzCyf5e/2u6Xqz1c7dukzOInPsA6nz46eka7N5til9Lnl1KnmSo8882D86zHC7nqmUd89eqtHQzBE+L6vg==",
+      "version": "15.2.0",
+      "resolved": "https://registry.npmjs.org/@doist/todoist-sdk/-/todoist-sdk-15.2.0.tgz",
+      "integrity": "sha512-DqEiAE9K+xukh8gl0nYrS+fg1Pub7fSF1B2MNY7pjsYS4dLitTwFcEo4eLWLX3Jv0yz9s22JAhYsfyjL59835g==",
       "license": "MIT",
       "dependencies": {
         "@doist/sdk-kmp": "0.2.3",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   ],
   "dependencies": {
     "@doist/cli-core": "1.4.0",
-    "@doist/todoist-sdk": "15.1.0",
+    "@doist/todoist-sdk": "15.2.0",
     "@napi-rs/keyring": "2.0.0",
     "@pnpm/tabtab": "0.5.4",
     "chalk": "6.0.0",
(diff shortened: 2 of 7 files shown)


doistbot (commented):
This PR moves date-driven filter classification from a hand-rolled English-only regex to the SDK's `isDateDrivenQuery`, now fed the account language off the existing `getUser` lookup via a shared memoized helper — no extra round trip, and `td filter view` ordering now matches what the web, Android, and iOS lexers produce. The SDK swap, the deduplicated `getAccountUser` caching (including rejected-promise cache clearing), and the well-chosen `sin fecha` test that genuinely fails without `{ lang }` all look solid. No inline issues were flagged; the only nit worth noting is a now-stale entry in the repo's doc catalog for the removed helper.

<details>

- **<a href="#">P3</a> CODEBASE.md:176**: The `src/lib/` catalog still lists `queryUsesDates` under `task-sort.ts`, which this PR removes. A reader checking the "don't reimplement" catalog would look for a function that no longer exists. Follow-up: drop the mention (the classification now lives in the SDK's `isDateDrivenQuery`).
- **<a href="#">P3</a> src/lib/api/core.ts:369**: `getAccountUser` promises one `getUser` per process, but `getCurrentUserId` (line 343) still calls `api.getUser()` directly and keeps its own cache. Route `getCurrentUserId` through `getAccountUser` so the shared single-flight memo actually covers every field-only caller.
- **<a href="#">P3</a> src/lib/api/core.ts:413**: `getAccountLanguage` duplicates the `user.lang` fetch already in `resolveLocale` (`src/commands/billing/format.ts`). Reuse this cached getter there — `(await getAccountLanguage())?.replace(/_/g, '-')` — instead of paying a second, uncached `getUser` round trip.
- **<a href="#">P3</a> src/lib/api/core.ts:404**: Remove `accountLanguageCache`. A successful `accountUserPromise` already memoizes the complete user record for the process, so `getAccountLanguage` can derive `lang` directly from it. This duplicate cache adds invalidation state without avoiding requests; `null` values are not cached by the truthiness check anyway.
- **<a href="#">P3</a> src/commands/filter/filter.test.ts:1326**: The new tests only cover the `sorting === true` path. If the `sorting ? getAccountLanguage() : Promise.resolve(undefined)` guard in `view.ts` regresses to an unconditional lookup, no test fails. Extend the existing 'does not look up the timezone when nothing is sorted' test with `expect(mockAccountLanguage).not.toHaveBeenCalled()` (or add a sibling test) so the no-sort path keeps its no-extra-`getUser` contract.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F513&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A63ed7520-abb6-11f1-87d6-287c1e9266c2%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A63ed7520-abb6-11f1-87d6-287c1e9266c2&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

henningmu (commented):
✅

PR #517, opened 2026-09-10
docs: record decisions on extensions spec open questions

scottlovegrove:
## Summary

Resolves three of the five open questions in `docs/specs/extensions.md` and moves them into a new **Decisions** section:

- **Build here first, extract to cli-core later.** The manager goes in `src/lib/extensions/` with host-specific values injected through its options object and no imports from the rest of `td`, so the eventual move to `@doist/cli-core` is a file move. Phase 1 and phase 2 updated to match.
- **First-party marker.** Extensions whose install source is owned by the `Doist` GitHub organisation show `✓ Todoist` in `list` (text `official` in `--accessible` mode, `official: true` in `--json`). The owner is one constant, to be updated when the org is renamed. The trust warning still prints for them.
- **Windows is best effort in phase 1.** All Windows paths stay specified and implemented, but the first release does not wait on a full Windows pass; `doctor` labels extensions experimental on Windows until phase 2 completes it.

Telemetry and `extension` vs `plugin` naming remain open.

### DIFF
--- a/docs/specs/extensions.md
+++ b/docs/specs/extensions.md
@@ -115,14 +115,14 @@ Review the source before use.
 #### `list`
 
 ```
-NAME        SOURCE                       VERSION   KIND
-goals       Doist/td-goals               v0.3.0    binary   ✓ Todoist
+NAME        SOURCE                       VERSION   KIND                                  OFFICIAL
+goals       Doist/td-goals               v0.3.0    binary                                ✓ Todoist
 standup     example/td-standup           a1b2c3d4  git (pinned)
 scratch     ~/code/td-scratch            —         local
 weekly      example/td-weekly            0.1.0     git · shadowed by built-in "weekly"
 ```
 
-Extensions published by Todoist carry a `✓ Todoist` marker, in `--accessible` mode rendered as the literal text `official`. The check is the install source's owner: the `Doist` GitHub organisation until it is renamed, held in one constant so the rename is a one-line change. It is an ownership label, not a signature, and it does not suppress the trust warning.
+Extensions published by Todoist show `✓ Todoist` in the `OFFICIAL` column (`Todoist` without the tick in `--accessible` mode). An extension is official when its install source is the `Doist` organisation on `github.com`: both host and owner must match, taken from `.td-manifest.json` for binary installs and from the parsed `remote.origin.url` for git installs. Owner alone is not enough, because `install` accepts any git host and `https://git.example/Doist/td-x` must not pass. Local installs are never official. The host and owner pair is one constant so the organisation rename is a one-line change. It is an ownership label, not a signature, and it does not suppress the trust warning.
 
 No network access. `--json` returns the manifest fields plus `path`, `kind`, `shadowed`, `official`, and `executable` (false when the file is missing or not executable, which is the usual state of a freshly cloned Go extension that has not been built).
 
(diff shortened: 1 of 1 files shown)


docs/specs/extensions.md
```diff
@@ -325,10 +327,11 @@ An unknown command that is not an extension keeps Commander's current message, w
...
-- cli-core: `registerExtensionCommands(program, manager)` adding `extension` / `ext` and the per-extension pass-through commands.
-- todoist-cli: wire it in `src/index.ts`; add `TD_USER` env support to the user resolver; adjust `--user` stripping; add the unknown-command hint; doctor checks; `SKILL_CONTENT` entries for `td extension …`; `CODEBASE.md` registration-pattern update.
-- Trust warning on install/upgrade. Checksum verification when a checksums asset exists.
+- `src/lib/extensions/`: `createExtensionManager({ binName, dataDir, stateDir, envPrefix, reservedNames, officialOwners })` returning discover, install (GitHub git/binary, local), list, upgrade, remove, dispatch. Nothing in it may import from `src/commands/` or read `td`-specific config directly: every host-specific value comes in through that options object, and its only dependencies are Node built-ins and what `@doist/cli-core` already exports. That is what makes the later move to cli-core a file move rather than a rewrite.
+- `src/commands/extension/`: `registerExtensionCommands(program, manager)` adding `extension` / `ext` and the per-extension pass-through commands, following the usual group-command layout.
```
doistbot: <a href="#">P2</a>  `registerExtensionCommands` is now host-side in `src/commands/extension/`, but phase 2 only extracts `src/lib/extensions/` to `@doist/cli-core`. The summary still says `tdc`/`tda` can adopt the feature 'with a one-line registration', which previously came from `registerExtensionCommands` living in cli-core. As written, adopting the extracted manager alone doesn't provide the registration helper. Either move `registerExtensionCommands` into `src/lib/extensions/` so it is extracted with the manager, or update the summary/phase 2 to state that the registration glue stays per host.

docs/specs/extensions.md
```diff
@@ -116,13 +116,15 @@ Review the source before use.
...
 NAME        SOURCE                       VERSION   KIND
-goals       Doist/td-goals               v0.3.0    binary
+goals       Doist/td-goals               v0.3.0    binary   ✓ Todoist
 standup     example/td-standup           a1b2c3d4  git (pinned)
 scratch     ~/code/td-scratch            —         local
 weekly      example/td-weekly            0.1.0     git · shadowed by built-in "weekly"
 ```
 
-No network access. `--json` returns the manifest fields plus `path`, `kind`, `shadowed`, and `executable` (false when the file is missing or not executable, which is the usual state of a freshly cloned Go extension that has not been built).
+Extensions published by Todoist carry a `✓ Todoist` marker, in `--accessible` mode rendered as the literal text `official`. The check is the install source's owner: the `Doist` GitHub organisation until it is renamed, held in one constant so the rename is a one-line change. It is an ownership label, not a signature, and it does not suppress the trust warning.
```
doistbot: <a href="#">P2</a>  Make the `official` check explicitly require the canonical `github.com` host as well as the `Doist` owner (ideally using the GitHub API-resolved repository identity). `install` accepts arbitrary git hosts, so an attacker-controlled remote such as `https://git.example/Doist/td-x` can use the same owner path and otherwise be displayed as `✓ Todoist` / `official: true`.

doistbot (commented):
This PR resolves three of the five open questions in the extensions spec — local-first manager with later cli-core extraction, the first-party `✓ Todoist` marker, and Windows as best-effort for phase 1 — and moves them into a new Decisions section.

Few things worth tightening:
- The summary still says other CLIs can adopt the feature with a one-line registration, but `registerExtensionCommands` now lives host-side in `src/commands/extension/` while phase 2 only extracts `src/lib/extensions/`. Either move the registration helper into `src/lib/extensions/` or state that the glue stays per host.
- Bind the `official` check to the canonical `github.com` host (ideally the API-resolved repo identity), not just the `Doist` owner path — `install` accepts arbitrary git hosts, so e.g. `https://git.example/Doist/td-x` would otherwise show as `✓ Todoist` / `official: true`.

<details>

- **<a href="#">P3</a> docs/specs/extensions.md:119**: The sample `list` output now has a trailing `✓ Todoist` value on the `goals` row, but the header still lists only `NAME SOURCE VERSION KIND`. Give the marker its own header column or make clear it's a badge appended to `KIND`, so the column layout is self-consistent.
- **<a href="#">P3</a> docs/specs/extensions.md:330**: `officialOwners` is plural, but the `list` section and Decision 2 both describe the owner as a single constant. Align the naming: use `officialOwner`/`officialOwnerName` if there is always one owner, or reword the decision prose if the manager is meant to accept several owners.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F517&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A45d74900-ad03-11f1-96d6-9fad38f57e7b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A45d74900-ad03-11f1-96d6-9fad38f57e7b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

docs/specs/extensions.md
scottlovegrove: Done in c15a50b. registerExtensionCommands now lives in src/lib/extensions/commands.ts under the same no-host-imports rule, so it is extracted with the manager and the one-call adoption claim holds; src/commands/extension/index.ts is a thin entry point that only calls it. Decision 1 updated to match.

docs/specs/extensions.md
scottlovegrove: Done in c15a50b. The official check now requires both host github.com and owner Doist, read from .td-manifest.json for binary installs and the parsed remote.origin.url for git installs; local installs are never official. The manager option is now officialSource: { host, owner } (singular, replacing officialOwners), and the git.example/Doist/td-x case is called out explicitly as not passing.

gnapse (commented):
✅

PR #520, opened 2026-09-10
feat(extensions): add the extension manager library

scottlovegrove:
## Summary

First half of phase 1 of [the extensions spec](https://github.com/Doist/todoist-cli/blob/main/docs/specs/extensions.md): the library that discovers, installs, upgrades, removes and runs extensions. There is no user-facing command yet, so this changes nothing for users; the `td extension` group and the `src/index.ts` wiring follow in the next PR.

Everything lives in `src/lib/extensions/` and takes the host CLI's name, directories, version, reserved command names and first-party source through one options object, with no imports from the rest of the repo. That is decision 1 in the spec, and it is what makes the eventual move to `@doist/cli-core` a file move rather than a rewrite.

## What is here

- **Discovery** is one `readdir` plus a couple of small file reads, with no subprocesses and no network, because it runs on every invocation of `td` including the common case of no extensions at all. Kind is inferred from the directory in the spec's order: symlink or path file means local, `.git` means git, `.td-manifest.json` means binary.
- **Install** covers release binaries with checksum verification, git clones with npm dependency installation, and symlinks for local development. Everything is assembled in a staging directory and moved into place only once complete, so a failed install never leaves a half-written extension behind and never destroys the one already installed. The trust warning goes out before anything is downloaded, cloned or executed.
- **Upgrade** fast-forwards clones, re-downloads a binary when the release tag moves, honours pins unless forced, leaves local installs alone, and bounds GitHub requests with a worker pool of four.
- **Remove** deletes only the link for a local install, and refuses a clone with uncommitted work unless forced.
- **Dispatch** passes arguments through verbatim, sets the documented environment contract, runs node-shebang scripts under the host's own Node, and returns the extension's exit code. The API token is deliberately absent from the child environment.

## Testing

112 tests over nine files, using real files in temporary directories rather than a mocked filesystem, a stubbed GitHub API, and a fixture extension in `src/test-support/` that reports the arguments and environment it was given. The spec's named cases are covered: a mismatched checksum refuses the install and leaves nothing behind, `--pin` resolves the tagged release rather than the latest one, a reserved name fails before any request is made, and the trust warning precedes the first fetch.

## Not in this PR

The command group, root-command wiring, the command-token lookup change, `--user` scoping, doctor checks, and `SKILL_CONTENT`. Those are the user-facing half and land next.

### DIFF
--- a/src/lib/extensions/dispatch.ts
+++ b/src/lib/extensions/dispatch.ts
@@ -1,0 +1,178 @@
+/**
+ * Running an extension.
+ *
+ * The contract is the process boundary: arguments are passed through exactly
+ * as the user typed them, the three standard streams are the host's own, and
+ * the host exits with whatever the extension exited with. Nothing is parsed,
+ * rewritten or added on either side.
+ */
+
+import { spawn } from 'node:child_process'
+import { open, stat } from 'node:fs/promises'
+import { constants } from 'node:os'
+import { CliError } from '@doist/cli-core'
+import type { DispatchOptions, Extension } from './types.js'
+
+export type SpawnPlan = { command: string; args: string[] }
+
+/** First line of a file, when it is a shebang. */
+async function readShebang(path: string): Promise<string | undefined> {
+    let handle: Awaited<ReturnType<typeof open>>
+    try {
+        handle = await open(path, 'r')
+    } catch {
+        return undefined
+    }
+    try {
+        const buffer = Buffer.alloc(256)
+        const { bytesRead } = await handle.read(buffer, 0, buffer.length, 0)
+        const head = buffer.subarray(0, bytesRead).toString('utf8')
+        if (!head.startsWith('#!')) return undefined
+        return head.split('\n', 1)[0].trim()
+    } finally {
+        await handle.close()
+    }
+}
+
+async function isExecutable(path: string): Promise<boolean> {
+    try {
+        const stats = await stat(path)
+        if (!stats.isFile()) return false
+        // Windows has no executable bit; existence is the only signal there.
+        return process.platform === 'win32' ? true : (stats.mode & 0o111) !== 0
+    } catch {
+        return false
+    }
+}
+
+/**
+ * Decide how to launch an extension's executable.
+ *
+ * Node scripts are launched with the host's own Node rather than through the
+ * shebang, which makes them work on Windows and pins them to the Node version
+ * the host already requires. Everything else runs directly on POSIX, and
+ * through `sh` on Windows, where shebangs mean nothing.
+ */
+export async function buildSpawnPlan(executablePath: string, args: string[]): Promise<SpawnPlan> {
+    const onWindows = process.platform === 'win32'
+
+    if (!(await isExecutable(executablePath))) {
+        throw new CliError(
+            'EXTENSION_NOT_EXECUTABLE',
+            `The executable for this extension is missing or not executable: ${executablePath}`,
+            {
+                hints: [
+                    'Compiled extensions need building after a source install.',
+                    'Script extensions need an executable bit: chmod +x the file.',
+                ],
+            },
+        )
+    }
+
+    if (onWindows && /\.(exe|cmd|bat)$/i.test(executablePath)) {
+        return { command: executablePath, args }
+    }
+
+    const shebang = await readShebang(executablePath)
+    if (shebang && /\bnode\b/.test(shebang)) {
+        return { command: process.execPath, args: [executablePath, ...args] }
+    }
+
+    if (onWindows) {
+        // Matches how `gh` runs script extensions on Windows: Git for Windows
+        // provides the interpreter, and without it a shebang script cannot run.
+        return {
+            command: 'sh',
+            args: ['-c', '"$0" "$@"', executablePath, ...args],
+        }
+    }
+
+    return { command: executablePath, args }
+}
+
+export type ExtensionEnvOptions = {
+    envPrefix: string
+    version: string
+    configDir: string
+    hostPath: string
+    extension: Extension
+    user?: string
+    accessible?: boolean
+    extra?: Record<string, string | undefined>
+}
+
+/**
+ * The variables an extension can rely on. The API token is deliberately not
+ * among them: an extension that needs it runs `<bin> auth token view`, which
+ * keeps the secret out of the environment of every process it starts.
+ */
+export function buildExtensionEnv(options: ExtensionEnvOptions): NodeJS.ProcessEnv {
+    const { envPrefix: prefix, extension } = options
+    const env: NodeJS.ProcessEnv = {
+        ...process.env,
+        [`${prefix}_EXTENSION`]: '1',
+        [`${prefix}_EXTENSION_NAME`]: extension.name,
+        [`${prefix}_EXTENSION_DIR`]: extension.dir,
+        [`${prefix}_PATH`]: options.hostPath,
+        [`${prefix}_NODE`]: process.execPath,
+        [`${prefix}_VERSION`]: options.version,
+        [`${prefix}_CONFIG_DIR`]: options.configDir,
+    }
+
+    if (options.user) env[`${prefix}_USER`] = options.user
+    if (options.accessible) env[`${prefix}_ACCESSIBLE`] = '1'
+
+    for (const [key, value] of Object.entries(options.extra ?? {})) {
+        if (value === undefined) delete env[key]
+        else env[key] = value
+    }
+
+    return env
+}
+
+/** Exit code convention for a child killed by a signal, as the shell reports it. */
+function exitCodeForSignal(signal: NodeJS.Signals): number {
+    const number = (constants.signals as Record<string, number | undefined>)[signal]
+    return number === undefined ? 1 : 128 + number
+}
+
+/** Run an extension to completion and return the code the host should exit with. */
+export async function dispatchExtension(
+    extension: Extension,
+    args: string[],
+    env: NodeJS.ProcessEnv,
+    _options: DispatchOptions = {},
+): Promise<number> {
+    const plan = await buildSpawnPlan(extension.executablePath, args)
+
+    return new Promise((resolve, reject) => {
+        const child = spawn(plan.command, plan.args, {
+            cwd: process.cwd(),
+            env,
+            stdio: 'inherit',
+        })
+
+        child.on('error', (error: NodeJS.ErrnoException) => {
+            if (error.code === 'ENOENT' && plan.command === 'sh') {
+                reject(
+                    new CliError(
+                        'EXTENSION_NEEDS_SHELL',
+                        `Running "${extension.name}" needs a POSIX shell, which was not found.`,
+                        { hints: ['Install Git for Windows, which provides sh.exe.'] },
+                    ),
+                )
+                return
+            }
+            reject(
+                new CliError(
+                    'EXTENSION_NOT_EXECUTABLE',
+                    `Could not run "${extension.name}": ${error.message}`,
+                ),
+            )
+        })
+
+        child.on('close', (code, signal) => {
+            resolve(signal ? exitCodeForSignal(signal) : (code ?? 0))
+        })
+    })
+}
--- a/src/lib/extensions/git.ts
+++ b/src/lib/extensions/git.ts
@@ -1,0 +1,104 @@
+/**
+ * The git operations the extension system needs. Every one of them shells out
+ * to the user's own git, so clones inherit whatever credential helper and SSH
+ * configuration they already have.
+ */
+
+import { CliError } from '@doist/cli-core'
+import { outputHints, run, type RunResult } from './run.js'
+
+function requireGit(result: RunResult): void {
+    if (result.missing) {
+        throw new CliError('EXTENSION_INSTALL_FAILED', 'git was not found on PATH.', {
+            hints: ['Install git, or install this extension from a release binary instead.'],
+        })
+    }
+}
+
+export async function clone(url: string, destination: string): Promise<void> {
+    const result = await run('git', ['clone', '--quiet', url, destination])
+    requireGit(result)
+    if (result.code !== 0) {
+        throw new CliError('EXTENSION_NOT_INSTALLABLE', `Could not clone ${url}.`, {
+            hints: outputHints(result),
+        })
+    }
+}
+
+export async function checkout(dir: string, ref: string): Promise<void> {
+    const result = await run('git', ['checkout', '--quiet', ref], { cwd: dir })
+    requireGit(result)
+    if (result.code !== 0) {
+        throw new CliError('EXTENSION_NOT_INSTALLABLE', `Could not check out "${ref}".`, {
+            hints: outputHints(result),
+        })
+    }
+}
+
+/** Full commit SHA of HEAD, or undefined when the directory is not a clone. */
+export async function headSha(dir: string): Promise<string | undefined> {
+    const result = await run('git', ['rev-parse', 'HEAD'], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim() || undefined
+}
+
+/** True when the working tree has changes the user would not want deleted. */
+export async function isDirty(dir: string): Promise<boolean> {
+    const result = await run('git', ['status', '--porcelain'], { cwd: dir })
+    if (result.code !== 0) return false
+    return result.stdout.trim().length > 0
+}
+
+export type PullResult = { changed: boolean; from?: string; to?: string; error?: string }
+
+/** Fast-forward the clone. Never rewrites local work; `resetToRemote` does that. */
+export async function pull(dir: string): Promise<PullResult> {
+    const from = await headSha(dir)
+    const result = await run('git', ['pull', '--quiet', '--ff-only'], { cwd: dir })
+    requireGit(result)
+    if (result.code !== 0) {
+        return { changed: false, from, error: outputHints(result).join(' ') }
+    }
+    const to = await headSha(dir)
+    return { changed: from !== to, from, to }
+}
+
+/** Discard local commits and working-tree changes in favour of the remote. */
+export async function resetToRemote(dir: string): Promise<PullResult> {
+    const from = await headSha(dir)
+    const fetched = await run('git', ['fetch', '--quiet', 'origin'], { cwd: dir })
+    requireGit(fetched)
+    if (fetched.code !== 0) {
+        return { changed: false, from, error: outputHints(fetched).join(' ') }
+    }
+    const reset = await run('git', ['reset', '--quiet', '--hard', 'origin/HEAD'], { cwd: dir })
+    if (reset.code !== 0) {
+        return { changed: false, from, error: outputHints(reset).join(' ') }
+    }
+    const to = await headSha(dir)
+    return { changed: from !== to, from, to }
+}
+
+/** Commit the remote's default branch points at, without fetching it. */
+export async function remoteHeadSha(dir: string): Promise<string | undefined> {
+    const result = await run('git', ['ls-remote', 'origin', 'HEAD'], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim().split(/\s+/)[0] || undefined
+}
+
+/** Resolve a ref to the SHA it names, used when pinning a clone. */
+export async function resolveRef(dir: string, ref: string): Promise<string | undefined> {
+    const result = await run('git', ['rev-parse', ref], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim() || undefined
+}
+
+/** Files that changed between two commits, used to decide if deps need reinstalling. */
+export async function changedFiles(dir: string, from: string, to: string): Promise<string[]> {
+    const result = await run('git', ['diff', '--name-only', `${from}..${to}`], { cwd: dir })
+    if (result.code !== 0) return []
+    return result.stdout
+        .split('\n')
+        .map((line) => line.trim())
+        .filter(Boolean)
+}
--- a/src/lib/extensions/install.ts
+++ b/src/lib/extensions/install.ts
@@ -1,0 +1,343 @@
+/**
+ * Installing an extension.
+ *
+ * Three shapes are supported: a prebuilt release binary, a git clone, and a
+ * symlink to a directory the user is working in. Everything is assembled in a
+ * staging directory inside the extensions directory and moved into place only
+ * once it is complete, so a failed install never leaves a half-written
+ * extension behind and never destroys the one already installed.
+ */
+
+import { chmod, mkdir, readdir, rename, rm, stat, symlink, writeFile } from 'node:fs/promises'
+import { homedir } from 'node:os'
+import { basename, join, resolve } from 'node:path'
+import { CliError } from '@doist/cli-core'
+import { discoverExtensions } from './discover.js'
+import { checkout, clone, headSha, resolveRef } from './git.js'
+import {
+    assetSuffixes,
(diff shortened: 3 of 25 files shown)


doistbot (commented):
Adds the self-contained extension manager library under `src/lib/extensions/` — discovery, staged installs, upgrades, removal, and dispatch — with no repo imports, ahead of the user-facing `td extension` commands.

Few things worth tightening:
- Secrets leak across the process boundary: `buildExtensionEnv` passes `TODOIST_API_TOKEN`/`GH_TOKEN` through to extensions despite the documented contract, and npm lifecycle scripts during dependency installs inherit the full host environment — build from a minimal allowlist and add a test asserting the token is absent.
- Upgrade flow gaps: upgrades never print the trust warning the spec requires, a forced git upgrade resets the clone but leaves the stale pin in state so later plain upgrades still skip it (worth asserting in the test), and `upgradeGit` itself has no coverage.
- Removal can destroy uncommitted work: `isDirty` treats git errors as "clean" and the no-lockfile npm fallback generates a `package-lock.json` that dirties the clone, so `remove` wrongly refuses or wrongly deletes — distinguish the error case and avoid writing the lockfile.
- Smaller items: Windows spawning of `npm.cmd` without a shell fails with EINVAL under Node's 2024 security fix, child output capture is unbounded, non-directory local install targets are accepted but can never dispatch, and `^0.0.x` ranges are treated as minor-bounded rather than patch-only.

<details>
<summary>Optional follow-up notes (15)</summary>

- **<a href="#">P3</a> src/lib/extensions/install.ts:122**: `moveIntoPlace` deletes the existing install (and the destination) before renaming the staged directory in. If the process dies between the `rm` and the `rename`, or the rename fails — e.g. on Windows a transient file lock/EPERM on the staged directory is realistic — the previously working extension is gone, contradicting the module doc's "never destroys the one already installed" guarantee. Consider renaming the old directory aside to a temp path first, then renaming staged into place, then deleting the old one, so a failure leaves the previous install recoverable.
- **<a href="#">P3</a> src/lib/extensions/types.ts:157**: `withSpinner` is declared on `ExtensionManagerOptions` ("Wraps a slow operation in the host's spinner"), but `createExtensionManager` never reads it, so the option silently does nothing. Either wire it into install/upgrade/dispatch or drop it from the options until a caller uses it.
- **<a href="#">P3</a> src/lib/extensions/discover.ts:25**: `exists` is defined identically in three new modules: here, `install.ts` (line 48), and `npm.ts` (line 14). Extract one shared helper (for example, a small fs utility beside `run.ts`) and import it everywhere so the `stat`/catch semantics have a single source of truth.
- **<a href="#">P3</a> src/lib/extensions/manager.ts:131**: This `isExecutable` duplicates the private `isExecutable` in `dispatch.ts` (line 37) almost verbatim, differing only by a comment. Export that function from `dispatch.ts` and reuse it here instead of maintaining two copies of the same platform/executable-bit check.
- **<a href="#">P3</a> src/lib/extensions/discover.ts:54**: `parseRemote` re-implements git-remote parsing that `parseSource` in `source.ts` also does: both use the same scp-like regex and `.git`-suffix stripping. Extract one shared `host`/`owner`/`repo` parser and call it from both, so the two parsing paths cannot drift.
- **<a href="#">P3</a> src/lib/extensions/dispatch.ts:144**: `_options` is never referenced in `dispatchExtension`. Drop the parameter and the `dispatchOptions` argument forwarded at `manager.ts:219` until the function actually consumes it.
- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:44**: `before = await headSha(...)` runs on every non-dry upgrade too, but that branch never reads `before`; `pull` and `resetToRemote` each compute the same starting SHA themselves. Move the call inside the `dryRun` branch to avoid an extra `git rev-parse` subprocess per git extension upgrade.
- **<a href="#">P3</a> src/lib/extensions/install.ts:328**: This `if` block has no body — it only documents that a pinned tag with no release falls through to `installGit`. An empty conditional reads like missing code. Drop the `if` and keep the comment as a plain statement before the fall-through.
- **<a href="#">P3</a> src/lib/extensions/manager.ts:191**: `upgrade()` re-runs full extension discovery for every selector. Each `requireExtension` calls `find`, which calls `discover()` — a readdir plus manifest/state reads for every installed extension — so `upgrade(['a','b','c'])` performs three complete scans in parallel. Discover once and resolve all selectors against that single snapshot instead.
- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:149**: `readState` runs for every extension, but its result is only used inside the `extension.pinned && !options.force` branch (for the detail message). Move the call into that branch so unpinned extensions skip the state-file read.
- **<a href="#">P3</a> src/lib/extensions/discover.ts:118**: Regular entries pay for both `stat` and `lstat` here on every discovery. Call `lstat` first; it already identifies regular files/directories, and only follow it with `stat` when the entry is a symlink whose target must be inspected.
- **<a href="#">P3</a> src/lib/extensions/source.test.ts:36**: `'td--x'` appears twice in this reject matrix, so the second entry is a duplicate no-op case. Drop one.
- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:89**: These negative env assertions depend on the test process having no `TD_*` variables, but `buildExtensionEnv` starts from `...process.env`. If `TD_TOKEN`, `TD_API_TOKEN`, `TD_USER`, or `TD_ACCESSIBLE` is set in the environment, the matching `toBeUndefined()` assertion fails spuriously; likewise `expect(env.TD_EXTENSION).toBe(process.env.TD_EXTENSION)` passes even when an ambient value is present. Clear the relevant `TD_*` keys in `beforeEach` (or use `vi.stubEnv`) so the contract is tested deterministically.
- **<a href="#">P3</a> src/lib/extensions/install.test.ts:319**: `if (!hasGit) return` makes the test pass vacuously when git is missing, and the skip is invisible in the test report — a CI machine without git silently loses all clone-install coverage while everything looks green. remove.test.ts (lines 105, 116, 127) has the same pattern. Vitest reports skips properly with `it.skipIf(!hasGit)` / `describe.skipIf(!hasGit)`, so use that instead of an early return.
- **<a href="#">P3</a> src/lib/extensions/manager.ts:1**: AGENTS.md requires updating CODEBASE.md when a new broadly-reusable module is added under `src/lib/` — a new 24-file subsystem directory qualifies. Add an entry to the `src/lib/` catalog in CODEBASE.md (even a short one noting it's host-agnostic and headed to `@doist/cli-core`) so the map stays accurate.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F520&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A281fc1b0-ad1e-11f1-9560-c3d312d84ad2%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A281fc1b0-ad1e-11f1-9560-c3d312d84ad2&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,172 @@
...
+        asset,
+        {},
+        context,
+        extension,
+    )
+
+    return { name: extension.name, outcome: 'upgraded', from: manifest.tag, to: release.tag }
+}
+
+export async function upgradeExtension(
```
doistbot, line 136: <a href="#">P2</a>  `upgradeExtension` never emits `context.trustWarning`, so an upgrade downloads/pulls and — for git upgrades that re-run `npm ci` — can execute repository code without the warning the spec says is printed on every upgrade. `installExtension` warns before its first fetch; add `context.warn(context.trustWarning)` before the `upgradeGit`/`upgradeBinary` dispatch here.

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,172 @@
...
+        }
+        return {
+            name: extension.name,
+            outcome: 'would-upgrade',
+            from: before.slice(0, 8),
+            to: remote.slice(0, 8),
+        }
+    }
+
+    const result = options.force ? await resetToRemote(extension.dir) : await pull(extension.dir)
```
doistbot, line 60: <a href="#">P2</a>  A forced git upgrade resets the clone to `origin/HEAD` but never clears the pin in the state file, unlike `installBinary`, which writes `pinned: undefined`. The old `pinned` value is still read on the next discovery, so a plain `upgrade` skips the extension again even after `--force`. Clear the pin with `writeState` after a successful forced `resetToRemote`.

src/lib/extensions/git.ts
```diff
@@ -0,0 +1,104 @@
...
+export async function headSha(dir: string): Promise<string | undefined> {
+    const result = await run('git', ['rev-parse', 'HEAD'], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim() || undefined
+}
+
+/** True when the working tree has changes the user would not want deleted. */
+export async function isDirty(dir: string): Promise<boolean> {
+    const result = await run('git', ['status', '--porcelain'], { cwd: dir })
+    if (result.code !== 0) return false
```
doistbot, line 48: <a href="#">P2</a>  `isDirty` returns `false` on any non-zero `git status --porcelain` exit, conflating "clean" with "could not check". `removeExtension` then deletes the clone, so uncommitted work can be removed without `--force` when git is missing from PATH, the repo hits the safe.directory/dubious-ownership error, or `.git` is corrupt. Distinguish the error case and make remove refuse (or require `--force`) when the dirty state cannot be determined.

src/lib/extensions/npm.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Install production dependencies in place. Uses `npm ci` when a lockfile is
+ * present and falls back to `npm install` when it is not.
+ */
+export async function installDependencies(dir: string): Promise<void> {
+    if (!(await hasPackageJson(dir))) return
+
+    const hasLockfile = await exists(join(dir, 'package-lock.json'))
+    const args = hasLockfile ? ['ci', '--omit=dev'] : ['install', '--omit=dev']
```
doistbot, line 37: <a href="#">P2</a>  The no-lockfile fallback runs `npm install --omit=dev`, which creates an untracked `package-lock.json` inside the clone. `git status --porcelain` then reports the working tree as dirty, so `remove` refuses with `EXTENSION_DIRTY` on an extension the user never edited. Pass `--no-package-lock` (or otherwise avoid writing into the clone) for this fallback.

src/lib/extensions/npm.ts
```diff
@@ -0,0 +1,60 @@
...
+    } catch {
+        return false
+    }
+}
+
+export async function hasPackageJson(dir: string): Promise<boolean> {
+    return exists(join(dir, 'package.json'))
+}
+
+const NPM_COMMAND = process.platform === 'win32' ? 'npm.cmd' : 'npm'
```
doistbot, line 27: <a href="#">P2</a>  On Windows this spawns `npm.cmd` via `spawn()` without a shell. Since Node's 2024 child_process security fix (CVE-2024-27980), spawning `.bat`/`.cmd` files without `shell: true` fails with EINVAL, and this repo requires Node >= 24. So `installDependencies` always fails on Windows: `missing` is false (the error code is EINVAL, not ENOENT), so the user gets `EXTENSION_INSTALL_FAILED` ("Could not install the extension's dependencies") for any git-cloned extension with a package.json. Run npm through `cmd.exe /d /s /c npm ...` (or another shell-free approach such as resolving and spawning `node <npm-prefix>/npm-cli.js`) and treat the EINVAL case distinctly so the error isn't misdiagnosed.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,343 @@
...
+
+async function installLocal(
+    path: string,
+    options: InstallOptions,
+    context: InstallContext,
+): Promise<InstallResult> {
+    const target = resolve(expandHome(path))
+    const dirName = basename(target)
+
+    if (!(await exists(target))) {
```
doistbot, line 270: <a href="#">P2</a>  This accepts any existing filesystem entry as a local extension. For example, installing a regular file named `td-x` creates a local link, but dispatch then looks for `<file>/td-x` and can never run it. Check that `target` is a directory before linking it and reject other entry types.

src/lib/extensions/dispatch.ts
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * The variables an extension can rely on. The API token is deliberately not
+ * among them: an extension that needs it runs `<bin> auth token view`, which
+ * keeps the secret out of the environment of every process it starts.
+ */
+export function buildExtensionEnv(options: ExtensionEnvOptions): NodeJS.ProcessEnv {
+    const { envPrefix: prefix, extension } = options
+    const env: NodeJS.ProcessEnv = {
+        ...process.env,
```
doistbot, line 112: <a href="#">P1</a>  `buildExtensionEnv` copies `...process.env`, so the child extension inherits `TODOIST_API_TOKEN` — the CLI's canonical API-token environment variable (see `src/lib/auth-store.ts` and `src/lib/auth.ts`) — along with `GH_TOKEN`/`GITHUB_TOKEN`. This contradicts the function's own doc comment ("The API token is deliberately not among them") and the PR summary, and the test at dispatch.test.ts only asserts `TD_TOKEN`/`TD_API_TOKEN` are absent. A malicious extension can exfiltrate the raw Todoist token simply by reading its own environment. Build the child environment from a minimal allowlist (e.g. `PATH` plus the `TD_*` contract variables) or explicitly delete `TODOIST_API_TOKEN`, `GH_TOKEN`, and `GITHUB_TOKEN` before passing it to `spawn`, and add a test asserting `TODOIST_API_TOKEN` is not present.

src/lib/extensions/version-range.ts
```diff
@@ -0,0 +1,76 @@
...
+    return { operator: operator ?? '=', version }
+}
+
+/** Upper bound (exclusive) for `^` and `~`, following npm's rules. */
+function upperBound(operator: string, version: string): string | undefined {
+    const { major, minor } = parseVersion(version)
+    if (operator === '^') {
+        // ^0.2.3 allows patch changes only; ^1.2.3 allows minor changes.
+        return major === 0 ? `0.${minor + 1}.0` : `${major + 1}.0.0`
+    }
```
doistbot, line 31: <a href="#">P2</a>  Handle `^0.0.x` as a patch-only range. This treats every major-zero caret range as bounded by the next minor, so `satisfiesRange('0.0.4', '^0.0.3')` returns true even though npm semver limits that range to `<0.0.4`. The incompatibility warning is suppressed for those extensions.

src/lib/extensions/run.ts
```diff
@@ -0,0 +1,68 @@
...
+        const child = spawn(command, args, {
+            cwd: options.cwd,
+            env: options.env ?? process.env,
+            stdio: ['ignore', 'pipe', 'pipe'],
+        })
+
+        let stdout = ''
+        let stderr = ''
+        child.stdout?.on('data', (chunk: Buffer) => {
+            stdout += chunk.toString()
```
doistbot, line 40: <a href="#">P1</a>  Bound captured child output. Both streams are appended indefinitely, while most callers only use the last few lines via `outputHints`. `npm install` intentionally runs extension lifecycle scripts, so a verbose or looping script can exhaust the CLI heap. Keep a fixed-size output tail (or add a capped-capture mode for callers that do not parse stdout) while continuing to drain the streams.

src/lib/extensions/upgrade.test.ts
```diff
@@ -0,0 +1,234 @@
...
+        const results = await mapWithConcurrency([3, 1, 2], 2, async (item) => {
+            await new Promise((resolve) => setTimeout(resolve, item))
+            return item * 10
+        })
+
+        expect(results).toEqual([30, 10, 20])
+    })
+})
+
+describe('upgradeExtension', () => {
```
doistbot, line 82: <a href="#">P2</a>  No test exercises `upgradeGit`. The PR's headline git-upgrade behavior — `git pull --ff-only`, `resetToRemote` under `--force`, and reinstalling deps when `package.json`/`package-lock.json` change — can break without any test failing. `install.test.ts` already builds local `file://` origins, so a clone-then-advance-origin test fits the existing style and would cover the fast-forward branch; adding a `package.json` change between the two commits would also cover the dependency-reinstall branch.

src/lib/extensions/upgrade.test.ts
```diff
@@ -0,0 +1,234 @@
...
+        const result = await upgradeExtension(
+            await find('goals'),
+            {},
+            contextFor(stubGitHub('v2.0.0')),
+        )
+
+        expect(result).toMatchObject({ outcome: 'skipped', detail: 'pinned to v1.0.0' })
+    })
+
+    it('upgrades past a pin only when forced', async () => {
```
doistbot, line 155: <a href="#">P2</a>  This test proves the forced upgrade ran but not that the pin was cleared. `installBinary` writes state with `pinned: undefined`; if that write regresses, the next `upgrade` still sees `extension.pinned` and skips forever. Assert that `readState(stateDir, 'td-goals')` no longer contains a `pinned` value after the forced upgrade.

src/lib/extensions/npm.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Install production dependencies in place. Uses `npm ci` when a lockfile is
+ * present and falls back to `npm install` when it is not.
+ */
+export async function installDependencies(dir: string): Promise<void> {
+    if (!(await hasPackageJson(dir))) return
+
+    const hasLockfile = await exists(join(dir, 'package-lock.json'))
+    const args = hasLockfile ? ['ci', '--omit=dev'] : ['install', '--omit=dev']
+    const result = await run(NPM_COMMAND, args, { cwd: dir })
```
doistbot, line 38: <a href="#">P2</a>  `installDependencies` calls `run()` without an `env`, so `run.ts` passes `process.env` through unchanged. npm executes the extension's own `preinstall`/`install`/`postinstall` lifecycle scripts, which therefore receive every secret in the host environment (`TODOIST_API_TOKEN`, `GH_TOKEN`, `GITHUB_TOKEN`, etc.). Install scripts need registry access, not the host CLI's API token. Pass a minimal environment, or strip token-bearing variables, for dependency installs. See https://handbook.doist.com/doc/standard-secrets-management-wbKmIfrtgr (least privilege).

src/lib/extensions/dispatch.ts
scottlovegrove: Addressed in #523, though not by stripping. The spec records inheriting a user-exported TODOIST_API_TOKEN as a deliberate exception: the CLI does not inject secrets, but it does not scrub the environment its user chose either, and stripping it would break nested td calls for anyone who authenticates that way. What was wrong was the doc comment claiming otherwise, so that has been rewritten to describe what actually happens and why, and there is now a test pinning the pass-through so it cannot change by accident. npm lifecycle scripts are a different matter and are handled in #524.

src/lib/extensions/npm.ts
scottlovegrove: Done in #524. npm now runs with TODOIST_API_TOKEN, GH_TOKEN and GITHUB_TOKEN removed from its environment. A denylist rather than an allowlist: npm legitimately needs a large and open-ended slice of the environment for proxies, certificate bundles and registry credentials, so naming what must not travel is easier to keep correct than naming everything that may. Covered by a test that puts a fake npm on PATH and inspects the environment it received.

src/lib/extensions/upgrade.ts
scottlovegrove: Done in #526. The trust warning is now printed before an upgrade pulls or downloads, and once per run rather than once per extension, so upgrade --all does not repeat it. Dry runs stay silent because nothing is fetched or executed. Covered by a test.

src/lib/extensions/upgrade.ts
scottlovegrove: Done in #526. A forced upgrade now clears the pin it moved past. There is a test that pins a clone, confirms a plain upgrade skips it, forces one, and then confirms the state file no longer carries a pin and the next plain upgrade is no longer skipped.

src/lib/extensions/git.ts
scottlovegrove: Done in #524. isDirty is now workingTreeState returning clean, dirty or unknown, and remove refuses on unknown with its own message rather than treating it as clean. Covered by a test that makes .git unreadable and asserts the clone survives.

src/lib/extensions/npm.ts
scottlovegrove: Done in #524. The no-lockfile path now passes --no-package-lock, so dependency installation no longer leaves the clone looking modified. Asserted in the npm tests.

src/lib/extensions/npm.ts
scottlovegrove: Done in #524. npm is invoked through the command interpreter on Windows (cmd.exe /d /s /c npm) rather than spawning npm.cmd directly.

src/lib/extensions/install.ts
scottlovegrove: Done in #525. A local install now requires the target to be a directory, and rejects anything else with EXTENSION_NOT_INSTALLABLE plus a hint about the expected layout.

src/lib/extensions/version-range.ts
scottlovegrove: Done in #527. Caret now narrows as the version approaches zero: ^1.2.3 allows minor changes, ^0.2.3 allows patch changes, ^0.0.3 allows nothing above itself.

src/lib/extensions/run.ts
scottlovegrove: Done in #524. Both streams are still drained, but only the last 64KB of each is retained, which is well beyond what outputHints reads.

src/lib/extensions/upgrade.test.ts
scottlovegrove: Done in #526. There is now a git-clone upgrade suite built on a local file:// origin covering up-to-date, fast-forward, dry run, the trust warning, pin clearing, and dependency reinstallation firing only when package.json changed.

src/lib/extensions/upgrade.test.ts
scottlovegrove: Done in #526, as described on the pin thread: the forced-upgrade test now asserts the state file no longer carries a pin and that the following plain upgrade is not skipped.

scottlovegrove:
Closing in favour of a seven-part stack, following the slicing plan suggested above. The content is the same work with this review's feedback applied; each thread above says which slice carries its fix.

- #521 core types, naming and manifests
- #522 discovery and test fixtures
- #523 dispatch
- #524 git, npm and GitHub helpers, and removal
- #525 install
- #526 upgrade
- #527 the manager facade

Every slice type-checks and passes its own tests on its own branch, and the tip of the stack is byte-identical to what this branch would have been.

PR #521, opened 2026-09-10
feat(extensions): core types, naming and manifests

scottlovegrove:
Slice 1 of 7. Replaces #520, which was one 3.7k-line PR; this is the same work split along doistbot's suggested plan, with its review feedback already applied.

The vocabulary every later slice builds on: the types describing an installed extension, the naming rules derived from the host binary name (`td` gives `td-<name>` directories, `td-extension.json` and `.td-manifest.json`), one parser shared by install sources and by git remotes read back off disk, manifest reading and writing including the `package.json` fallback for Node extensions, and per-extension state kept outside the extension directory so that deleting that directory by hand is still a clean uninstall.

Nothing is wired up yet. See [the spec](https://github.com/Doist/todoist-cli/blob/main/docs/specs/extensions.md) for where this is going.

### DIFF
--- a/src/lib/extensions/fs-utils.ts
+++ b/src/lib/extensions/fs-utils.ts
@@ -1,0 +1,37 @@
+/**
+ * The two filesystem questions the extension system asks everywhere: does this
+ * path exist, and can it be run.
+ */
+
+import { stat } from 'node:fs/promises'
+
+export async function exists(path: string): Promise<boolean> {
+    try {
+        await stat(path)
+        return true
+    } catch {
+        return false
+    }
+}
+
+export async function isDirectory(path: string): Promise<boolean> {
+    try {
+        return (await stat(path)).isDirectory()
+    } catch {
+        return false
+    }
+}
+
+/**
+ * True when the path is a file the operating system will run. Windows has no
+ * executable bit, so there existence is the only signal available.
+ */
+export async function isExecutable(path: string): Promise<boolean> {
+    try {
+        const stats = await stat(path)
+        if (!stats.isFile()) return false
+        return process.platform === 'win32' ? true : (stats.mode & 0o111) !== 0
+    } catch {
+        return false
+    }
+}
--- a/src/lib/extensions/fs-utils.ts
+++ b/src/lib/extensions/fs-utils.ts
@@ -3,7 +3,8 @@
  * path exist, and can it be run.
  */
 
-import { stat } from 'node:fs/promises'
+import { constants } from 'node:fs'
+import { access, stat } from 'node:fs/promises'
 
 export async function exists(path: string): Promise<boolean> {
     try {
@@ -23,14 +24,17 @@ export async function isDirectory(path: string): Promise<boolean> {
 }
 
 /**
- * True when the path is a file the operating system will run. Windows has no
- * executable bit, so there existence is the only signal available.
+ * True when the path is a file this process can actually run.
+ *
+ * `access` is asked rather than the mode bits, because permission can come
+ * from an access-control list the mode does not describe, and on Windows it
+ * reduces to an existence check, which is the only signal available there.
  */
 export async function isExecutable(path: string): Promise<boolean> {
     try {
-        const stats = await stat(path)
-        if (!stats.isFile()) return false
-        return process.platform === 'win32' ? true : (stats.mode & 0o111) !== 0
+        if (!(await stat(path)).isFile()) return false
+        await access(path, constants.X_OK)
+        return true
     } catch {
         return false
     }
--- a/src/lib/extensions/manifest.ts
+++ b/src/lib/extensions/manifest.ts
@@ -7,30 +7,17 @@
  * dot-prefixed so it cannot collide with anything the extension itself ships.
  */
 
-import { readFile, writeFile } from 'node:fs/promises'
 import { join } from 'node:path'
+import { isRecord, readJsonFile, readJsonValue, writeJsonFile } from './json-file.js'
 import { authoredManifestFileName, manifestFileName } from './source.js'
 import type { AuthoredManifest, InstalledManifest } from './types.js'
 
-async function readJson(path: string): Promise<unknown | undefined> {
-    let raw: string
-    try {
-        raw = await readFile(path, 'utf8')
-    } catch {
-        return undefined
-    }
-    try {
-        return JSON.parse(raw)
-    } catch {
-        return undefined
-    }
-}
-
-function isRecord(value: unknown): value is Record<string, unknown> {
-    return typeof value === 'object' && value !== null && !Array.isArray(value)
-}
-
-function pickAuthoredFields(value: unknown): AuthoredManifest | undefined {
+/**
+ * Take only the fields this system understands, and only when they are the
+ * right type. A manifest is written by hand, so a wrong type is likelier than
+ * a missing field and neither should stop an extension from working.
+ */
+export function pickAuthoredFields(value: unknown): AuthoredManifest | undefined {
     if (!isRecord(value)) return undefined
     const manifest: AuthoredManifest = {}
     if (typeof value.description === 'string') manifest.description = value.description
@@ -55,23 +42,47 @@ export async function readAuthoredManifest(
     binName: string,
 ): Promise<AuthoredManifest | undefined> {
     const dedicated = pickAuthoredFields(
-        await readJson(join(dir, authoredManifestFileName(binName))),
+        await readJsonValue(join(dir, authoredManifestFileName(binName))),
     )
     if (dedicated) return dedicated
 
-    const packageJson = await readJson(join(dir, 'package.json'))
+    const packageJson = await readJsonValue(join(dir, 'package.json'))
     if (!isRecord(packageJson)) return undefined
     return pickAuthoredFields(packageJson[binName])
 }
 
+/**
+ * The manifest the CLI wrote at install time, or undefined when it is missing
+ * or does not describe an install.
+ *
+ * Every required field is checked rather than cast over, because upgrades read
+ * `tag` and `owner` back to decide what to fetch: a truncated or hand-edited
+ * file should read as "no manifest", which callers already handle, rather than
+ * as a manifest full of undefined.
+ */
 export async function readInstalledManifest(
     dir: string,
     binName: string,
 ): Promise<InstalledManifest | undefined> {
-    const value = await readJson(join(dir, manifestFileName(binName)))
+    const value = await readJsonValue(join(dir, manifestFileName(binName)))
     if (!isRecord(value)) return undefined
-    if (typeof value.name !== 'string' || typeof value.owner !== 'string') return undefined
-    return value as InstalledManifest
+
+    const strings = ['owner', 'name', 'host', 'tag', 'asset', 'installedAt'] as const
+    if (strings.some((field) => typeof value[field] !== 'string')) return undefined
+    if (typeof value.pinned !== 'boolean') return undefined
+    if (value.sha256 !== undefined && typeof value.sha256 !== 'string') return undefined
+
+    return {
+        owner: value.owner as string,
+        name: value.name as string,
+        host: value.host as string,
+        tag: value.tag as string,
+        asset: value.asset as string,
+        installedAt: value.installedAt as string,
+        pinned: value.pinned,
+        sha256: value.sha256 as string | undefined,
+        ...pickAuthoredFields(value),
+    }
 }
 
 export async function writeInstalledManifest(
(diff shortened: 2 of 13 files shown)


src/lib/extensions/manifest.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export async function readInstalledManifest(
+    dir: string,
+    binName: string,
+): Promise<InstalledManifest | undefined> {
+    const value = await readJson(join(dir, manifestFileName(binName)))
+    if (!isRecord(value)) return undefined
+    if (typeof value.name !== 'string' || typeof value.owner !== 'string') return undefined
+    return value as InstalledManifest
```
doistbot: <a href="#">P2</a>  `readInstalledManifest` checks only `name` and `owner`, then casts the raw JSON to `InstalledManifest`, whose type requires `host`, `tag`, `pinned`, `asset`, and `installedAt`. A hand-edited or truncated `.td-manifest.json` (for example `{"name":"td-goals","owner":"Doist"}`, or `tag` as a number) is accepted, so later code reading `manifest.tag` or `manifest.installedAt` gets `undefined`/wrong types under a type that says otherwise. Validate the remaining required fields here or make them optional in the type instead of the blind cast.

src/lib/extensions/fs-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * True when the path is a file the operating system will run. Windows has no
+ * executable bit, so there existence is the only signal available.
+ */
+export async function isExecutable(path: string): Promise<boolean> {
+    try {
+        const stats = await stat(path)
+        if (!stats.isFile()) return false
+        return process.platform === 'win32' ? true : (stats.mode & 0o111) !== 0
```
doistbot: <a href="#">P2</a>  Use `fs.promises.access(path, constants.X_OK)` instead of deriving executability from `stat.mode`. Mode bits do not reflect the calling process's effective access under ACLs or other filesystem policy, while `access` checks that directly and already treats `X_OK` as an existence check on Windows.

src/lib/extensions/manifest.ts
```diff
@@ -0,0 +1,114 @@
...
+ * not a problem — both manifests are optional.
+ */
+export async function findManifestProblems(dir: string, binName: string): Promise<string[]> {
+    const problems: string[] = []
+    for (const fileName of [authoredManifestFileName(binName), manifestFileName(binName)]) {
+        const path = join(dir, fileName)
+        let raw: string
+        try {
+            raw = await readFile(path, 'utf8')
+        } catch {
```
doistbot: <a href="#">P2</a>  Only suppress a missing-file error here. This also suppresses permission and other I/O failures, so an unreadable manifest is reported exactly like an absent one and `doctor` returns no problem. Distinguish `ENOENT` from other read failures and report the latter.

doistbot (commented):
This PR lays down the extension core vocabulary — types, naming rules, manifest reading/writing, and per-extension state — cleanly split from the larger extensions effort, with solid test coverage and nothing wired up yet. Few things worth tightening:
- `readInstalledManifest` validates only `name` and `owner` before blindly casting to `InstalledManifest`; a truncated or hand-edited manifest (missing `tag`, `installedAt`, etc., or wrong types) slips through and later code reads undefined under a type that claims otherwise. Either validate the remaining required fields or make them optional.
- Prefer `fs.promises.access(path, constants.X_OK)` over deriving executability from `stat.mode` — mode bits don't reflect effective access under ACLs, while `access` checks that directly.
- In the manifest reader, only suppress `ENOENT`; currently any permission or I/O failure reads as "absent," so `doctor` misses an unreadable manifest.

<details>

- **<a href="#">P3</a> src/lib/extensions/source.ts:67**: SCP-style remotes preserve the host exactly as typed, so `parseSource('git@GitHub.com:Doist/td-goals.git')` yields `host: 'GitHub.com'` and `isGitHub: false`, while the URL branch lowercases via `new URL` and returns true. Hostnames are case-insensitive, so the official marker and the GitHub-API gate become dependent on how the user typed the remote. Lowercase `scpLike[1]` in `parseRepoRef` (or normalize the host before the `GITHUB_HOST` comparison).
- **<a href="#">P3</a> src/lib/extensions/fs-utils.ts:27**: Typo in the doc comment: "so there existence is the only signal available" should be "its existence".
- **<a href="#">P3</a> src/lib/extensions/state.ts:40**: `writeState` and `writeInstalledManifest` (manifest.ts:82-88) hand-roll the same JSON file write — `${JSON.stringify(x, null, 2)}\n` with `mode: 0o600` — and `readState`/`readJson` each re-implement tolerant JSON reads. Extract one shared `readJson`/`writeJson` helper in `src/lib/extensions/` and call it from both modules so the serialization format and file permissions can't drift.
- **<a href="#">P3</a> src/lib/extensions/state.ts:26**: This inline check (`typeof parsed !== 'object' || parsed === null || Array.isArray(parsed)`) re-implements the `isRecord` guard already defined in `manifest.ts`. Consider exporting `isRecord` from `manifest.ts` (or a tiny shared guard in this directory) so the two files can't drift on what counts as a JSON object.
- **<a href="#">P3</a> src/lib/extensions/state.ts:27**: `parsed as ExtensionState` accepts any object without checking field types, so `{"pinned": 5}` comes back typed as `pinned: string`. The state file lives in the user state dir and is read back as-is, so validate that `pinned`, `checkedForUpdateAt`, and `latestRelease` are strings when present before returning, or type the result more loosely.
- **<a href="#">P3</a> src/lib/extensions/manifest.ts:15**: `Promise<unknown | undefined>` collapses to `Promise<unknown>` in TypeScript — the `| undefined` is dead. `Promise<unknown>` says the same thing without implying the union does work.
- **<a href="#">P3</a> src/lib/extensions/source.test.ts:68**: The full-URL and scp-style `parseSource` cases use `toMatchObject` but never assert `url` (the non-GitHub case also omits `type` and `repo`). `url` is the value install will clone from, so a dropped or mangled URL would pass these tests. Assert the full object for the URL and scp cases, mirroring the `toEqual` used for the shorthand case.
- **<a href="#">P3</a> src/lib/extensions/source.test.ts:94**: The local-path matrix only exercises POSIX spellings, leaving `looksLikeLocalPath`'s Windows branches (`.\`, `..\`, and the drive-letter regex) uncovered. Add portable cases such as `.\td-scratch`, `..\td-scratch`, and `C:\code\td-scratch`; they resolve to `local` on every OS and would catch a regression in that branch.
- **<a href="#">P3</a> src/lib/extensions/state.ts:22**: state.ts has no tests. Two behaviors later slices depend on and that could silently regress: the read/write round-trip (writeState creates the `extensions` dir and readState returns what was written), and readState's tolerance of corrupt or non-object state files (returns `{}` instead of throwing, so a hand-edited state file can't crash commands that read the pin). A small test mirroring the manifest round-trip + corrupt-file cases in manifest.test.ts would pin both.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F521&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A446e03f0-ad22-11f1-9540-ce2a2d849b23%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A446e03f0-ad22-11f1-9540-ce2a2d849b23&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/manifest.ts
scottlovegrove: Done. readInstalledManifest now checks every field the type promises — owner, name, host, tag, asset and installedAt as strings, pinned as a boolean, sha256 as a string when present — and returns undefined otherwise. A truncated or hand-edited manifest now reads as "no manifest", which callers already handle, rather than as a manifest full of undefined.

src/lib/extensions/fs-utils.ts
scottlovegrove: Done. isExecutable now asks access(X_OK) after confirming the path is a file, so permission granted by an access-control list is honoured, and the Windows behaviour stays an existence check.

src/lib/extensions/manifest.ts
scottlovegrove: Done. Reading a JSON file now distinguishes absent, unreadable and parsed, and findManifestProblems reports "could not be read" for a manifest that exists but cannot be opened. Missing files stay silent, since both manifests are optional. Also fixed the scp-host casing, the doc typo, the shared readJson/writeJson and isRecord duplication, state field validation, and added state.test.ts covering the round trip, corrupt contents and wrong types.

engfragui:
@scottlovegrove Even though this is already broken up from a 3k lines PR (so much smaller in its current form), it does still feel pretty lengthy.
I'm not sure how to properly review this... What I ended up doing was feeding it to an agent, they could only thing only one minor thing, which I'll post on this PR (not blocking/priority).
Either way, ✅

src/lib/extensions/manifest.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    // Nothing to validate, so nothing to load a validator for.
+    if (dedicatedValue === undefined && !isRecord(packageJson)) return undefined
+
+    const { parseAuthoredManifest } = await schemas()
+    if (dedicatedValue !== undefined) {
+        const dedicated = parseAuthoredManifest(dedicatedValue)
+        if (dedicated) return dedicated
+    }
+    return isRecord(packageJson) ? parseAuthoredManifest(packageJson[binName]) : undefined
```
engfragui, line 49: **`readAuthoredManifest`: an empty/unknown-key dedicated manifest silently suppresses the `package.json` fallback**

If `td-extension.json` exists but parses to `{}` — e.g. it contains only keys from a *newer* manifest format, which the schema strips — `parseAuthoredManifest` returns a truthy `{}`, so this early-return fires and the `package.json[binName]` fallback is never consulted. An extension that keeps its `description` in `package.json` would lose it in that case.

"The dedicated file wins when both exist" is the documented intent, and this may well be what you want. But the edge isn't pinned by a test — `manifest.test.ts` "prefers the dedicated file when both exist" uses a *non-empty* dedicated file, so the empty/all-unknown-keys case isn't covered.

Is losing the `package.json` metadata intended there? If not, a `Object.keys(dedicated).length > 0` guard would fall through to the fallback instead.

_(Post-merge review of #521 — non-blocking.)_

scottlovegrove: Confirmed and fixed in #541 — not intended, and thank you for catching it.

Reproduced exactly as you describe: a `td-extension.json` containing only `{"somethingFromTheFuture": true}` alongside a `package.json` with a `td.description` showed `Extension fb` in `td --help` rather than the description.

The `Object.keys(dedicated).length > 0` guard on its own was not enough, which is worth recording: `package.json` was only read when the dedicated file was *absent*, so with the guard in place the fallback had nothing to fall back to and the result was `undefined` instead. The read now happens when the dedicated file yields nothing, and not eagerly otherwise.

Both cases are pinned by tests now — the all-unknown-keys file and a literal `{}`.

scottlovegrove:
Fair, and the agent pass was worth it — it found a real one, which I have fixed in #541 along with a second thing the same review turned up on #527.

On the length: you are right that it is still a lot. For scale, phase 1 came to roughly 10k lines — 7k in `src/lib/extensions/` across the first seven PRs, then 3k for the command layer that makes any of it reachable from `td`, which went up as a nine-PR stack (#531–#539). This PR is the first slice of the first half, so "smaller but still lengthy" is exactly what it is.

If there is a shape that would review better next time, I would genuinely rather know now than after the next one. The thing I have not found a good answer for is that the bottom of a stack is unavoidably a pile of types and helpers with no caller yet, so there is nothing to react to except the shape.

gnapse (approved):

PR #522, opened 2026-09-10
feat(extensions): discover installed extensions

scottlovegrove:
Slice 2 of 7, on top of slice 1.

Discovery walks the extensions directory and works out what each entry is: a symlink or path file is a local install, a `.git` directory is a clone, anything else is a release binary. It runs on every invocation of the CLI, including the common case of no extensions at all, so it costs one `readdir` plus a couple of small file reads and never spawns a process or touches the network. The git remote is read out of `.git/config` rather than by asking git.

Also adds the test fixture helpers, which build real extension directories in a temporary location so later slices can install and run something genuine rather than a mock.

### DIFF
--- a/src/lib/extensions/discover.ts
+++ b/src/lib/extensions/discover.ts
@@ -9,10 +9,12 @@
  */
 
 import { lstat, readdir, readFile, realpath } from 'node:fs/promises'
-import { isAbsolute, join, resolve } from 'node:path'
-import { exists } from './fs-utils.js'
+import { dirname, isAbsolute, join, resolve } from 'node:path'
+import { mapWithConcurrency } from './concurrency.js'
+import { exists, isDirectory } from './fs-utils.js'
+import { isMissingFile } from './json-file.js'
 import { readAuthoredManifest, readInstalledManifest } from './manifest.js'
-import { parseRepoRef, toCommandName } from './source.js'
+import { parseRepoRef, toCommandName, toDirName } from './source.js'
 import { readState } from './state.js'
 import type { Extension, ExtensionKind } from './types.js'
 
@@ -55,15 +57,20 @@ async function resolveLocalTarget(entryPath: string, isLink: boolean): Promise<s
     try {
         const target = (await readFile(entryPath, 'utf8')).trim()
         if (!target) return undefined
-        return isAbsolute(target) ? target : resolve(target)
+        // Relative to the path file itself, not to wherever the CLI happens to
+        // have been run from, so the install points at one directory always.
+        return isAbsolute(target) ? target : resolve(dirname(entryPath), target)
     } catch {
         return undefined
     }
 }
 
-async function inferKind(entryPath: string, isDirectory: boolean): Promise<ExtensionKind> {
-    if (!isDirectory) return 'local'
-    if (await exists(join(entryPath, '.git'))) return 'git'
+async function inferKind(entryPath: string, entryIsDirectory: boolean): Promise<ExtensionKind> {
+    if (!entryIsDirectory) return 'local'
+    // A directory, not merely a `.git` of some kind: a release binary is free
+    // to ship a file by that name, and misreading it as a clone would hide its
+    // manifest and its source.
+    if (await isDirectory(join(entryPath, '.git'))) return 'git'
     return 'binary'
 }
 
@@ -95,10 +102,13 @@ async function describe(entry: string, options: DiscoverOptions): Promise<Extens
     const dir =
         kind === 'local' ? ((await resolveLocalTarget(entryPath, isLink)) ?? entryPath) : entryPath
 
-    const [authored, installed, state] = await Promise.all([
+    // Every per-extension read at once: they are independent, and discovery
+    // runs on every invocation of the CLI.
+    const [authored, installed, state, remote] = await Promise.all([
         readAuthoredManifest(dir, binName),
         kind === 'binary' ? readInstalledManifest(dir, binName) : Promise.resolve(undefined),
         readState(stateDir, entry),
+        kind === 'git' ? readGitRemote(dir) : Promise.resolve(undefined),
     ])
 
     let host: string | undefined
@@ -110,7 +120,6 @@ async function describe(entry: string, options: DiscoverOptions): Promise<Extens
         owner = installed.owner
         source = `${installed.owner}/${installed.name}`
     } else if (kind === 'git') {
-        const remote = await readGitRemote(dir)
         const parsed = remote ? parseRepoRef(remote) : undefined
         host = parsed?.host
         owner = parsed?.owner
@@ -133,6 +142,7 @@ async function describe(entry: string, options: DiscoverOptions): Promise<Extens
         host,
         owner,
         pinned: Boolean(state.pinned ?? installed?.pinned),
+        pinnedRef: state.pinned ?? (installed?.pinned ? installed.tag : undefined),
         official,
         description: authored?.description ?? installed?.description,
         requires: authored?.requires ?? installed?.requires,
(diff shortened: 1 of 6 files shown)


src/lib/extensions/discover.ts
```diff
@@ -0,0 +1,159 @@
...
+        try {
+            return await realpath(entryPath)
+        } catch {
+            return undefined
+        }
+    }
+    try {
+        const target = (await readFile(entryPath, 'utf8')).trim()
+        if (!target) return undefined
+        return isAbsolute(target) ? target : resolve(target)
```
doistbot: <a href="#">P2</a>  A Windows-style path file holding a relative target is resolved via `resolve(target)`, which is relative to `process.cwd()`. The same local install then points at a different directory depending on where `td` was invoked. Resolve it against the path file's own directory instead: `resolve(dirname(entryPath), target)` (and import `dirname`).

doistbot (commented):
This slice adds extension discovery that classifies each entry in the extensions directory (local install via symlink/path file, git clone via `.git`, or release binary) with a lightweight, read-only scan that runs on every CLI invocation, plus test fixture helpers for building real extension directories. The implementation reuses the existing slice-1 helpers well and covers the main cases. Few things worth tightening:

- Resolve relative targets in path files against the path file's own directory (`resolve(dirname(entryPath), target)`) instead of `process.cwd()`, so local installs behave deterministically regardless of where `td` is invoked.
- Check that `.git` is a directory (e.g. via the existing `isDirectory()` helper), not just that it exists, so binary installs containing a `.git` file aren't misclassified as clones.
- Only treat a missing extensions directory (`ENOENT`) as "no extensions" and rethrow other `readdir` failures, so permission/I/O errors don't silently masquerade as an empty install.
- Bound the concurrency of per-entry discovery instead of fanning out with `Promise.all`, and in the Node fixture prefer `process.exitCode` (or a sync write) over `process.exit` so the JSON report isn't truncated on pipes.

<details>

- **<a href="#">P3</a> src/lib/extensions/discover.test.ts:101**: This test unconditionally creates a directory symlink, which fails on Windows without elevated rights or Developer Mode — the exact limitation the Windows-style path-file test works around. Guard it with `it.skipIf(process.platform === 'win32')` (or another POSIX-only guard) so the suite stays runnable on Windows.
- **<a href="#">P3</a> src/test-support/extension-fixture.ts:73**: This fixture re-derives the extension naming conventions that are already the source of truth in `src/lib/extensions/source.ts`: `${binName}-${name}` (line 73), `${binName}-extension.json` (line 84), and `.${binName}-manifest.json` (line 90). Import and use `toDirName`, `authoredManifestFileName`, and `manifestFileName` instead, so the fixture cannot drift from the real on-disk contract.
- **<a href="#">P3</a> src/lib/extensions/discover.ts:113**: For git installs, `readGitRemote` runs after the `Promise.all` that already reads the authored manifest and state, so the `.git/config` read is serialized behind those independent reads. Add it as a fourth entry in that `Promise.all` (resolving only for `kind === 'git'`) so all per-extension file reads complete concurrently.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F522&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A46bde760-ad22-11f1-95ea-95dacef21fdb%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A46bde760-ad22-11f1-95ea-95dacef21fdb&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/test-support/extension-fixture.ts
```diff
@@ -0,0 +1,108 @@
...
+// Dynamic import works whether node loads this as CommonJS or ESM, which
+// depends on whatever package.json happens to sit above the fixture.
+if (process.env.XX_REPORT) {
+    import('node:fs').then((fs) => {
+        fs.writeFileSync(process.env.XX_REPORT, report)
+        process.exit(code)
+    })
+} else {
+    process.stdout.write(report)
+    process.exit(code)
```
doistbot: <a href="#">P2</a>  `process.stdout.write(report)` is asynchronous when stdout is a pipe — the normal case when a test spawns this fixture — and `process.exit(code)` can exit before the write flushes, truncating the JSON report and causing flaky assertions. Set `process.exitCode = code` and drop `process.exit`, or write synchronously with `fs.writeSync(process.stdout.fd, report)`.

src/lib/extensions/discover.ts
```diff
@@ -0,0 +1,159 @@
...
+    } catch {
+        return undefined
+    }
+}
+
+async function inferKind(entryPath: string, isDirectory: boolean): Promise<ExtensionKind> {
+    if (!isDirectory) return 'local'
+    if (await exists(join(entryPath, '.git'))) return 'git'
```
doistbot: <a href="#">P2</a>  Check that `.git` is a directory here, rather than only that it exists. `exists()` also returns true for regular files, so an otherwise binary install containing a `.git` file is classified as `git`; this skips its installed manifest and leaves its source metadata unset. Use the existing `isDirectory()` helper (or a stat-based directory check) for this condition.

src/lib/extensions/discover.ts
```diff
@@ -0,0 +1,159 @@
...
+        manifest: installed,
+    }
+}
+
+/** Every extension installed for this CLI, sorted by command name. */
+export async function discoverExtensions(options: DiscoverOptions): Promise<Extension[]> {
+    let entries: string[]
+    try {
+        entries = await readdir(options.extensionsDir)
+    } catch {
```
doistbot: <a href="#">P2</a>  Swallowing every `readdir` failure makes permission or I/O errors look like "no extensions installed", so dispatch/list silently behave as if the directory were empty. The documented benign case is a missing directory, so catch only `ENOENT` and rethrow the rest.

src/lib/extensions/discover.ts
```diff
@@ -0,0 +1,159 @@
...
+        return []
+    }
+
+    const prefix = `${options.binName}-`
+    const candidates = entries.filter((entry) => entry.startsWith(prefix) && entry !== prefix)
+    const described = await Promise.all(candidates.map((entry) => describe(entry, options)))
```
doistbot: <a href="#">P2</a>  `Promise.all` starts discovery for every matching directory entry at once, and each discovery performs several filesystem operations. A directory with many `td-` entries can create an unbounded backlog of promises and queued fs work on every CLI invocation. Process candidates with a bounded concurrency limit while retaining the final sort.

src/lib/extensions/discover.ts
scottlovegrove: Done. A relative target in a path file now resolves against the path file's own directory, so the install points at one directory regardless of where the CLI was run.

src/test-support/extension-fixture.ts
scottlovegrove: Done. The fixture sets process.exitCode instead of calling process.exit, so a report written to a pipe is flushed. It also uses toDirName, authoredManifestFileName and manifestFileName rather than repeating the conventions.

src/lib/extensions/discover.ts
scottlovegrove: Done. The check is isDirectory now, so a release binary that ships a file called .git keeps its manifest and its source rather than being misread as a clone. Covered by a test.

src/lib/extensions/discover.ts
scottlovegrove: Done. Only ENOENT means "no extensions"; anything else is rethrown, so a permission or I/O problem is not reported as an empty install.

src/lib/extensions/discover.ts
scottlovegrove: Done. Discovery now runs through a worker pool of eight, shared with upgrades in concurrency.ts rather than duplicated. The per-extension reads, including the git remote, now go out together instead of the remote waiting behind the others. Also added findExtension, which describes only the entry asked for, so dispatch and remove no longer inspect everything installed.

frankieyan (approved):

gnapse (approved):

PR #523, opened 2026-09-10
feat(extensions): run extensions

scottlovegrove:
Slice 3 of 7, on top of slice 2.

The execution half of the contract: arguments reach the extension exactly as the user typed them, the three standard streams are the host's own, and the host exits with whatever the extension exited with.

Node-shebang scripts are launched with the host's own Node, which makes them work on Windows and pins them to the Node version the CLI already requires. Other scripts run directly on POSIX and through `sh` on Windows.

The environment adds the documented `TD_*` contract. No credential is ever injected: an extension that needs a token runs `td auth token view`. A token the user exported themselves is inherited like any other variable, which the spec records as a deliberate exception, and there is now a test pinning that behaviour so it cannot change by accident.

### DIFF
--- a/src/lib/extensions/dispatch.ts
+++ b/src/lib/extensions/dispatch.ts
@@ -1,0 +1,172 @@
+/**
+ * Running an extension.
+ *
+ * The contract is the process boundary: arguments are passed through exactly
+ * as the user typed them, the three standard streams are the host's own, and
+ * the host exits with whatever the extension exited with. Nothing is parsed,
+ * rewritten or added on either side.
+ */
+
+import { spawn } from 'node:child_process'
+import { open } from 'node:fs/promises'
+import { constants } from 'node:os'
+import { CliError } from '@doist/cli-core'
+import { isExecutable } from './fs-utils.js'
+import type { Extension } from './types.js'
+
+export type SpawnPlan = { command: string; args: string[] }
+
+/** First line of a file, when it is a shebang. */
+async function readShebang(path: string): Promise<string | undefined> {
+    let handle: Awaited<ReturnType<typeof open>>
+    try {
+        handle = await open(path, 'r')
+    } catch {
+        return undefined
+    }
+    try {
+        const buffer = Buffer.alloc(256)
+        const { bytesRead } = await handle.read(buffer, 0, buffer.length, 0)
+        const head = buffer.subarray(0, bytesRead).toString('utf8')
+        if (!head.startsWith('#!')) return undefined
+        return head.split('\n', 1)[0].trim()
+    } finally {
+        await handle.close()
+    }
+}
+
+/**
+ * Decide how to launch an extension's executable.
+ *
+ * Node scripts are launched with the host's own Node rather than through the
+ * shebang, which makes them work on Windows and pins them to the Node version
+ * the host already requires. Everything else runs directly on POSIX, and
+ * through `sh` on Windows, where shebangs mean nothing.
+ */
+export async function buildSpawnPlan(executablePath: string, args: string[]): Promise<SpawnPlan> {
+    const onWindows = process.platform === 'win32'
+
+    if (!(await isExecutable(executablePath))) {
+        throw new CliError(
+            'EXTENSION_NOT_EXECUTABLE',
+            `The executable for this extension is missing or not executable: ${executablePath}`,
+            {
+                hints: [
+                    'Compiled extensions need building after a source install.',
+                    'Script extensions need an executable bit: chmod +x the file.',
+                ],
+            },
+        )
+    }
+
+    if (onWindows && /\.(exe|cmd|bat)$/i.test(executablePath)) {
+        return { command: executablePath, args }
+    }
+
+    const shebang = await readShebang(executablePath)
+    if (shebang && /\bnode\b/.test(shebang)) {
+        return { command: process.execPath, args: [executablePath, ...args] }
+    }
+
+    if (onWindows) {
+        // Matches how `gh` runs script extensions on Windows: Git for Windows
+        // provides the interpreter, and without it a shebang script cannot run.
+        return {
+            command: 'sh',
+            args: ['-c', '"$0" "$@"', executablePath, ...args],
+        }
+    }
+
+    return { command: executablePath, args }
+}
+
+export type ExtensionEnvOptions = {
+    envPrefix: string
+    version: string
+    configDir: string
+    hostPath: string
+    extension: Extension
+    user?: string
+    accessible?: boolean
+    extra?: Record<string, string | undefined>
+}
+
+/**
+ * The variables an extension can rely on, layered onto the environment the
+ * user already has.
+ *
+ * The CLI never reads the credential store on an extension's behalf, so no
+ * token is ever injected here: an extension that needs one runs
+ * `<bin> auth token view`. A token the user has themselves exported stays in
+ * the environment, as every other variable does — the CLI does not inject
+ * secrets, but neither does it scrub the environment its user chose.
+ */
+export function buildExtensionEnv(options: ExtensionEnvOptions): NodeJS.ProcessEnv {
+    const { envPrefix: prefix, extension } = options
+    const env: NodeJS.ProcessEnv = {
+        ...process.env,
+        [`${prefix}_EXTENSION`]: '1',
+        [`${prefix}_EXTENSION_NAME`]: extension.name,
+        [`${prefix}_EXTENSION_DIR`]: extension.dir,
+        [`${prefix}_PATH`]: options.hostPath,
+        [`${prefix}_NODE`]: process.execPath,
+        [`${prefix}_VERSION`]: options.version,
+        [`${prefix}_CONFIG_DIR`]: options.configDir,
+    }
+
+    if (options.user) env[`${prefix}_USER`] = options.user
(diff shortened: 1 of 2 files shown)


src/lib/extensions/dispatch.ts
```diff
@@ -0,0 +1,172 @@
...
+                hints: [
+                    'Compiled extensions need building after a source install.',
+                    'Script extensions need an executable bit: chmod +x the file.',
+                ],
+            },
+        )
+    }
+
+    if (onWindows && /\.(exe|cmd|bat)$/i.test(executablePath)) {
+        return { command: executablePath, args }
```
doistbot, line 97: <a href="#">P2</a>  `spawn()` without `shell: true` cannot launch `.cmd`/`.bat` files on Windows — `CreateProcess` only runs PE executables — so this branch always fails and surfaces `EXTENSION_NOT_EXECUTABLE`. This is reachable because `discover.ts` already resolves `.cmd` extensions on Windows. Keep only `.exe` direct and run `.cmd`/`.bat` via `cmd.exe /d /s /c` (or add a `shell` flag to the spawn plan).

src/lib/extensions/dispatch.ts
```diff
@@ -0,0 +1,172 @@
...
+        [`${prefix}_EXTENSION`]: '1',
+        [`${prefix}_EXTENSION_NAME`]: extension.name,
+        [`${prefix}_EXTENSION_DIR`]: extension.dir,
+        [`${prefix}_PATH`]: options.hostPath,
+        [`${prefix}_NODE`]: process.execPath,
+        [`${prefix}_VERSION`]: options.version,
+        [`${prefix}_CONFIG_DIR`]: options.configDir,
+    }
+
+    if (options.user) env[`${prefix}_USER`] = options.user
```
doistbot, line 160: <a href="#">P2</a>  When `options.user` is absent, a `TD_USER` already present in `process.env` is inherited rather than removed, so a nested `td` invocation without `--user` still passes the outer invocation's user to its extensions. This contradicts the `forwards --user only when one was given` test, which only passes because the `beforeEach` stubs `TD_USER` to `undefined`. Delete `${prefix}_USER` when no user is supplied, matching the delete-on-`undefined` behavior in the `extra` loop.

doistbot (commented):
This slice wires up extension execution: arguments pass through verbatim, the host shares the child's stdio streams, exit codes propagate, and the environment gets the `TD_*` contract with no credentials injected. Security-wise this looks solid — argument-array spawning with no shell interpolation on POSIX, and the Windows `sh -c` fallback keeps user input as positional parameters, so nothing user-controlled can escape into a command.

Few things worth tightening:
- The Windows `.cmd`/`.bat` branch can't work: `spawn()` without `shell: true` can't launch these via `CreateProcess`, so valid extensions resolved by discovery will always fail with `EXTENSION_NOT_EXECUTABLE`. Run them through `cmd.exe /d /s /c` instead.
- `buildExtensionEnv` leaks an inherited `TD_USER` into nested `td` invocations when no `--user` is given — the existing test only passes because it stubs the variable to `undefined`. Delete `${prefix}_USER` when no user is supplied, matching the `extra` loop's behavior.
- Node-shebang replacement drops the interpreter's options, so directives like `#!/usr/bin/node --conditions=development` break. Parse the Node directive (including `env -S` forms) and prepend its options before the script path.

<details>

- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:136**: `expect(env.TD_EXTENSION).toBe(process.env.TD_EXTENSION)` can never fail: `buildExtensionEnv` spreads `process.env`, so the two are identical by construction. If the intent is "the TDC prefix doesn't leak TD_* variables", assert `expect(env.TD_EXTENSION).toBeUndefined()`.
- **<a href="#">P3</a> src/lib/extensions/dispatch.ts:91**: `ExtensionEnvOptions` re-declares `user` and `extra` on top of the existing `DispatchOptions` in `./types.ts`, which already defines `user` and `env` with the same "merged last" child-environment semantics. Compose `DispatchOptions` (or reuse its fields) and keep one name for the child env override — the current `env`/`extra` split is an easy way for the dispatch and manager layers to drift.
- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:112**: Hardcoded `'TODOIST_API_TOKEN'` duplicates the canonical `TOKEN_ENV_VAR` from `../auth-store.js` (also re-exported by `../auth.js`). Import that constant so the pass-through test stays in sync if the token env var name changes.
- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:41**: This test asserts POSIX-only behavior but has no platform guard, so running `npm test` on Windows fails at `expect(plan.command).toBe(join(dir, 'td-shell'))` because the Windows branch routes through `sh`. Skip it on Windows (e.g., `it.skipIf(process.platform === 'win32')`) or branch the expectation.
- **<a href="#">P3</a> src/lib/extensions/dispatch.ts:49**: `buildSpawnPlan` stats the executable here via `isExecutable`, then `readShebang` opens the same path again on line 66. Since `readShebang` already has a file handle, it can call `handle.stat()` there to answer both "is this a runnable file" and "what is its shebang" in a single open. That removes the duplicate I/O and closes the TOCTOU window where the file changes between the two checks.
- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:22**: `buildSpawnPlan`'s Windows branches have no coverage: `.exe/.cmd/.bat` files that should run directly, and every other script that should become `sh -c '"$0" "$@"'`. That `sh` quoting is what preserves argument passthrough on Windows, and CI runs only on Linux, so a regression there would pass silently. Add a platform-forced test (e.g., temporarily redefine `process.platform` as `win32` and restore it) covering both branches.
- **<a href="#">P3</a> src/lib/extensions/dispatch.test.ts:195**: The signal branch in dispatchExtension is untested: a child killed by a signal must yield 128+signal (e.g. 130 for SIGINT), and a regression that resolves with `code ?? 0` would make the host silently report success after an extension is killed. A fixture that runs `process.kill(process.pid, 'SIGKILL')` (POSIX) would pin this. The three `--exit-code` cases at lines 196-198 only cover the normal-close path.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F523&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A486be580-ad22-11f1-9f91-9b54fb6e3a2a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A486be580-ad22-11f1-9f91-9b54fb6e3a2a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/dispatch.ts
```diff
@@ -0,0 +1,172 @@
...
+        )
+    }
+
+    if (onWindows && /\.(exe|cmd|bat)$/i.test(executablePath)) {
+        return { command: executablePath, args }
+    }
+
+    const shebang = await readShebang(executablePath)
+    if (shebang && /\bnode\b/.test(shebang)) {
+        return { command: process.execPath, args: [executablePath, ...args] }
```
doistbot: <a href="#">P2</a>  Preserve options from a Node shebang when replacing its interpreter. For example, `#!/usr/bin/node --conditions=development` requires that option before the script path; this plan drops it and can cause module resolution failures. Parse the Node directive (including `env -S` forms) and prepend its Node options before `executablePath`.

src/lib/extensions/dispatch.ts
scottlovegrove: Done. Only .exe runs directly now; .cmd and .bat go through the command interpreter, which is the only thing that can run them. Covered by tests that declare the platform, along with the sh fallback and the node shortcut.

src/lib/extensions/dispatch.ts
scottlovegrove: Done. The variable is now set or deleted, never left inherited, so a nested call without --user does not keep acting as the account the outer call chose. There is a test that stubs an inherited value and asserts it is cleared.

src/lib/extensions/dispatch.ts
scottlovegrove: Done. The shebang is parsed rather than pattern-matched, including the env -S form, and any node options it carries are passed before the script path. Also merged the executable check into the same open as the shebang read, which removes the second stat and the window between the two.

craigcarlyle:
✅

gnapse (approved):

PR #524, opened 2026-09-10
feat(extensions): git, npm and GitHub helpers, and removal

scottlovegrove:
Slice 4 of 7, on top of slice 3.

The outside world an install has to talk to, plus the one operation that needs no installer.

- A subprocess wrapper that captures a bounded tail of output, since npm runs extension lifecycle scripts that could otherwise print without end.
- Git operations, with working-tree state reported as clean, dirty or unknown, so that "could not tell" is never mistaken for "safe to delete".
- npm dependency installation, run without the CLI's own credentials, invoked through the command interpreter on Windows because a `.cmd` shim cannot be spawned directly on supported Node versions, and without writing a lockfile into a clone that has none.
- The GitHub client: releases, asset downloads and checksum verification.
- Removal, which deletes only the link for a local install and refuses a clone with uncommitted or unreadable state unless forced.

### DIFF
--- a/src/lib/extensions/git.ts
+++ b/src/lib/extensions/git.ts
@@ -1,0 +1,114 @@
+/**
+ * The git operations the extension system needs. Every one of them shells out
+ * to the user's own git, so clones inherit whatever credential helper and SSH
+ * configuration they already have.
+ */
+
+import { CliError } from '@doist/cli-core'
+import { outputHints, run, type RunResult } from './run.js'
+
+function requireGit(result: RunResult): void {
+    if (result.missing) {
+        throw new CliError('EXTENSION_INSTALL_FAILED', 'git was not found on PATH.', {
+            hints: ['Install git, or install this extension from a release binary instead.'],
+        })
+    }
+}
+
+export async function clone(url: string, destination: string): Promise<void> {
+    const result = await run('git', ['clone', '--quiet', url, destination])
+    requireGit(result)
+    if (result.code !== 0) {
+        throw new CliError('EXTENSION_NOT_INSTALLABLE', `Could not clone ${url}.`, {
+            hints: outputHints(result),
+        })
+    }
+}
+
+export async function checkout(dir: string, ref: string): Promise<void> {
+    const result = await run('git', ['checkout', '--quiet', ref], { cwd: dir })
+    requireGit(result)
+    if (result.code !== 0) {
+        throw new CliError('EXTENSION_NOT_INSTALLABLE', `Could not check out "${ref}".`, {
+            hints: outputHints(result),
+        })
+    }
+}
+
+/** Full commit SHA of HEAD, or undefined when the directory is not a clone. */
+export async function headSha(dir: string): Promise<string | undefined> {
+    const result = await run('git', ['rev-parse', 'HEAD'], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim() || undefined
+}
+
+/**
+ * Whether the working tree holds changes the user would not want deleted.
+ *
+ * `unknown` is its own answer rather than being folded into `clean`: git may
+ * be missing, the repository may be flagged as unsafe to use, or `.git` may be
+ * damaged, and in each of those cases the honest position is that the state
+ * could not be established — callers treat that as a reason to stop, not a
+ * licence to delete.
+ */
+export type WorkingTreeState = 'clean' | 'dirty' | 'unknown'
+
+export async function workingTreeState(dir: string): Promise<WorkingTreeState> {
+    const result = await run('git', ['status', '--porcelain'], { cwd: dir })
+    if (result.missing || result.code !== 0) return 'unknown'
+    return result.stdout.trim().length > 0 ? 'dirty' : 'clean'
+}
+
+export type PullResult = { changed: boolean; from?: string; to?: string; error?: string }
+
+/** Fast-forward the clone. Never rewrites local work; `resetToRemote` does that. */
+export async function pull(dir: string): Promise<PullResult> {
+    const from = await headSha(dir)
+    const result = await run('git', ['pull', '--quiet', '--ff-only'], { cwd: dir })
+    requireGit(result)
+    if (result.code !== 0) {
+        return { changed: false, from, error: outputHints(result).join(' ') }
+    }
+    const to = await headSha(dir)
+    return { changed: from !== to, from, to }
+}
+
+/** Discard local commits and working-tree changes in favour of the remote. */
+export async function resetToRemote(dir: string): Promise<PullResult> {
+    const from = await headSha(dir)
+    const fetched = await run('git', ['fetch', '--quiet', 'origin'], { cwd: dir })
+    requireGit(fetched)
+    if (fetched.code !== 0) {
+        return { changed: false, from, error: outputHints(fetched).join(' ') }
+    }
+    const reset = await run('git', ['reset', '--quiet', '--hard', 'origin/HEAD'], { cwd: dir })
+    if (reset.code !== 0) {
+        return { changed: false, from, error: outputHints(reset).join(' ') }
+    }
+    const to = await headSha(dir)
+    return { changed: from !== to, from, to }
+}
+
+/** Commit the remote's default branch points at, without fetching it. */
+export async function remoteHeadSha(dir: string): Promise<string | undefined> {
+    const result = await run('git', ['ls-remote', 'origin', 'HEAD'], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim().split(/\s+/)[0] || undefined
+}
+
+/** Resolve a ref to the SHA it names, used when pinning a clone. */
+export async function resolveRef(dir: string, ref: string): Promise<string | undefined> {
+    const result = await run('git', ['rev-parse', ref], { cwd: dir })
+    if (result.code !== 0) return undefined
+    return result.stdout.trim() || undefined
+}
+
+/** Files that changed between two commits, used to decide if deps need reinstalling. */
+export async function changedFiles(dir: string, from: string, to: string): Promise<string[]> {
+    const result = await run('git', ['diff', '--name-only', `${from}..${to}`], { cwd: dir })
+    if (result.code !== 0) return []
+    return result.stdout
+        .split('\n')
+        .map((line) => line.trim())
+        .filter(Boolean)
+}
--- a/src/lib/extensions/git.ts
+++ b/src/lib/extensions/git.ts
@@ -7,6 +7,33 @@
 import { CliError } from '@doist/cli-core'
 import { outputHints, run, type RunResult } from './run.js'
 
+/**
+ * Remove credentials from a URL before it is shown to anyone.
+ *
+ * A private-repo remote can legitimately carry `user:token@`, and a failed
+ * clone would otherwise print that token to the terminal and into any log
+ * that captured it.
+ */
+export function redactUrl(url: string): string {
+    return url.replace(/(\w+:\/\/)[^/@\s]*@/, '$1***@')
+}
+
+/**
+ * Refuse an argument that git would read as an option.
+ *
+ * Repository URLs and refs arrive from manifests and from the command line. A
+ * value starting with a dash would otherwise choose git's behaviour rather
+ * than being the thing git acts on, and for local transports some of those
+ * options run a command.
+ */
+function requirePositional(value: string, what: string): void {
+    if (value.startsWith('-')) {
+        throw new CliError('EXTENSION_NOT_INSTALLABLE', `A ${what} cannot start with "-".`, {
+            hints: [`Got: ${redactUrl(value)}`],
+        })
+    }
+}
+
 function requireGit(result: RunResult): void {
     if (result.missing) {
         throw new CliError('EXTENSION_INSTALL_FAILED', 'git was not found on PATH.', {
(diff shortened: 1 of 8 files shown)


src/lib/extensions/npm.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const hasLockfile = await exists(join(dir, 'package-lock.json'))
+    const args = hasLockfile ? ['ci', '--omit=dev'] : ['install', '--omit=dev', '--no-package-lock']
+    const { command, prefix } = npmCommand()
+    const result = await run(command, [...prefix, ...args], {
+        cwd: dir,
+        env: environmentWithoutSecrets(),
+    })
+
+    if (result.missing) {
```
doistbot, line 97: <a href="#">P2</a>  On Windows `npm` is invoked through `cmd.exe`, so `run` only reports `missing` when `cmd.exe` itself cannot be spawned. A machine without npm therefore falls through to the generic `EXTENSION_INSTALL_FAILED` branch with cmd's raw "not recognized" output, and the dedicated `EXTENSION_NPM_MISSING` message and hints are unreachable on Windows. Detect npm before installing (for example via `cmd /c where npm`) or map cmd's not-found exit code/output to `EXTENSION_NPM_MISSING`.

src/lib/extensions/git.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/** Discard local commits and working-tree changes in favour of the remote. */
+export async function resetToRemote(dir: string): Promise<PullResult> {
+    const from = await headSha(dir)
+    const fetched = await run('git', ['fetch', '--quiet', 'origin'], { cwd: dir })
+    requireGit(fetched)
+    if (fetched.code !== 0) {
+        return { changed: false, from, error: outputHints(fetched).join(' ') }
+    }
+    const reset = await run('git', ['reset', '--quiet', '--hard', 'origin/HEAD'], { cwd: dir })
```
doistbot: <a href="#">P2</a>  `resetToRemote` resets to `origin/HEAD`, but `git fetch origin` does not update that symref — it is set at clone time and only refreshed by `git remote set-head`. If a repo renames its default branch after the clone, `origin/HEAD` still points at the old, now stale branch, so this can reset to the wrong commit or fail. Use `remoteHeadSha(dir)` (or refresh `origin/HEAD`) and reset to the SHA it returns.

doistbot: <a href="#">P2</a>  `git reset --hard` leaves untracked files intact. A forced reset of a clone with a new untracked source file therefore retains local code despite this helper's stated guarantee to discard working-tree changes. Follow the reset with an appropriate `git clean` invocation (including a deliberate policy for ignored files).

src/lib/extensions/git.ts
```diff
@@ -0,0 +1,114 @@
...
+function requireGit(result: RunResult): void {
+    if (result.missing) {
+        throw new CliError('EXTENSION_INSTALL_FAILED', 'git was not found on PATH.', {
+            hints: ['Install git, or install this extension from a release binary instead.'],
+        })
+    }
+}
+
+export async function clone(url: string, destination: string): Promise<void> {
+    const result = await run('git', ['clone', '--quiet', url, destination])
```
doistbot: <a href="#">P2</a>  `url` is passed as a bare positional, so a value starting with `-` is parsed by git as an option rather than a URL — e.g. `--upload-pack=<cmd>` (executed locally for local transports) instead of being treated as a repository. The same applies to `ref` in `checkout`/`resolveRef` if it starts with `-`. Reject arguments with a leading `-` (or insert `--` before positional args) so a manifest- or user-supplied string can never change which git options run.

src/lib/extensions/github.ts
```diff
@@ -0,0 +1,191 @@
...
+
+        async fetchAssetText(asset) {
+            const response = await request(asset.url, 'application/octet-stream')
+            if (!response.ok) throw failed(response, `asset ${asset.name}`)
+            return response.text()
+        },
+
+        async fetchRepoFile(owner, repo, path, ref) {
+            const query = ref ? `?ref=${encodeURIComponent(ref)}` : ''
+            const url = `${API_ROOT}/repos/${owner}/${repo}/contents/${path}${query}`
```
doistbot: <a href="#">P2</a>  `path` is interpolated into the URL without encoding, so a value containing `../` traverses out of `/contents/` to any other api.github.com path — e.g. path `../../../other-owner/other-repo/contents/file` makes this authenticated request (it carries the user's GH_TOKEN/GITHUB_TOKEN via `authHeaders()`) read a file from a different repo. A `?` in the path would also swallow the `?ref=` query. Encode each path segment with `encodeURIComponent` before interpolating.

src/lib/extensions/npm.ts
```diff
@@ -0,0 +1,85 @@
...
+ * npm legitimately needs a large and open-ended slice of the environment
+ * (proxies, certificate bundles, registry credentials, package-manager
+ * settings), so naming what must not travel is both safer in practice and
+ * easier to keep correct than naming everything that may.
+ */
+const SECRET_ENV_VARS = ['TODOIST_API_TOKEN', 'GH_TOKEN', 'GITHUB_TOKEN']
+
+function environmentWithoutSecrets(): NodeJS.ProcessEnv {
+    const env = { ...process.env }
+    for (const name of SECRET_ENV_VARS) delete env[name]
```
doistbot: <a href="#">P1</a>  Remove secret variables case-insensitively. On Windows `process.env` lookup is case-insensitive, so a credential set as `gh_token` or `Todoist_Api_Token` is still used by the CLI, but this exact-key deletion leaves it in the copied environment passed to npm lifecycle scripts. Filter environment keys by an uppercased denylist before spawning npm.

src/lib/extensions/github.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    const combined = release.assets.find(
+        (asset) => asset.name === 'checksums.txt' || asset.name.endsWith('_checksums.txt'),
+    )
+    if (!combined) return undefined
+
+    const text = await client.fetchAssetText(combined)
+    for (const line of text.split('\n')) {
+        const [hash, ...rest] = line.trim().split(/\s+/)
```
doistbot: <a href="#">P2</a>  This only recognizes coreutils-style checksum entries (`<hash> <filename>`). A valid BSD/macOS `checksums.txt` entry such as `SHA256 (asset-name) = <hash>` never matches `assetName`, so this returns `undefined` and `verifyChecksum` skips integrity verification entirely. Parse the BSD format as well, or reject an unparseable checksum file rather than treating it as absent.

src/lib/extensions/npm.test.ts
```diff
@@ -0,0 +1,104 @@
...
+describe.skipIf(process.platform === 'win32')('installDependencies', () => {
+    let root: string
+    let extensionDir: string
+    let recordPath: string
+
+    async function fakeNpmOnPath(exitCode = 0): Promise<void> {
+        const binDir = join(root, 'bin')
+        await writeFile(
+            join(binDir, 'npm'),
+            `#!/bin/sh\n{ echo "ARGS:$*"; env; } > "${recordPath}"\nexit ${exitCode}\n`,
```
doistbot: <a href="#">P2</a>  This credential test passes vacuously. `fakeNpmOnPath` stubs PATH to only the fake `bin` directory, so the script's `env` command isn't on PATH and its environment dump never reaches the record file. The `not.toContain(...)` assertions then pass whether or not the secrets leaked. Use a shell builtin that doesn't depend on PATH (for example `export -p` or `set`) for the dump, or append the original PATH after `binDir` so `env` is still resolvable.

doistbot (commented):
This slice adds the external-facing pieces of the extension system: a bounded subprocess wrapper, git helpers with explicit working-tree state, credential-stripped npm installs, a GitHub client with checksum verification, and safe removal for local installs. Few things worth tightening:

- Windows npm detection is effectively dead code — running through `cmd.exe` means a missing npm surfaces as a generic install failure, so detect npm up front or map cmd's not-found output to `EXTENSION_NPM_MISSING`.
- `resetToRemote` can reset to a stale `origin/HEAD` after a default-branch rename, and `git reset --hard` leaves untracked files behind; use the remote HEAD SHA and follow with a deliberate `git clean` policy.
- Harden externally-influenced inputs: reject leading-`-` args to git (or insert `--`), and encode repo-file paths before interpolating into GitHub API URLs to prevent traversal.
- Close the smaller gaps: strip secret env vars case-insensitively (relevant on Windows), parse BSD-style checksum entries (or fail loudly instead of skipping verification), and fix the credential-stripping test whose environment dump never actually runs.

<details>

- **<a href="#">P3</a> src/lib/extensions/git.ts:22**: `url` is interpolated into the user-facing error verbatim. Git URLs can legitimately embed credentials (for example `https://user:TOKEN@host/owner/repo.git`), so a failed clone prints the token. Strip the `user:password@` userinfo from the URL before including it in the message, and apply the same redaction to any git output surfaced through `outputHints(result)`.
- **<a href="#">P3</a> src/lib/extensions/git.ts:39**: `headSha` is exactly `resolveRef(dir, 'HEAD')`: both run `git rev-parse`, return `undefined` on failure, and trim stdout. Implement it as `return resolveRef(dir, 'HEAD')` (or drop it and call `resolveRef` at the two call sites) so the git wrapper doesn't carry two copies of the same logic.
- **<a href="#">P3</a> src/lib/extensions/npm.ts:27**: Reuse the exported `TOKEN_ENV_VAR` constant from `src/lib/auth-store.ts` instead of hardcoding the `'TODOIST_API_TOKEN'` string here. The constant exists precisely so the env-var name has one source of truth; if it ever changes, this denylist would silently stop scrubbing the real token.
- **<a href="#">P3</a> src/lib/extensions/git.ts:109**: `changedFiles` returns `[]` for any non-zero `git diff`, which makes "no files changed" indistinguishable from "could not run git" (missing git, bad refs, corrupt repo). A caller using this to decide whether to reinstall dependencies would then skip reinstall exactly when it cannot tell. Return a result that carries the failure, or throw, so errors aren't folded into an empty diff.
- **<a href="#">P3</a> src/lib/extensions/run.ts:54**: `chunk.toString()` per data event can split a multibyte UTF-8 sequence across chunk boundaries, garbling the captured tail (npm progress bars and box-drawing output are common offenders). Either call `child.stdout.setEncoding('utf8')` / use `node:string_decoder`, or drop the `MAX_CAPTURED_BYTES` name since the bound is actually measured in UTF-16 code units, not bytes.
- **<a href="#">P3</a> src/lib/extensions/npm.test.ts:100**: The test name promises that npm's output is surfaced as hints, but it only asserts the `EXTENSION_INSTALL_FAILED` code. The fake npm writes its only output to `recordPath`, so `run` captures no stdout/stderr and `outputHints` has nothing to return. Have the fake emit a recognizable line to stderr before exiting non-zero and assert it appears in `error.hints`, or drop the output claim from the name.
- **<a href="#">P3</a> src/lib/extensions/run.ts:40**: `run`'s bounded-tail guarantee (keep only the last `MAX_CAPTURED_BYTES` while still draining the stream) has no test. That is the property that prevents a chatty or looping npm lifecycle script from exhausting the heap, so a regression would be silent. Add a test that spawns a script emitting more than 64 KiB and asserts the result resolves with a truncated tail.
- **<a href="#">P3</a> src/lib/extensions/remove.test.ts:68**: This test creates a real directory symlink, which throws EPERM on Windows without admin rights or Developer Mode. The suite already acknowledges platform differences elsewhere (npm.test.ts uses `skipIf(process.platform === 'win32')`), and discovery explicitly supports the plain-file link form on Windows, so this test will fail for any Windows contributor running the suite locally. Either skip it on win32 or write the Windows-style link file (a plain file containing the target path) so the test runs on both platforms.
- **<a href="#">P3</a> src/lib/extensions/github.ts:142**: Checksum verification ships with no tests, and its failure mode is silent: if `findExpectedChecksum` stops matching (e.g. the per-asset `<name>.sha256` lookup or the checksums.txt line parsing regresses), it returns `undefined` and `verifyChecksum` skips verification entirely — a corrupt or tampered asset installs without an error. `findExpectedChecksum` and `verifyChecksum` are pure functions over a client that already accepts an injectable `fetchImpl`, so covering the mismatch-throws, no-checksum-passes, and both checksum-file formats cases is straightforward and protects a security boundary.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F524&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4ae32710-ad22-11f1-9c0a-87fc08962911%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4ae32710-ad22-11f1-9c0a-87fc08962911&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/npm.ts
scottlovegrove: Done. The denylist is compared against the uppercased key, so a token exported as gh_token is stripped too. The test now sets a lowercase variable and asserts it does not reach npm.

src/lib/extensions/npm.ts
scottlovegrove: Done. npm availability is probed before installing, so the Windows path reaches EXTENSION_NPM_MISSING instead of surfacing the interpreter's own "not recognized" wording as a generic install failure.

src/lib/extensions/git.ts
scottlovegrove: Done. The reset now targets the SHA that git ls-remote reports, not origin/HEAD, which is written at clone time and never refreshed by a fetch.

scottlovegrove: Done. The reset is followed by git clean -fd. Ignored files are deliberately left alone: they are build output and caches, node_modules among them, not local work.

src/lib/extensions/git.ts
scottlovegrove: Done. A URL or ref starting with a dash is refused, and "--" separates the positional arguments, so a value from a manifest can never choose which git options run.

src/lib/extensions/github.ts
scottlovegrove: Done, though not by encoding alone: encodeURIComponent leaves ".." intact because dots are unreserved. The path is now split into segments, any "." or ".." segment is refused outright, and the remaining segments are encoded. Covered by a test asserting no request is made for a traversing path.

src/lib/extensions/github.ts
scottlovegrove: Done. The BSD form is parsed as well as the coreutils one, and a release that publishes checksums but names none for the asset being installed is now refused rather than installed unverified under a misleading warning.

src/lib/extensions/npm.test.ts
scottlovegrove: Good catch, that test was passing vacuously. The fake npm now dumps with export -p, a shell builtin that does not depend on PATH, and there is a canary test asserting the dump really contains the environment, so the credential assertions cannot silently stop testing anything.

gnapse (approved):

rmartins90:
✅

PR #525, opened 2026-09-10
feat(extensions): install extensions

scottlovegrove:
Slice 5 of 7, on top of slice 4.

Installs from a release binary, a git clone, or a local directory.

Everything is assembled in a staging directory and moved into place only once complete, with the previous install moved aside rather than deleted, so a failure part-way through leaves the working extension recoverable. The trust warning is printed before anything is downloaded, cloned or executed.

Release assets are matched by platform and architecture, verified against the release's checksums when it publishes any, and recorded in a manifest that carries the author's description and version requirement.

### DIFF
--- a/src/lib/extensions/install.test.ts
+++ b/src/lib/extensions/install.test.ts
@@ -7,7 +7,7 @@ import { writeFixtureExtension } from '../../test-support/extension-fixture.js'
 import { createGitHubClient } from './github.js'
 import { installExtension, type InstallContext } from './install.js'
 import { run } from './run.js'
-import { readState } from './state.js'
+import { readState, writeState } from './state.js'
 
 // Resolved once: a machine without git skips the clone tests visibly rather
 // than passing them vacuously.
@@ -241,22 +241,65 @@ describe('installExtension', () => {
             await expect(
                 installExtension('Doist/td-goals', { force: true }, contextFor(github.impl)),
             ).resolves.toMatchObject({ kind: 'binary' })
+
+            // The fixture really is gone, not merely reported as replaced.
+            const executable = await readFile(join(extensionsDir, 'td-goals', 'td-goals'))
+            expect(executable.equals(BINARY)).toBe(true)
+        })
+
+        it('puts the previous install back when the replacement cannot be finished', async () => {
+            await mkdir(extensionsDir, { recursive: true })
+            await writeFixtureExtension(extensionsDir, 'goals', {
+                authoredManifest: { description: 'the original' },
+            })
+            const github = stubGitHub()
+            const context = contextFor(github.impl)
+            // The state directory is a file, so recording the install fails
+            // after the new version is already in place.
+            await writeFile(stateDir, 'not a directory')
+
+            await expect(
+                installExtension('Doist/td-goals', { force: true }, context),
+            ).rejects.toThrow()
+
+            const manifest = JSON.parse(
+                await readFile(join(extensionsDir, 'td-goals', 'td-extension.json'), 'utf8'),
+            )
+            expect(manifest.description).toBe('the original')
         })
     })
 
     describe('local installs', () => {
-        it('links the directory instead of copying it', async () => {
+        it.skipIf(process.platform === 'win32')(
+            'links the directory instead of copying it',
+            async () => {
+                const workspace = join(root, 'code')
+                await mkdir(workspace, { recursive: true })
+                const target = await writeFixtureExtension(workspace, 'scratch')
+                const github = stubGitHub()
+
+                const result = await installExtension(target, {}, contextFor(github.impl))
+
+                expect(result).toMatchObject({ name: 'scratch', kind: 'local', dir: target })
+                const entry = join(extensionsDir, 'td-scratch')
+                expect((await lstat(entry)).isSymbolicLink()).toBe(true)
+                expect(github.calls).toEqual([])
+            },
+        )
+
+        it('clears state left behind by whatever was installed under that name', async () => {
             const workspace = join(root, 'code')
             await mkdir(workspace, { recursive: true })
-            const target = await writeFixtureExtension(workspace, 'scratch')
-            const github = stubGitHub()
+            const target = await writeFixtureExtension(workspace, 'goals')
+            await mkdir(extensionsDir, { recursive: true })
+            await writeFixtureExtension(extensionsDir, 'goals')
+            await writeState(stateDir, 'td-goals', { pinned: 'v1.0.0' })
 
-            const result = await installExtension(target, {}, contextFor(github.impl))
+            await installExtension(target, { force: true }, contextFor(stubGitHub().impl))
 
-            expect(result).toMatchObject({ name: 'scratch', kind: 'local', dir: target })
-            const entry = join(extensionsDir, 'td-scratch')
-            expect((await lstat(entry)).isSymbolicLink()).toBe(true)
-            expect(github.calls).toEqual([])
+            // A stale pin would otherwise make the new local install report as
+            // pinned and be skipped by every upgrade.
+            await expect(readState(stateDir, 'td-goals')).resolves.toEqual({})
         })
 
         it('warns when the directory has no executable yet, but still links it', async () => {
@@ -277,6 +320,16 @@ describe('installExtension', () => {
             await expect(
                 installExtension(workspace, {}, contextFor(github.impl)),
             ).rejects.toMatchObject({ code: 'EXTENSION_NAME_INVALID' })
+        })
+
+        it('refuses a plain file, which could never be dispatched', async () => {
+            const file = join(root, 'code', 'td-file')
+            await mkdir(join(root, 'code'), { recursive: true })
+            await writeFile(file, 'not a directory')
+
+            await expect(
+                installExtension(file, {}, contextFor(stubGitHub().impl)),
+            ).rejects.toMatchObject({ code: 'EXTENSION_NOT_INSTALLABLE' })
         })
 
         it('refuses a directory that does not exist', async () => {
@@ -312,7 +365,7 @@ describe('installExtension', () => {
             return repo
         }
 
-        it('clones when the repository publishes no matching asset', async () => {
+        it('clones a repository given as a file URL', async () => {
             const repo = await makeRepo('td-cloned')
             const github = stubGitHub({ assetName: 'td-cloned_v1_other-arch' })
 
@@ -321,6 +374,34 @@ describe('installExtension', () => {
             expect(result).toMatchObject({ kind: 'git', name: 'cloned' })
             await expect(stat(join(extensionsDir, 'td-cloned', 'td-cloned'))).resolves.toBeTruthy()
             await expect(stat(join(extensionsDir, 'td-cloned', '.git'))).resolves.toBeTruthy()
+        })
+
+        it('refuses a repository whose entry point is not executable', async () => {
+            const repo = join(root, 'origin', 'Doist', 'td-unexecutable')
+            await mkdir(repo, { recursive: true })
+            await writeFile(join(repo, 'td-unexecutable'), '#!/bin/sh\necho hi\n', { mode: 0o644 })
+            await run('git', ['init', '--quiet', '--initial-branch=main'], { cwd: repo })
+            await run('git', ['add', '.'], { cwd: repo })
+            await run(
+                'git',
+                [
+                    '-c',
+                    'user.email=t@example.com',
+                    '-c',
+                    'user.name=T',
+                    'commit',
+                    '--quiet',
+                    '-m',
+                    'init',
+                ],
+                { cwd: repo },
+            )
+
+            await expect(
+                installExtension(`file://${repo}`, {}, contextFor(stubGitHub().impl)),
+            ).rejects.toMatchObject({ code: 'EXTENSION_NOT_INSTALLABLE' })
+
+            await expect(stat(join(extensionsDir, 'td-unexecutable'))).rejects.toThrow()
         })
 
         it('refuses a repository with no executable at its root, leaving nothing behind', async () => {
--- a/src/lib/extensions/install.test.ts
+++ b/src/lib/extensions/install.test.ts
@@ -126,6 +126,7 @@ describe('installExtension', () => {
                 await readFile(join(extensionsDir, 'td-goals', '.td-manifest.json'), 'utf8'),
             )
             expect(manifest).toMatchObject({
+                manifestVersion: 1,
                 owner: 'Doist',
                 name: 'td-goals',
                 host: 'github.com',
@@ -145,6 +146,26 @@ describe('installExtension', () => {
             )
             expect(manifest.description).toBe('Track goals')
             expect(manifest.requires).toEqual({ td: '>=4.0.0' })
+        })
+
+        it('installs an extension whose manifest format is newer, and says what it ignored', async () => {
+            const github = stubGitHub({
+                authoredManifest: JSON.stringify({
+                    manifestVersion: 99,
+                    description: 'from the future',
+                }),
+            })
+
+            await expect(
+                installExtension('Doist/td-goals', {}, contextFor(github.impl)),
+            ).resolves.toMatchObject({ kind: 'binary' })
+
+            expect(warnings.some((warning) => warning.includes('format newer than'))).toBe(true)
+            const manifest = JSON.parse(
+                await readFile(join(extensionsDir, 'td-goals', '.td-manifest.json'), 'utf8'),
+            )
+            // The fields this version knows are still kept.
+            expect(manifest.description).toBe('from the future')
         })
 
         it('installs without a manifest when the repository ships none', async () => {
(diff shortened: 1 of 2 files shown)


src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+
+        const executableName = asset.name.endsWith('.exe') ? `${dirName}.exe` : dirName
+        const executablePath = join(staged, executableName)
+        await writeFile(executablePath, data)
+        await chmod(executablePath, 0o755)
+
+        let authored: AuthoredManifest = {}
+        if (authoredRaw) {
+            try {
+                authored = JSON.parse(authoredRaw) as AuthoredManifest
```
doistbot: <a href="#">P2</a>  `authored` is only a type assertion. `JSON.parse` returns `null` for a valid `td-extension.json` that contains `null`, and `authored.description` then throws a raw `TypeError` instead of being treated as invalid metadata. Validate that the parsed value is a non-array object before reading its fields (or reuse the manifest field-picking logic) so `null` and primitives are ignored with the existing warning path.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+
+    if (!(await exists(join(target, dirName)))) {
+        context.warn(
+            `${target} has no executable named "${dirName}" yet. Build it before running \`${context.binName} ${toCommandName(context.binName, dirName)}\`.`,
+        )
+    }
+
+    await mkdir(context.extensionsDir, { recursive: true })
+    const destination = join(context.extensionsDir, dirName)
+    if (existing) await rm(existing.entryPath, { recursive: true, force: true })
```
doistbot: <a href="#">P2</a>  Force-replacing a binary/git extension with a local install removes the old entry but leaves its state file at `stateDir/extensions/<dirName>.json`. `discoverExtensions` still reads that file, so the new local install is reported as `pinned: true` and stale update bookkeeping survives. Call `removeState(context.stateDir, dirName)` when `existing` is present, matching what `removeExtension` does for local installs.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+    existing: Extension | undefined,
+): Promise<void> {
+    const displaced = `${destination}.replaced-${process.pid}-${Date.now()}`
+    let movedAside = false
+
+    if (existing && existing.entryPath !== destination) {
+        await rm(existing.entryPath, { recursive: true, force: true })
+    }
+
+    if (await exists(destination)) {
```
doistbot: <a href="#">P2</a>  Test the destination with `lstat` rather than `exists` here. `exists()` uses `stat`, so it returns false for a dangling local-extension symlink. Discovery still finds that link, but a forced binary/git replacement skips moving it aside and `rename(staged, destination)` then fails because the dangling link occupies the destination. This prevents recovering from a deleted local target with `--force`.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+            context.client.fetchRepoFile(
+                parsed.owner,
+                parsed.repo,
+                authoredManifestFileName(context.binName),
+                release.tag,
+            ),
+        ])
+
+        verifyChecksum(data, expected, asset.name)
+        if (!expected) {
```
doistbot: <a href="#">P2</a>  Distinguish “no checksum file was published” from “a published checksum file omits this asset.” `findExpectedChecksum` returns `undefined` in both cases, so an incomplete `checksums.txt` installs the selected asset unverified and emits the inaccurate “publishes no checksums” warning. Reject a release that publishes checksums but has no entry for the selected asset.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+            asset: asset.name,
+            sha256: sha256(data),
+            installedAt: new Date().toISOString(),
+            description: authored.description,
+            requires: authored.requires,
+            completion: authored.completion,
+        })
+
+        await moveIntoPlace(staged, destination, existing)
+        await writeState(context.stateDir, dirName, {
```
doistbot: <a href="#">P2</a>  Preserve the displaced install until state persistence succeeds, or roll back on a `writeState` failure. If the state path is unwritable, a forced install has already replaced and deleted the previous extension before this call rejects. The command then reports failure while the recoverable prior install is gone; the git path has the same ordering.

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+): Promise<InstallResult> {
+    const dirName = parsed.repo
+    const destination = join(context.extensionsDir, dirName)
+    const staged = await stagingDir(context.extensionsDir, dirName)
+    // git refuses to clone into a directory that already has entries.
+    await rm(staged, { recursive: true, force: true })
+
+    let moved = false
+    try {
+        await clone(parsed.url, staged)
```
doistbot, line 310: <a href="#">P2</a>  `installGit` passes the raw source URL to `clone`, and on failure `clone` throws `Could not clone ${url}.` (git.ts), printing the full URL. A private-repo URL with embedded credentials (`https://user:TOKEN@github.com/...`) will therefore echo the token to the terminal/logs when cloning fails (e.g., a typo in the repo path). Redact the `user:pass@` userinfo in the error path while still using the original URL for the actual `git clone`.

doistbot (commented):
Slice 5 adds extension installation from release binaries, git clones, or local directories, with staging-then-move semantics, checksum verification, and a trust warning emitted before any download or clone — the overall flow is solid and well tested. Few things worth tightening:
- Harden manifest parsing and checksum handling: validate that a parsed `td-extension.json` is a non-array object before reading fields, and distinguish "no checksum file published" from "checksum file omits the selected asset" so incomplete checksum files don't install unverified binaries under a misleading warning.
- Fix force-replace recovery gaps: a dangling local symlink defeats the `exists()` destination check, replaced installs are deleted before state persistence succeeds (leaving no rollback on write failure), and swapping a binary/git install for a local one leaves stale state files that make the new install report as pinned.
- Address the Windows-specific issues: case-insensitive filtering of protected env vars before spawning npm, and a platform-dependent test assertion around local symlink recording.
- Misc: redact embedded credentials from git clone failure messages, and replace the O(n) discovery sweep used to detect `dirName` collisions with a direct check of the destination entry.

<details>
<summary>Optional follow-up notes (13)</summary>

- **<a href="#">P3</a> src/lib/extensions/install.ts:120**: The displaced backup name `<destination>.replaced-…` lives inside `extensionsDir` and still starts with `<bin>-`, so `discoverExtensions` treats a leftover backup as an installed extension. If the process dies between the two renames, or `rm(displaced)` fails at the end, the old install reappears on the next run as an extra extension named `<name>.replaced-…`. Prefix it with a dot like the staging directory, or stage it outside `extensionsDir`.
- **<a href="#">P3</a> src/lib/extensions/install.ts:243**: `exists` is true for directories and for files without the executable bit, so a clone with a directory (or a non-executable script) named after the repo installs successfully and only fails later at dispatch with `EXTENSION_NOT_EXECUTABLE`. Check `isExecutable(join(staged, dirName))` — or at least `stat(...).isFile()` — so this surfaces during install; binary installs already `chmod` the executable.
- **<a href="#">P3</a> src/lib/extensions/install.test.ts:319**: This test title says it covers the "no matching asset → clone" fallback, but `file://${repo}` is parsed as a non-GitHub git source, so `fetchRelease`/`pickAsset` never run and the `stubGitHub({ assetName: … })` setup is unused. Rename it to describe the clone-from-file-URL behavior, or drive it through a GitHub shorthand source with git/network mocked so the fallback branch is actually exercised.
- **<a href="#">P3</a> src/lib/extensions/install.ts:311**: Unlike `moveIntoPlace`, a forced local replacement deletes the old install before creating the new link: if `symlink` (or the Windows `writeFile`) fails at lines 314-319, the previously installed extension is gone with nothing to restore, contradicting the recovery design documented at the top of this module. Creating the link at a temp path in the extensions dir and renaming it over the old entry would preserve the old install on failure.
- **<a href="#">P3</a> src/lib/extensions/install.ts:102**: `stagingDir` hand-rolls a unique directory name from `process.pid` + `Date.now()` and then does `rm` + `mkdir`. Node's `mkdtemp` (already used throughout this repo) creates a unique directory atomically and without collision risk. Use `await mkdtemp(join(extensionsDir, `.staging-${dirName}-`))` after the `mkdir` of `extensionsDir`.
- **<a href="#">P3</a> src/lib/extensions/install.ts:240**: `resolveRef(staged, 'HEAD')` duplicates `headSha(staged)`, which is already imported and called again at line 257 — both shell out to `rev-parse HEAD`. Reuse the one call: `pinnedSha = await headSha(staged)` and derive `version` from it, saving a subprocess and the redundant helper use.
- **<a href="#">P3</a> src/lib/extensions/install.ts:196**: `sha256(data)` recomputes the digest that `verifyChecksum` just computed internally. Having `verifyChecksum` return the hash it already calculated (or hashing once here and passing it in) avoids hashing the asset buffer twice.
- **<a href="#">P3</a> src/lib/extensions/install.ts:231**: `stagingDir` already returns a fresh empty directory, then this line deletes it and lets `git clone` recreate it. `git clone <url> <existing-empty-dir>` is supported, so the `rm` (and the comment above it) is unnecessary; clone directly into the staged path.
- **<a href="#">P3</a> src/lib/extensions/install.ts:123**: This guard is unreachable: `preflight` only returns an extension whose `entryPath` is `join(extensionsDir, dirName)`, which is exactly `destination` (discovery entries map 1:1 to `td-<name>` and `dirName` is validated to that shape). Dropping the branch removes a second deletion path. `installLocal` has the same redundancy: `rm(existing.entryPath)` followed immediately by `rm(destination)`.
- **<a href="#">P3</a> src/lib/extensions/install.ts:273**: The `moved` flag isn't needed: after `moveIntoPlace` succeeds, `staged` has been renamed away, so an unconditional `rm(staged, { recursive: true, force: true })` is a no-op — which is exactly what `installBinary`'s `finally` (lines 216-218) already does. Dropping the flag makes the two installers consistent and removes a line of state.
- **<a href="#">P3</a> src/lib/extensions/install.ts:176**: `writeFile` accepts a `mode` option, so the separate `chmod` is an extra syscall on every install. Pass `{ mode: 0o755 }` to the `writeFile` call at line 175 and drop the follow-up `chmod`.
- **<a href="#">P3</a> src/lib/extensions/install.test.ts:291**: `installGit`'s `--pin` branch (checkout the ref, resolve HEAD, and record the SHA in state) has no coverage. The binary-pin test only covers the release path, and neither git-clone test passes `pin`.
- **<a href="#">P3</a> src/lib/extensions/install.test.ts:326**: This test copy-pastes the ~20-line git init/add/commit sequence from `makeRepo` above it, differing only in that it writes a README instead of an executable. Give `makeRepo` a `files: Record<string, string>` option (like `writeFixtureExtension` does) and call it here, so a change to the repo-creation incantation only needs editing once.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F525&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4c5b9640-ad22-11f1-9475-80451161e3c9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4c5b9640-ad22-11f1-9475-80451161e3c9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/install.ts
```diff
@@ -0,0 +1,369 @@
...
+            `"${name}" is the name of a built-in ${context.binName} command.`,
+            {
+                hints: [
+                    `An extension cannot take that name. It would only be reachable as \`${context.binName} extension exec ${name}\`.`,
+                ],
+            },
+        )
+    }
+
+    const installed = await discoverExtensions({
```
doistbot: <a href="#">P2</a>  This enumerates and fully describes every installed extension (including manifests, state, and executable checks) solely to detect a collision for the known `dirName`. Check the single destination entry directly instead; installs with many extensions currently incur O(n) filesystem work before any download or clone.

src/lib/extensions/install.test.ts
```diff
@@ -0,0 +1,355 @@
...
+            await writeFixtureExtension(extensionsDir, 'goals')
+            const github = stubGitHub()
+
+            await expect(
+                installExtension('Doist/td-goals', {}, contextFor(github.impl)),
+            ).rejects.toMatchObject({ code: 'EXTENSION_ALREADY_INSTALLED' })
+
+            await expect(
+                installExtension('Doist/td-goals', { force: true }, contextFor(github.impl)),
+            ).resolves.toMatchObject({ kind: 'binary' })
```
doistbot, line 264: <a href="#">P2</a>  The forced replacement is only asserted via `{ kind: 'binary' }`. It never verifies the existing `td-goals` fixture was actually replaced, and there is no test for the `moveIntoPlace` recovery branch where a failed staged->destination rename restores the displaced old install. A test that forces a replace, makes that final move fail, and asserts the previous extension still exists would pin down the PR's stated "previous install moved aside" guarantee.

src/lib/extensions/install.test.ts
```diff
@@ -0,0 +1,355 @@
...
+            const workspace = join(root, 'code')
+            await mkdir(workspace, { recursive: true })
+            const target = await writeFixtureExtension(workspace, 'scratch')
+            const github = stubGitHub()
+
+            const result = await installExtension(target, {}, contextFor(github.impl))
+
+            expect(result).toMatchObject({ name: 'scratch', kind: 'local', dir: target })
+            const entry = join(extensionsDir, 'td-scratch')
+            expect((await lstat(entry)).isSymbolicLink()).toBe(true)
```
doistbot: <a href="#">P2</a>  This assertion fails on Windows, where installLocal deliberately writes a plain file recording the target instead of a symlink (see the win32 branch in install.ts, with a comment explaining symlinks need elevated rights there). Guard with `it.skipIf(process.platform === 'win32')` or branch the assertion on the platform, otherwise any contributor running `npm test` on Windows gets a false failure for behavior the code intentionally supports.

src/lib/extensions/install.ts
scottlovegrove: Done. A downloaded author manifest now goes through the same field checks as one read from disk, so null, an array or wrong types are ignored with the existing warning rather than throwing.

src/lib/extensions/install.ts
scottlovegrove: Done. A local install now clears state left by whatever was installed under that name, so a stale pin cannot make the new install report as pinned and be skipped by every upgrade.

src/lib/extensions/install.ts
scottlovegrove: Done. The destination is tested with lstat, so a local install whose target has been deleted is seen as occupying the path and can be replaced with --force.

src/lib/extensions/install.ts
scottlovegrove: Done. The checksum lookup now reports whether the release publishes checksums at all, separately from whether it names this asset, and a release that publishes them but omits the asset being installed is refused rather than installed unverified.

src/lib/extensions/install.ts
scottlovegrove: Done. The displaced install now stays put until everything that can still fail has succeeded, and is restored if recording the install fails. Both the binary and the git path do this, and there is a test that makes the state write fail and asserts the previous extension is still there.

src/lib/extensions/install.ts
scottlovegrove: Done in the tools slice: the URL is redacted before it appears in an error, and the same redaction is applied to git's own output, so a failed clone of a private repository cannot print a token.

src/lib/extensions/install.ts
scottlovegrove: Done. Collision detection now looks at the single entry that could match, through findExtension, instead of describing every installed extension before any download.

src/lib/extensions/install.test.ts
scottlovegrove: Done. The forced replacement now asserts the new binary really is on disk, and there is a separate test that makes the step after the move fail and asserts the previous extension is restored intact.

src/lib/extensions/install.test.ts
scottlovegrove: Done. That assertion is skipped on Windows, where installLocal deliberately writes a path file instead. Also added a test that a local source which is not a directory is refused, renamed the file-URL clone test to say what it actually covers, moved staging to mkdtemp, dot-prefixed the set-aside copy so a leftover one is not discovered as an extension, made the local replacement atomic, and made a clone's entry point have to be executable at install time rather than at first run.

gnapse (approved):

PR #526, opened 2026-09-10
feat(extensions): upgrade extensions

scottlovegrove:
Slice 6 of 7, on top of slice 5.

Clones fast-forward, release binaries are re-downloaded when the tag moves, and local installs are left alone because the user's own directory is already the source of truth.

Pins are honoured unless forced, and a forced upgrade clears the pin it just moved past rather than leaving a stale one to skip every later upgrade. Dependencies are reinstalled only when `package.json` or the lockfile changed. Upgrading many extensions bounds how many talk to GitHub at once, and warns about trust once for the whole run rather than once per extension.

### DIFF
--- a/src/lib/extensions/upgrade.test.ts
+++ b/src/lib/extensions/upgrade.test.ts
@@ -10,7 +10,7 @@ import type { InstallContext } from './install.js'
 import { installDependencies } from './npm.js'
 import { run } from './run.js'
 import { readState, writeState } from './state.js'
-import { mapWithConcurrency, upgradeExtension } from './upgrade.js'
+import { upgradeExtension } from './upgrade.js'
 
 // Dependency installation is a real npm run; the upgrade tests only care that
 // it is triggered by the right change, so stand in for it here.
@@ -61,36 +61,6 @@ function stubGitHub(latestTag: string) {
     }) as unknown as typeof fetch
 }
 
-describe('mapWithConcurrency', () => {
-    it('never runs more than the limit at once', async () => {
-        let active = 0
-        let peak = 0
-
-        await mapWithConcurrency(
-            Array.from({ length: 12 }, (_, i) => i),
-            4,
-            async (item) => {
-                active += 1
-                peak = Math.max(peak, active)
-                await new Promise((resolve) => setTimeout(resolve, 5))
-                active -= 1
-                return item
-            },
-        )
-
-        expect(peak).toBeLessThanOrEqual(4)
-    })
-
-    it('keeps results in the order of the input', async () => {
-        const results = await mapWithConcurrency([3, 1, 2], 2, async (item) => {
-            await new Promise((resolve) => setTimeout(resolve, item))
-            return item * 10
-        })
-
-        expect(results).toEqual([30, 10, 20])
-    })
-})
-
 describe('upgradeExtension', () => {
     let root: string
     let extensionsDir: string
@@ -137,7 +107,13 @@ describe('upgradeExtension', () => {
         const workspace = join(root, 'code')
         await mkdir(workspace, { recursive: true })
         const target = await writeFixtureExtension(workspace, 'scratch')
-        await symlink(target, join(extensionsDir, 'td-scratch'), 'dir')
+        // Matching installLocal, which writes a path file on Windows because
+        // symlinks there need rights the test runner will not have.
+        if (process.platform === 'win32') {
+            await writeFile(join(extensionsDir, 'td-scratch'), target, 'utf8')
+        } else {
+            await symlink(target, join(extensionsDir, 'td-scratch'), 'dir')
+        }
 
         const result = await upgradeExtension(
             await find('scratch'),
@@ -164,7 +140,7 @@ describe('upgradeExtension', () => {
         expect(result).toMatchObject({ outcome: 'skipped', detail: 'pinned to v1.0.0' })
     })
 
-    it('upgrades past a pin only when forced', async () => {
+    it('upgrades past a pin only when forced, and clears the pin it moved off', async () => {
         await writeFixtureExtension(extensionsDir, 'goals', {
             installedManifest: manifestFor('v1.0.0'),
         })
@@ -177,6 +153,16 @@ describe('upgradeExtension', () => {
         )
 
         expect(result).toMatchObject({ outcome: 'upgraded', from: 'v1.0.0', to: 'v2.0.0' })
+
+        // A pin left behind would make every later upgrade skip this
+        // extension, forever.
+        await expect(readState(stateDir, 'td-goals')).resolves.not.toHaveProperty('pinned')
+        const next = await upgradeExtension(
+            await find('goals'),
+            {},
+            contextFor(stubGitHub('v3.0.0')),
+        )
+        expect(next.outcome).not.toBe('skipped')
     })
 
     it('reports an up-to-date binary without downloading anything', async () => {
@@ -342,12 +328,23 @@ describe.skipIf(!HAS_GIT)('upgradeExtension for git clones', () => {
         expect(executable).toContain('echo v2')
     })
 
-    it('prints the trust warning before pulling new code', async () => {
-        await advanceOrigin()
+    it('prints the trust warning before it runs any of the new code', async () => {
+        await advanceOrigin({ 'package.json': JSON.stringify({ name: 'td-cloned' }) })
+        const order: string[] = []
+        const context = contextFor()
+        context.warn = (message) => {
+            order.push(`warn:${message}`)
+            warnings.push(message)
+        }
+        vi.mocked(installDependencies).mockImplementation(async () => {
+            order.push('install-dependencies')
+        })
 
-        await upgradeExtension(await find('cloned'), {}, contextFor())
+        await upgradeExtension(await find('cloned'), {}, context)
 
-        expect(warnings).toEqual(['trust warning'])
+        // Dependency installation runs the repository's own lifecycle scripts,
+        // so the warning has to come first, not merely be present.
+        expect(order).toEqual(['warn:trust warning', 'install-dependencies'])
     })
 
     it('reports what a dry run would do without moving the clone', async () => {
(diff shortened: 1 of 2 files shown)


src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,206 @@
...
+            outcome: 'would-upgrade',
+            from: before.slice(0, 8),
+            to: remote.slice(0, 8),
+        }
+    }
+
+    warnTrust()
+    const result = options.force ? await resetToRemote(extension.dir) : await pull(extension.dir)
+
+    if (options.force && extension.pinned) {
```
doistbot: <a href="#">P2</a>  The pin is cleared before checking whether `resetToRemote` actually succeeded. `resetToRemote` returns `{ changed: false, error }` when the fetch or reset fails, and `{ changed: false }` when remote HEAD equals HEAD, but this block still writes `pinned: undefined`. A failed forced upgrade therefore drops the pin, so a later non-forced upgrade can move the clone. Move this block below the `result.error` / `!result.changed` checks and only clear when the reset really moved the clone.

doistbot (commented):
This slice adds extension upgrades: git clones fast-forward, release binaries re-download when the tag moves, local installs are skipped, pins are honoured unless forced, dependencies reinstall only when manifests change, and concurrency/trust-warning handling is shared across the run.

Few things worth tightening:
- A forced git upgrade clears the pin before verifying the reset actually succeeded, so a failed or no-op reset drops the pin and later plain upgrades can move the clone off the pinned commit.
- If npm fails after a fast-forward, the clone is left at the new commit with stale `node_modules` and no retry will happen; consider installing in a staging clone or resetting back on failure.
- `changedFiles` parses every changed path just to check two root files — a targeted `git diff --quiet` with a pathspec would be cheaper.
- Test gaps: the local-install test uses `symlink` directly (fails on Windows; match `installLocal`'s plain-file behavior), and the forced binary-upgrade test never asserts the stale pin was cleared — add the same follow-up state/non-forced-upgrade assertions as the git test.

<details>

- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:40**: `mapWithConcurrency` duplicates the worker-pool loop in `mapFilterSections` (`src/commands/filter/view.ts:116`): both claim the next index before awaiting to cap concurrency and preserve input order. Extract one generic helper (for example in `src/lib/`) and call it from both places instead of maintaining two copies.
- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:70**: Dry run reports `up-to-date` when it could not reach the remote. `remoteHeadSha` returns `undefined` when `git ls-remote` fails (offline, auth failure, missing origin), and the `!remote` branch folds that into `up-to-date`; a real upgrade would surface that as a `skipped` result with `error`. Distinguish "could not check" from "no update", e.g. return `skipped` with an error detail when `remote` is undefined.
- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:183**: `discoverExtensions` already reads this state file to compute `extension.pinned`, so an `upgrade --all` over pinned extensions reads each one's state twice. For binary installs `extension.manifest.tag` already carries the pinned tag, so this read is fully redundant there; carry the pinned ref on `Extension` during discovery (or branch on `extension.kind`) and reuse it in the skip detail instead of re-reading.
- **<a href="#">P3</a> src/lib/extensions/upgrade.test.ts:350**: The test name promises the warning comes *before* pulling, but the assertion only checks that it was emitted. With no `package.json` change, no extension code runs after the pull, so moving `warnTrust()` after `pull` would still pass. Record order explicitly — e.g. push to a shared array from `warn` and from a mocked `installDependencies` on a package.json change — and assert the warning is first.
- **<a href="#">P3</a> src/lib/extensions/upgrade.ts:202**: `upgradeExtensions` (and its shared `createTrustWarner`) has no test, so the one-warning-per-run behavior in the PR description can regress silently: if it stopped passing `warnTrust` and used `upgradeExtension`'s default, each extension would warn and all current tests would still pass. Add a test that upgrades two discovered extensions and asserts `warn` was called exactly once.
- **<a href="#">P3</a> src/lib/extensions/upgrade.test.ts:81**: `expect(peak).toBeLessThanOrEqual(4)` also passes for a serial implementation (peak would be 1), so this test doesn't prove items actually run concurrently up to the limit. Asserting `peak` to be 4 covers both the bound and the parallelism in one line.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F526&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4e883a40-ad22-11f1-9845-0407c175dd60%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4e883a40-ad22-11f1-9845-0407c175dd60&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,206 @@
...
+        return { name: extension.name, outcome: 'skipped', detail: result.error }
+    }
+    if (!result.changed) {
+        return { name: extension.name, outcome: 'up-to-date' }
+    }
+
+    if (result.from && result.to) {
+        const changed = await changedFiles(extension.dir, result.from, result.to)
+        if (changed.some((file) => DEPENDENCY_FILES.has(file))) {
+            await installDependencies(extension.dir)
```
doistbot: <a href="#">P2</a>  `installDependencies` runs after the clone has already fast-forwarded. If npm fails (missing toolchain, network, native build error), the function throws but the clone is left at the new commit with stale or missing `node_modules`. Subsequent upgrades see HEAD already at the remote and never retry dependency installation. Install dependencies in a staging clone and swap it into place, or reset the clone back to `result.from` when npm fails.

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    if (result.error) {
+        return { name: extension.name, outcome: 'skipped', detail: result.error }
+    }
+    if (!result.changed) {
+        return { name: extension.name, outcome: 'up-to-date' }
+    }
+
+    if (result.from && result.to) {
+        const changed = await changedFiles(extension.dir, result.from, result.to)
```
doistbot: <a href="#">P2</a>  `changedFiles` collects and parses every changed path between the two commits just to detect either of two root files. For extensions with a large update, this can produce a large subprocess payload and array unnecessarily. Add a targeted git helper using `git diff --quiet <from>..<to> -- package.json package-lock.json` (or equivalent pathspec) and use its boolean result here.

src/lib/extensions/upgrade.test.ts
```diff
@@ -0,0 +1,395 @@
...
+
+    afterEach(async () => {
+        await rm(root, { recursive: true, force: true })
+    })
+
+    it('leaves a local install alone', async () => {
+        const workspace = join(root, 'code')
+        await mkdir(workspace, { recursive: true })
+        const target = await writeFixtureExtension(workspace, 'scratch')
+        await symlink(target, join(extensionsDir, 'td-scratch'), 'dir')
```
doistbot: <a href="#">P2</a>  This test calls `symlink` directly, which needs elevated rights (or Developer Mode) on Windows. The production code deliberately avoids symlinks on win32 and writes a plain target file instead (`installLocal` in install.ts), so this test will fail on Windows before it can exercise the local-install skip. Branch on `process.platform === 'win32'` and write the target path to `join(extensionsDir, 'td-scratch')`, matching `installLocal`.

src/lib/extensions/upgrade.test.ts
```diff
@@ -0,0 +1,395 @@
...
+        })
+        await writeState(stateDir, 'td-goals', { pinned: 'v1.0.0' })
+
+        const result = await upgradeExtension(
+            await find('goals'),
+            { force: true },
+            contextFor(stubGitHub('v2.0.0')),
+        )
+
+        expect(result).toMatchObject({ outcome: 'upgraded', from: 'v1.0.0', to: 'v2.0.0' })
```
doistbot, line 155: <a href="#">P2</a>  This only proves the forced upgrade succeeded; it never checks that the stale pin was cleared. `upgradeBinary` relies on `installBinary` writing `{ pinned: undefined }`, and if that state write regresses this test still passes while every later non-forced upgrade would be skipped. Add the same follow-up assertions as the git test: `readState(stateDir, 'td-goals')` has no `pinned`, and a subsequent non-forced upgrade is not `'skipped'`.

src/lib/extensions/upgrade.ts
scottlovegrove: Done. The pin is cleared only after the reset has been shown to have moved the clone, so a failed or no-op forced upgrade no longer unpins the extension.

src/lib/extensions/upgrade.ts
scottlovegrove: Done. If installing dependencies fails, the clone is reset back to the commit it came from, so a later upgrade sees work still to do rather than finding it already at the remote and never retrying.

src/lib/extensions/upgrade.ts
scottlovegrove: Done. There is a pathsChanged helper now using git diff --quiet with a pathspec, and it reports when git could not answer. "Could not tell" reinstalls, since a slow upgrade costs less than an extension whose dependencies no longer match its code.

src/lib/extensions/upgrade.test.ts
scottlovegrove: Done. The test writes a path file on Windows and a symlink elsewhere, matching what installLocal does.

src/lib/extensions/upgrade.test.ts
scottlovegrove: Done. That test now asserts the state file no longer carries a pin, and that the following non-forced upgrade is not skipped. The trust-warning test also records the order of the warning and of dependency installation, so it fails if the warning ever moves after the code it warns about, and the dry run now says it could not reach the remote instead of reporting up to date.

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,204 @@
...
+        // "Could not tell" reinstalls: a slow upgrade costs less than an
+        // extension left with dependencies that no longer match its code.
+        if (changed !== false) {
+            try {
+                await installDependencies(extension.dir)
+            } catch (error) {
+                // The clone has already moved, and a later upgrade would find
+                // it up to date and never retry, so put it back where its
+                // dependencies still match its code.
+                await run('git', ['reset', '--quiet', '--hard', result.from], {
```
gnapse, line 96: A failed upgrade can leave the previously working extension broken. For example:

1. Version 1 works, and the user upgrades to version 2.
2. npm starts replacing the libraries the extension needs, but installation fails.
3. This reset restores version 1's source code, but not its libraries.

The user now has a failed upgrade and an old extension that may no longer run. I reproduced this locally: HEAD returned to the old commit, but a previous dependency file was gone. The existing test mocks npm and checks only HEAD, so it misses this.

Could we prepare the new version and its dependencies in a staging directory, then replace the working installation only after success? Also, when this is a forced upgrade of a pinned extension, the pin cleared above stays cleared after rollback. That should only be cleared once the full upgrade succeeds.

gnapse (commented):
Two upgrade cases found while testing extensions locally:

src/lib/extensions/upgrade.ts
```diff
@@ -0,0 +1,204 @@
...
+        return {
+            name: extension.name,
+            outcome: 'would-upgrade',
+            from: before.slice(0, 8),
+            to: remote.slice(0, 8),
+        }
+    }
+
+    warnTrust()
+    const result = options.force ? await resetToRemote(extension.dir) : await pull(extension.dir)
```
gnapse, line 69: Forcing an upgrade past a pinned tag leaves subsequent ordinary upgrades stuck. I reproduced this through the CLI:

1. Install an extension with `--pin v1`.
2. Run `extension upgrade <name> --force`: it successfully moves to the latest code and removes the pin.
3. Run `extension upgrade <name>` again: it is skipped with "You are not currently on a branch."

Checking out the tag puts Git in detached HEAD mode. `resetToRemote()` moves the commit but leaves Git in that mode, so the next `pull` does not know which branch to follow.

The extension still runs, but normal upgrades stop working. Could forcing past a pin also restore tracking of the intended remote branch? A test that installs an actual tag, forces an upgrade, then performs an ordinary upgrade would cover this.

PR #527, opened 2026-09-10
feat(extensions): the extension manager facade

scottlovegrove:
Slice 7 of 7, on top of slice 6. This completes the library half of phase 1.

Ties the slices together behind `createExtensionManager`, which is the only thing a host CLI needs to call.

Includes the version-range check behind an extension's `requires` field. A range the running CLI does not satisfy is a warning rather than a refusal: upgrading the CLI should not silently break an extension that still works.

Also records the subsystem in `CODEBASE.md`, including why nothing in the directory imports from the rest of the repo.

Next, in a separate PR: the `td extension` command group, the `src/index.ts` wiring, the command-token lookup change, `--user` scoping, doctor checks and `SKILL_CONTENT`.

### DIFF
--- a/src/lib/extensions/manager.ts
+++ b/src/lib/extensions/manager.ts
@@ -10,7 +10,8 @@
 
 import { join } from 'node:path'
 import { CliError } from '@doist/cli-core'
-import { discoverExtensions } from './discover.js'
+import { mapWithConcurrency } from './concurrency.js'
+import { discoverExtensions, findExtension } from './discover.js'
 import { buildExtensionEnv, dispatchExtension } from './dispatch.js'
 import { isExecutable } from './fs-utils.js'
 import { headSha } from './git.js'
@@ -31,7 +32,7 @@ import type {
     UpgradeOptions,
     UpgradeResult,
 } from './types.js'
-import { mapWithConcurrency, upgradeExtensions } from './upgrade.js'
+import { upgradeExtensions } from './upgrade.js'
 import { satisfiesRange } from './version-range.js'
 
 export type ExtensionManager = {
@@ -111,20 +112,21 @@ export function createExtensionManager(options: ExtensionManagerOptions): Extens
         return discoverExtensions(discoverOptions)
     }
 
+    function notFound(name: string): CliError {
+        return new CliError('EXTENSION_NOT_FOUND', `No extension named "${name}" is installed.`, {
+            hints: [`Run \`${binName} extension list\` to see what is installed.`],
+        })
+    }
+
+    /** Inspect the one entry that could match, rather than describing them all. */
     async function find(selector: string): Promise<Extension | undefined> {
-        const name = normalizeSelector(binName, selector)
-        return (await discover()).find((extension) => extension.name === name)
+        return findExtension(normalizeSelector(binName, selector), discoverOptions)
     }
 
     async function requireExtension(selector: string): Promise<Extension> {
-        const extension = await find(selector)
-        if (!extension) {
-            throw new CliError(
-                'EXTENSION_NOT_FOUND',
-                `No extension named "${normalizeSelector(binName, selector)}" is installed.`,
-                { hints: [`Run \`${binName} extension list\` to see what is installed.`] },
-            )
-        }
+        const name = normalizeSelector(binName, selector)
+        const extension = await find(name)
+        if (!extension) throw notFound(name)
         return extension
     }
 
@@ -178,21 +180,21 @@ export function createExtensionManager(options: ExtensionManagerOptions): Extens
         },
 
         async upgrade(selectors, upgradeOptions = {}) {
-            // One scan for the whole set: resolving each selector separately
-            // would re-read every installed extension once per name.
+            // Deduplicated first: `goals` and `td-goals` name one extension,
+            // and two upgrades of one directory at once would write over each
+            // other.
+            const names = [
+                ...new Set(selectors.map((selector) => normalizeSelector(binName, selector))),
+            ]
+
+            // One scan for the whole set, rather than one per name.
             const installed = await discover()
-            const extensions = selectors.map((selector) => {
-                const name = normalizeSelector(binName, selector)
+            const extensions = names.map((name) => {
                 const extension = installed.find((candidate) => candidate.name === name)
-                if (!extension) {
-                    throw new CliError(
-                        'EXTENSION_NOT_FOUND',
-                        `No extension named "${name}" is installed.`,
-                        { hints: [`Run \`${binName} extension list\` to see what is installed.`] },
-                    )
-                }
+                if (!extension) throw notFound(name)
                 return extension
             })
+
             return upgradeExtensions(extensions, upgradeOptions, installContext)
         },
 
@@ -217,7 +219,7 @@ export function createExtensionManager(options: ExtensionManagerOptions): Extens
                 extension,
                 user: dispatchOptions.user,
                 accessible: isAccessible(),
-                extra: dispatchOptions.env,
+                env: dispatchOptions.env,
             })
 
             return dispatchExtension(extension, args, env)
--- a/src/lib/extensions/manager.ts
+++ b/src/lib/extensions/manager.ts
@@ -1,0 +1,230 @@
+/**
+ * The extension manager: one object holding every operation the CLI needs,
+ * configured entirely through its options.
+ *
+ * Nothing in this directory knows which CLI it is serving. The binary name,
+ * directories, version, reserved command names and first-party source all
+ * arrive here, which is what allows the whole thing to move to
+ * `@doist/cli-core` and be adopted by another CLI with one call.
+ */
+
+import { join } from 'node:path'
+import { CliError } from '@doist/cli-core'
+import { discoverExtensions } from './discover.js'
+import { buildExtensionEnv, dispatchExtension } from './dispatch.js'
+import { isExecutable } from './fs-utils.js'
+import { headSha } from './git.js'
+import { createGitHubClient } from './github.js'
+import { installExtension, type InstallContext, isEmpty } from './install.js'
+import { removeExtension } from './remove.js'
+import { normalizeSelector } from './source.js'
+import type {
+    DispatchOptions,
+    Extension,
+    ExtensionListing,
+    ExtensionManagerOptions,
+    ExtensionTheme,
+    InstallOptions,
+    InstallResult,
+    RemoveOptions,
+    RemoveResult,
+    UpgradeOptions,
+    UpgradeResult,
+} from './types.js'
+import { mapWithConcurrency, upgradeExtensions } from './upgrade.js'
+import { satisfiesRange } from './version-range.js'
+
+export type ExtensionManager = {
+    readonly binName: string
+    readonly envPrefix: string
+    readonly extensionsDir: string
+    readonly officialLabel: string
+    readonly theme: ExtensionTheme
+    readonly trustWarning: string
+    isAccessible(): boolean
+    /** Everything installed, cheaply: no subprocesses, no network. */
+    discover(): Promise<Extension[]>
+    /** Resolve `goals`, `td-goals` or `owner/td-goals` to an installed extension. */
+    find(selector: string): Promise<Extension | undefined>
+    /** Like `find`, but throws the standard not-found error. */
+    require(selector: string): Promise<Extension>
+    /** Everything installed, with the details that cost extra work to learn. */
+    list(): Promise<ExtensionListing[]>
+    install(source: string, options?: InstallOptions): Promise<InstallResult>
+    upgrade(selectors: string[], options?: UpgradeOptions): Promise<UpgradeResult[]>
+    upgradeAll(options?: UpgradeOptions): Promise<UpgradeResult[]>
+    remove(selector: string, options?: RemoveOptions): Promise<RemoveResult>
+    /** Run an extension and resolve with the exit code the host should use. */
+    dispatch(selector: string, args: string[], options?: DispatchOptions): Promise<number>
+    /** True when no extension is installed, used to decide whether to advertise the feature. */
+    isEmpty(): Promise<boolean>
+}
+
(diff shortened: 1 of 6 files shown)


src/lib/extensions/manager.ts
```diff
@@ -0,0 +1,230 @@
...
+
+        async install(source, installOptions = {}) {
+            return installExtension(source, installOptions, installContext)
+        },
+
+        async upgrade(selectors, upgradeOptions = {}) {
+            // One scan for the whole set: resolving each selector separately
+            // would re-read every installed extension once per name.
+            const installed = await discover()
+            const extensions = selectors.map((selector) => {
```
doistbot: <a href="#">P2</a>  `upgrade()` maps selectors straight into `extensions` without deduplicating after `normalizeSelector`, so `['goals', 'td-goals']` (or a repeated selector) passes the same `Extension` twice to `upgradeExtensions`. With concurrency 4, that starts two upgrades on the same directory at once — two `git pull --ff-only`s or two `installBinary`/`moveIntoPlace` calls — which can race and corrupt the install. Dedupe by normalized name before calling `upgradeExtensions`.

src/lib/extensions/version-range.ts
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * True when `version` satisfies `range`, or when the range is empty or in a
+ * form this checker does not understand.
+ */
+export function satisfiesRange(version: string, range: string | undefined): boolean {
+    const trimmed = range?.trim()
+    if (!trimmed || trimmed === '*' || trimmed === 'x') return true
+
+    return trimmed.split('||').some((clause) => {
+        const comparators = clause.trim().split(/\s+/).filter(Boolean).map(parseComparator)
```
doistbot: <a href="#">P2</a>  Preserve an operator separated from its version by whitespace. A valid semver range such as `>= 4.0.0` is split into `>=` and `4.0.0`; the first token is ignored and the second becomes an exact-match comparator. Consequently, a host at `5.3.1` incorrectly warns that it does not satisfy `>= 4.0.0`. Parse each comparator with optional whitespace after the operator (and add coverage for this form).

src/lib/extensions/manager.ts
```diff
@@ -0,0 +1,230 @@
...
+        trustWarning,
+    }
+
+    async function discover(): Promise<Extension[]> {
+        return discoverExtensions(discoverOptions)
+    }
+
+    async function find(selector: string): Promise<Extension | undefined> {
+        const name = normalizeSelector(binName, selector)
+        return (await discover()).find((extension) => extension.name === name)
```
doistbot: <a href="#">P2</a>  `find()` discovers and describes every installed extension before selecting one. As a result, each `dispatch`, `remove`, and targeted `upgrade` reads manifests/state (and git remotes for git installs) for all extensions, despite the normalized selector already identifying the only directory needed. Add a targeted lookup/describe path for this method so these per-extension operations only inspect the selected entry.

doistbot (commented):
This PR completes the library half of phase 1 by tying the extension slices together behind `createExtensionManager`, adding the fail-open `requires` version-range check, and documenting the subsystem in CODEBASE.md. Few things worth tightening:
- `upgrade()` passes selectors straight through without deduping after `normalizeSelector`, so aliased selectors (e.g. `['goals', 'td-goals']`) can trigger two concurrent upgrades against the same directory — dedupe by normalized name first.
- The range parser splits comparators on whitespace, so a valid range like `>= 4.0.0` misreads `4.0.0` as an exact-match comparator; parse operators with optional trailing whitespace and add a test for this form.
- `find()` discovers and describes every installed extension before picking one, making `dispatch`, `remove`, and targeted `upgrade` needlessly inspect all extensions; add a targeted lookup/describe path.

<details>

- **<a href="#">P3</a> src/lib/extensions/version-range.ts:73**: An empty `||` alternative is treated as a satisfied clause (`comparators.length === 0` returns true), so `'>=9.0.0 ||'` (or a leading/consecutive `||`) makes the whole range pass and suppresses the warning even though the parseable part is unmet. Drop empty clauses before `some`, returning true only when no non-empty clause remains.
- **<a href="#">P3</a> src/lib/extensions/version-range.ts:72**: A hyphen range like `4.0.0 - 5.0.0` is half-parsed rather than passed through: the split yields comparators `{='4.0.0'}`, undefined (for `-`), and `{='5.0.0'}`, so `td 4.5.0` fails the range even though it sits inside it — producing exactly the spurious warning this module's header says false negatives should not cause, since an unparseable token makes the clause pass but parseable ones still enforce equality. Either reject clauses containing an unparseable comparator wholesale (treat the clause as satisfied), or explicitly document hyphen ranges as unsupported; the current middle ground silently misjudges them.
- **<a href="#">P3</a> src/lib/extensions/manager.ts:188**: The `EXTENSION_NOT_FOUND` `CliError` is built twice in this file (here and in `requireExtension` above) with the same code, message, and hint. Extract one `notFound(name)` helper and call it from both paths so the user-facing message and hint cannot drift between the single-selector and batch paths.
- **<a href="#">P3</a> src/lib/extensions/version-range.test.ts:24**: The caret test says it covers the 0.x special case, but it only exercises `^0.2.3` (the `minor > 0` branch). The `^0.0.x` branch in `upperBound` (`return 0.0.${patch + 1}`) is never hit, so a regression there would pass. Add `satisfiesRange('0.0.3', '^0.0.3')` as true and `satisfiesRange('0.0.4', '^0.0.3')` as false.
- **<a href="#">P3</a> src/lib/extensions/manager.test.ts:72**: The test title promises the error points at the list command, but the assertion only checks `code: 'EXTENSION_NOT_FOUND'`. The hint (`Run \`td extension list\``) is the user-visible behavior being named, and it can be dropped or mistyped without this test failing. Assert the message or `hints` contains `extension list` too.
- **<a href="#">P3</a> src/lib/extensions/manager.test.ts:149**: `manager.upgrade` has its own `EXTENSION_NOT_FOUND` branch for unknown selectors (it does not delegate to `require`), and nothing in this suite exercises it. A regression there would break the future `td extension upgrade <name>` error path with no failing test. Add a case asserting `makeManager().upgrade(['ghost'])` rejects with `code: 'EXTENSION_NOT_FOUND'` and a list-command hint.
- **<a href="#">P3</a> src/lib/extensions/manager.test.ts:46**: Low-value test: this only asserts `join(dataDir, 'extensions')`, a constant. The `isEmpty` test below already verifies behaviorally that extensions are discovered from `<dataDir>/extensions` (it writes a fixture there and watches `isEmpty()` flip). I'd delete this test and let the behavioral one carry the contract.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F527&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A50a6fb90-ad22-11f1-9e1e-4d25ef69f3d2%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A50a6fb90-ad22-11f1-9e1e-4d25ef69f3d2&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/manager.ts
scottlovegrove: Done. Selectors are deduplicated by normalized name before upgrading, so naming one extension twice cannot start two upgrades of the same directory at once. Covered by a test.

src/lib/extensions/version-range.ts
scottlovegrove: Done. An operator written apart from its version is joined before the clause is split, so ">= 4.0.0" is one comparator. Also fixed two neighbours: an empty alternative no longer satisfies a range, and a clause that cannot be read in full is treated as satisfied rather than half-enforced, which is what made a hyphen range warn about a version inside it.

src/lib/extensions/manager.ts
scottlovegrove: Done. find now goes through findExtension, which describes only the entry asked for, so dispatch, remove and a named upgrade no longer read every installed extension's manifests and git remote. The not-found error is built in one place, and there are tests for the hint and for the upgrade path's own not-found branch.

pawelgrimm:
@scottlovegrove Sorry, I ran out of time today. I'll get a review in tomorrow!

engfragui:
💭 For some reason, when I asked an agent to help me review https://github.com/Doist/todoist-cli/pull/521, they went ahead and did a review pass on the entire extensions feature/stack. The main issue they flagged after doing that was on this specific PR for `satisfiesRange`. I'll leave the feedback here for your consideration.

src/lib/extensions/version-range.ts
```diff
@@ -0,0 +1,94 @@
...
+        if (minor > 0) return `0.${minor + 1}.0`
+        return `0.0.${patch + 1}`
+    }
+    if (operator === '~') return `${major}.${minor + 1}.0`
+    return undefined
+}
+
+function satisfiesComparator(version: string, comparator: Comparator): boolean {
+    const { operator, version: target } = comparator
+    const compared = compareVersions(version, target)
```
engfragui, line 42: **`satisfiesRange` throws on common range syntaxes, making an extension unrunnable**

The module's doc comment promises that anything it can't parse is *treated as satisfied* (a range is advisory — a false negative is just noise). But that guarantee doesn't hold. The `COMPARATOR` regex accepts a version starting with any digit (`v?(\d.*)`), so it builds a **defined** comparator for inputs like `>=1.2` — which means the "unparseable ⇒ satisfied" guard never fires, and `compareVersions`/`parseVersion` (from `@doist/cli-core`, which throws on anything that isn't a full `x.y.z`) throws instead. That exception propagates straight out of `satisfiesRange` — there's no try/catch here.

Reproduced against the real `@doist/cli-core@1.4.0`:

```
satisfiesRange("5.3.2", ">=1.2")     THREW: Invalid version string: '1.2'
satisfiesRange("5.3.2", "^1.2")      THREW: Invalid version string: '1.2'
satisfiesRange("5.3.2", "~2")        THREW: Invalid version string: '2'
satisfiesRange("5.3.2", ">=1.2.x")   THREW: Invalid version string: '1.2.x'
satisfiesRange("5.3.2", "1.x")       THREW: Invalid version string: '1.x'
satisfiesRange("5.3.2", ">=1.2.3.4") => true   // separate quirk: silently truncated to 1.2.3
```

Two-part versions (`>=1.2`) and `.x` wildcards are among the most common ways people write a range. Since `warnIfIncompatible` runs on **every dispatch** (`manager.ts`, before the extension is spawned) *and* in `doctor`, an extension author who writes `"requires": { "td": ">=1.2" }` in their manifest makes their extension throw on every single `td <ext>` invocation. It's honest-but-imprecise manifest input turning into a hard failure — exactly the outcome the "advisory" design says it wants to avoid.

**Fix options:**
- Wrap the `parseVersion`/`compareVersions` calls in try/catch and treat a throw as satisfied (matches the module's stated philosophy); or
- Tighten `COMPARATOR` to require a full `x.y.z` so these fall through to the existing "cannot read ⇒ satisfied" branch.

The `>=1.2.3.4 ⇒ true` case is a separate, lower-severity quirk: it's silently truncated to `1.2.3` rather than rejected, so a 4-part range compares wrong instead of being treated as unparseable.

_(Post-merge review of #527 — worth fixing before this reaches `main`, since it's confirmed and trivially triggered.)_

scottlovegrove: Confirmed, fixed in #541, and it is worse than the report suggests — thank you, this is the most serious thing anyone found across the whole feature.

It is not only dispatch. `checkExtensions` calls the same code, so the exception escapes `doctor` too:

```
$ td wild                  → Invalid version string: '1.x'    exit 1
$ td doctor --offline      → Invalid version string: '1.x'    exit 1   (no checks, no summary)
```

One extension with a malformed range takes `doctor` down for the whole CLI, not just for that extension.

Went with your first option — evaluating a range is now wrapped, and anything `parseVersion` refuses counts as satisfied, which is what the module said all along. Tightening the comparator pattern instead would have lost correct warnings: `^1.2` genuinely is unsatisfied for td 5.3.2, and under that approach it would have become "unparseable, so satisfied".

One correction to the repro table, in case it matters for whatever produced it. Against `@doist/cli-core@1.4.0` as pinned here:

```
>=1.2       => true     (reported as throwing)
^1.2        => false    (reported as throwing)
~2          => false    (reported as throwing)
>=1.2.x     THREW  ✓
1.x         THREW  ✓
>=1.2.3.4   THREW      (reported as silently truncating to 1.2.3)
```

That version tolerates a two-part version but not a wildcard or a fourth part. Your conclusion holds either way; only the inputs differ. After the fix all six return `true`, and the two-part forms keep evaluating properly.

scottlovegrove:
That whole-stack pass paid off — the `satisfiesRange` finding is the most serious thing anyone caught across the feature, and it was not hypothetical: an extension declaring `"requires": { "td": "1.x" }` could not run at all, and it took `td doctor` down with it for anyone who had it installed.

Fixed in #541, with the detail and a correction to the repro table on the inline thread.

gnapse (approved):

PR #531, opened 2026-09-11
feat(paths): resolve the per-user data and state directories

scottlovegrove:
cli-core owns `getConfigPath` but has no equivalent for the data and state directories, which the extension system needs: installed extensions live under the data directory, and their pins and update bookkeeping under the state directory.

Resolve both in the same shape cli-core uses for the config path — the XDG variable first, the platform default second — with Windows falling back to `%LOCALAPPDATA%`. macOS gets the XDG layout rather than `~/Library/Application Support`, because cli-core already puts the config at `~/.config` there and splitting the two would leave a user's files in two unrelated places.

`getConfigDir` joins `getConfigPath` in config.ts, derived from the path rather than rebuilt from the same environment variables so the two can never disagree.

### DIFF
--- a/src/lib/paths.test.ts
+++ b/src/lib/paths.test.ts
@@ -1,0 +1,113 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+
+const REAL_PLATFORM = process.platform
+const HOME = '/home/tester'
+
+/**
+ * Load the module with `homedir()` pointed somewhere predictable. `paths.ts`
+ * imports `node:os` directly, so mocking the module reaches it — unlike the
+ * config path, which is resolved inside cli-core's compiled output.
+ */
+async function loadPaths(home = HOME) {
+    vi.resetModules()
+    vi.doMock('node:os', () => ({ homedir: () => home }))
+    return import('./paths.js')
+}
+
+function setPlatform(platform: NodeJS.Platform) {
+    Object.defineProperty(process, 'platform', { value: platform, configurable: true })
+}
+
+/** The XDG variables are read from the real environment, so clear them first. */
+function clearEnv() {
+    for (const name of ['XDG_DATA_HOME', 'XDG_STATE_HOME', 'LOCALAPPDATA']) {
+        vi.stubEnv(name, undefined)
+    }
+}
+
+afterEach(() => {
+    setPlatform(REAL_PLATFORM)
+    vi.unstubAllEnvs()
+    vi.doUnmock('node:os')
+    vi.resetModules()
+})
+
+describe('getDataDir', () => {
+    it.each(['linux', 'darwin'] as const)('uses the XDG layout on %s', async (platform) => {
+        clearEnv()
+        setPlatform(platform)
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('/home/tester/.local/share/todoist-cli')
+    })
+
+    it('prefers XDG_DATA_HOME when it is set', async () => {
+        clearEnv()
+        setPlatform('linux')
+        vi.stubEnv('XDG_DATA_HOME', '/elsewhere/data')
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('/elsewhere/data/todoist-cli')
+    })
+
+    it('uses LOCALAPPDATA on Windows', async () => {
+        clearEnv()
+        setPlatform('win32')
+        vi.stubEnv('LOCALAPPDATA', 'C:\\Users\\tester\\AppData\\Local')
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('C:\\Users\\tester\\AppData\\Local/todoist-cli')
+    })
+
+    it('falls back to the documented location when LOCALAPPDATA is missing', async () => {
+        clearEnv()
+        setPlatform('win32')
+        const { getDataDir } = await loadPaths('C:\\Users\\tester')
+        expect(getDataDir()).toBe('C:\\Users\\tester/AppData/Local/todoist-cli')
+    })
+
+    it('lets XDG_DATA_HOME win on Windows too', async () => {
+        clearEnv()
+        setPlatform('win32')
+        vi.stubEnv('LOCALAPPDATA', 'C:\\Users\\tester\\AppData\\Local')
+        vi.stubEnv('XDG_DATA_HOME', '/elsewhere/data')
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('/elsewhere/data/todoist-cli')
+    })
+})
+
+describe('getStateDir', () => {
+    it.each(['linux', 'darwin'] as const)('uses the XDG layout on %s', async (platform) => {
+        clearEnv()
+        setPlatform(platform)
+        const { getStateDir } = await loadPaths()
+        expect(getStateDir()).toBe('/home/tester/.local/state/todoist-cli')
+    })
+
+    it('prefers XDG_STATE_HOME when it is set', async () => {
+        clearEnv()
+        setPlatform('linux')
+        vi.stubEnv('XDG_STATE_HOME', '/elsewhere/state')
+        const { getStateDir } = await loadPaths()
+        expect(getStateDir()).toBe('/elsewhere/state/todoist-cli')
+    })
+
+    it('nests under the data root on Windows, which has no state root', async () => {
+        clearEnv()
+        setPlatform('win32')
+        vi.stubEnv('LOCALAPPDATA', 'C:\\Users\\tester\\AppData\\Local')
+        const { getDataDir, getStateDir } = await loadPaths()
+        expect(getStateDir()).toBe('C:\\Users\\tester\\AppData\\Local/todoist-cli/state')
+        // The extensions directory is a sibling of `state`, so the two cannot
+        // collide however the data root is spelled.
+        expect(getStateDir().startsWith(`${getDataDir()}/`)).toBe(true)
+    })
+})
+
+describe('the two directories', () => {
+    it('stay distinct on every platform', async () => {
+        for (const platform of ['linux', 'darwin', 'win32'] as const) {
+            clearEnv()
+            setPlatform(platform)
+            const { getDataDir, getStateDir } = await loadPaths()
+            expect(getDataDir()).not.toBe(getStateDir())
+        }
+    })
+})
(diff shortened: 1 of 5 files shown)


src/lib/paths.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('/elsewhere/data/todoist-cli')
+    })
+
+    it('uses LOCALAPPDATA on Windows', async () => {
+        clearEnv()
+        setPlatform('win32')
+        vi.stubEnv('LOCALAPPDATA', 'C:\\Users\\tester\\AppData\\Local')
+        const { getDataDir } = await loadPaths()
+        expect(getDataDir()).toBe('C:\\Users\\tester\\AppData\\Local/todoist-cli')
```
doistbot: <a href="#">P2</a>  The Windows tests hardcode POSIX-joined separators (`...AppData\\Local/todoist-cli`), but `join` in paths.ts comes from the host's `node:path`, so running `npm test` on a Windows machine produces backslash-joined paths (`...AppData\\Local\\todoist-cli`) and these expectations fail. CI only runs on ubuntu-latest so it won't catch this, but Windows contributors will. Build the expected values with `path.join` (or compare with `path.sep`) so the assertions hold on any host.

src/lib/paths.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Where installed extensions live, under an `extensions` subdirectory.
+ *
+ * macOS gets the XDG layout rather than `~/Library/Application Support`,
+ * because cli-core already puts the config at `~/.config` there and splitting
+ * the two would leave a user's files in two unrelated places.
+ */
+export function getDataDir(): string {
+    if (process.env.XDG_DATA_HOME) return join(process.env.XDG_DATA_HOME, APP_NAME)
```
doistbot: <a href="#">P2</a>  Ignore relative `XDG_DATA_HOME` values rather than joining against them. The XDG Base Directory spec requires these values to be absolute; with `XDG_DATA_HOME=cache`, this returns a directory relative to whichever working directory launches `td`, so installed extensions appear to move or become undiscoverable. Validate with `isAbsolute()` before using this value, and apply the same check to `XDG_STATE_HOME` below.

src/lib/paths.test.ts
doistbot: <a href="#">P2</a>  These Windows assertions hard-code `/` separators, but `node:path` uses `\` when the test suite itself runs on Windows; changing `process.platform` does not change that. The Windows cases (including the `startsWith(`${getDataDir()}/`)` check below) will fail there. Build expected paths with `join()` or use the platform separator.

doistbot: <a href="#">P2</a>  These assertions mix separators (`C:\\Users\\tester\\AppData\\Local/todoist-cli`) because the test fakes `process.platform` but `paths.ts` uses the real `node:path.join`, whose separator depends on the host OS. On an actual Windows machine (dev checkout, Windows CI) `join` emits backslashes and these tests fail. Either build expected values with `path.posix.join`/`path.win32.join` to match the host, or assert platform-independent structure (e.g. ends with the expected segments) instead of exact strings. Same issue at lines 63, 97, and 100.

doistbot (commented):
This PR adds per-user data and state directory resolution (XDG-first with platform defaults, Windows via %LOCALAPPDATA%, macOS on the XDG layout) plus a config-dir helper derived from the resolved path — matching cli-core's documented shape. Few things worth tightening:

- The Windows test expectations hardcode POSIX separators (e.g. `AppData\\Local/todoist-cli`), but `node:path.join` uses `\` on a Windows host — since CI only runs on Linux this passes today, but Windows contributors would see failures. Build expected values with `path.join`/`path.sep` (or `path.win32.join`) so they hold on any host.
- Relative `XDG_DATA_HOME`/`XDG_STATE_HOME` values are currently joined against as-is, which violates the XDG spec's absolute-path requirement and makes installed extensions appear to move depending on the launch directory. Validate with `isAbsolute()` and fall through to the platform default otherwise.
- `APP_NAME` is duplicated across config.ts and paths.ts — consider sharing one constant.
- CODEBASE.md wasn't updated for the new `src/lib` helper module, which AGENTS.md asks for when a reusable helper is added.

<details>

- **<a href="#">P3</a> src/lib/paths.ts:35**: This adds the first host-level data/state path helpers under `src/lib/`, but `CODEBASE.md`'s `src/lib/` catalog still has no `paths.ts` entry. AGENTS.md says to update CODEBASE.md when a new broadly-reusable helper lands in `src/lib/`; add `paths.ts` (and mention `getConfigDir` in the existing `config.ts` entry) so the map stays accurate.
- **<a href="#">P3</a> src/lib/paths.ts:17**: `APP_NAME = 'todoist-cli'` is now defined independently in both config.ts and paths.ts. If one ever changes, the config and data/state directories silently drift apart. Export it from config.ts (or a small shared module) and import it here.
- **<a href="#">P3</a> src/lib/paths.test.ts:56**: These Windows assertions hardcode `/` separators, which only match the POSIX `join` used when the suite runs on Linux/macOS. On Windows, `node:path.join` emits backslashes, so lines 56, 63, 97 and the `startsWith(`${getDataDir()}/`)` check on line 100 fail there. Build the expected paths with `join(...)` from `node:path` (the pattern already used in `extensions/dispatch.test.ts`) instead of literal strings.
- **<a href="#">P3</a> src/lib/paths.test.ts:8**: This comment says mocking `node:os` "reaches" paths.ts "unlike the config path, which is resolved inside cli-core's compiled output." That contradicts this repo's test setup: `vitest.config.ts` lists `@doist/cli-core` in `server.deps.inline` specifically so `vi.doMock('node:fs/promises', …)` / `vi.doMock('node:os', …)` reach cli-core's compiled imports (see the comment there and CODEBASE.md). The mock approach works either way, but the stated justification is wrong and could steer a future reader away from a technique that does work. Reword to say the direct import simply makes the mock setup simpler, or drop the contrast with cli-core.
- **<a href="#">P3</a> src/lib/paths.ts:17**: `APP_NAME = 'todoist-cli'` is now defined in both `config.ts` and `paths.ts`. If one copy is ever renamed, the config directory and the data/state directories silently diverge, orphaning a user's installed extensions. Export the constant from `config.ts` (or a tiny shared module) and import it here so there's one definition.
- **<a href="#">P3</a> src/lib/paths.test.ts:56**: These Windows cases only mock `process.platform`; `paths.ts` still uses the runner's `node:path`, so on the Linux CI `join` produces POSIX `/` and the assertions (also lines 63, 97, and 100) bake in mixed separators like `Local/todoist-cli`. That shape never occurs on real Windows (which would use `\`), so the tests don't validate the Windows path behavior and would fail if run on a Windows host. Mock `node:path` with `path.win32` for these cases, or assert relative structure rather than hard-coded separators.
- **<a href="#">P3</a> src/lib/config.test.ts:81**: This assertion is tautological with the test above it: `getConfigPath` is a fixed mock and `getConfigDir()` is literally `dirname(getConfigPath())`, so `getConfigPath().startsWith(`${getConfigDir()}/`)` cannot fail independently of the first test. Drop it; the first test already pins the directory relationship.
- **<a href="#">P3</a> src/lib/config.test.ts:80**: This test restates the one above it. With `getConfigPath` mocked to a fixed value, `getConfigDir` has no other input, so both tests assert the same dir-of-path relation. Delete this test; the first already pins the behavior.
- **<a href="#">P3</a> src/lib/paths.ts:35**: Repo-standard nit: AGENTS.md/CODEBASE.md call for a CODEBASE.md update when a new broadly-reusable helper lands in `src/lib/` (the "don't reimplement" catalog). `getDataDir`/`getStateDir` here qualify — the extension system consumes them per docs/specs/extensions.md. Add a one-line entry to the `src/lib/` catalog in CODEBASE.md.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F531&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Afc7a5c00-add6-11f1-9f5c-f306fb63c63c%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Afc7a5c00-add6-11f1-9f5c-f306fb63c63c&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/paths.test.ts
scottlovegrove: Done in d715352 — every expected path is now built with `join`, so the Windows cases hold whichever host runs the suite rather than only on Linux.

src/lib/paths.ts
scottlovegrove: Done in d715352. `xdgBase()` ignores a value that is not absolute and falls through to the platform default, for `XDG_DATA_HOME` and `XDG_STATE_HOME` alike, with tests for `cache`, `./cache`, `../cache` and the empty string.

The same commit also moves `APP_NAME` into `src/lib/app-name.ts`, which was the other point from the summary — a rename can no longer move the config without moving the extensions with it.

src/lib/paths.test.ts
scottlovegrove: Same fix as the other separator thread: d715352 builds every expected path with `join`, including the state-under-data check, which is now `join(getDataDir(), 'state')`.

scottlovegrove: Same fix as the other separator threads — d715352. The Windows constants are now joined rather than spelled out, so the assertions follow the host's separator.

nvignola (commented):
✅

PR #532, opened 2026-09-11
refactor(cli): resolve the command token before scanning argv

scottlovegrove:
The lazy loader found the command to import by scanning the whole of argv for the first token matching a known command name. That reads arguments that belong to something else: `td --progress-jsonl task today` found `task`, imported its module and started the early spinner, while commander went on to run `today`.

Walk argv from the start instead, stepping over global flags and the values of the two that take one, and stop at the first remaining token. Nothing after it is inspected. The tests run commander itself over the same arguments and compare, so a root option added without a matching entry in `VALUE_FLAGS` fails there rather than misrouting at runtime.

Resolving a token to a command now uses `Object.hasOwn` rather than `in`, so a token like `constructor` cannot reach through the prototype chain to something that is not a command.

### DIFF
--- a/src/lib/command-token.test.ts
+++ b/src/lib/command-token.test.ts
@@ -1,0 +1,178 @@
+import { Command } from 'commander'
+import { describe, expect, it } from 'vitest'
+import { findCommandToken } from './command-token.js'
+
+describe('findCommandToken', () => {
+    it('returns nothing when there are no arguments', () => {
+        expect(findCommandToken([])).toEqual({ token: undefined, index: 0 })
+    })
+
+    it('returns the first positional argument and its index', () => {
+        expect(findCommandToken(['task', 'list'])).toEqual({ token: 'task', index: 0 })
+        expect(findCommandToken(['--accessible', 'task'])).toEqual({ token: 'task', index: 1 })
+    })
+
+    it('does not inspect anything after the token', () => {
+        // `task` here belongs to the extension, not to td.
+        expect(findCommandToken(['goals', 'task', 'list'])).toEqual({ token: 'goals', index: 0 })
+    })
+
+    it('steps over the value of an option that takes one', () => {
+        expect(findCommandToken(['--user', 'alice', 'task'])).toEqual({ token: 'task', index: 2 })
+        expect(findCommandToken(['--progress-jsonl', '/tmp/out', 'task'])).toEqual({
+            token: 'task',
+            index: 2,
+        })
+    })
+
+    it('steps over nothing for the inline form, which carries its own value', () => {
+        expect(findCommandToken(['--user=alice', 'task'])).toEqual({ token: 'task', index: 1 })
+    })
+
+    it('leaves a forgotten value alone rather than hiding the token behind it', () => {
+        // `--json` is not a value, it is the next flag. Commander reports the
+        // missing value itself; swallowing it here would lose `today`.
+        expect(findCommandToken(['--user', '--json', 'today'])).toEqual({
+            token: 'today',
+            index: 2,
+        })
+    })
+
+    it('never treats a boolean option as taking a value', () => {
+        expect(findCommandToken(['--accessible', 'task'])).toEqual({ token: 'task', index: 1 })
+        expect(findCommandToken(['--no-spinner', 'task'])).toEqual({ token: 'task', index: 1 })
+    })
+
+    it.each([['-v'], ['-q'], ['-vvv'], ['-vq']])('skips the short option %s', (flag) => {
+        expect(findCommandToken([flag, 'task'])).toEqual({ token: 'task', index: 1 })
+    })
+
+    it('reads the entry after `--` as the command, dash or not', () => {
+        expect(findCommandToken(['--', 'task'])).toEqual({ token: 'task', index: 1 })
+        expect(findCommandToken(['--', '--json'])).toEqual({ token: '--json', index: 1 })
+    })
+
+    it('returns nothing for a trailing `--`', () => {
+        expect(findCommandToken(['--user', 'alice', '--'])).toEqual({
+            token: undefined,
+            index: 3,
+        })
+    })
+
+    it('treats a bare dash as a token, since it is the stdin placeholder', () => {
+        expect(findCommandToken(['-'])).toEqual({ token: '-', index: 0 })
+    })
+
+    it('treats an empty argument as a token', () => {
+        expect(findCommandToken(['', 'task'])).toEqual({ token: '', index: 0 })
+    })
+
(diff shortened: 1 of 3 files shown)


src/lib/command-token.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        ['task', 'list', '--json'],
+        ['today'],
+        ['--accessible', 'task', 'list'],
+        ['-v', 'today'],
+        ['-vvv', 'today'],
+        ['-vq', 'today'],
+        ['-q', 'doctor', '--offline'],
+        ['--user', 'alice', 'today'],
+        ['--user=alice', 'today'],
+        ['--user', '--quiet', 'today'],
```
doistbot, line 150: <a href="#">P1</a>  This agreement case fails. In the mirrored program, `--user` is declared as `--user <ref>`, so Commander treats the following `--quiet` as a missing option-argument, throws (caught by `catch {}`), and `commanderDispatch` returns `undefined`. `findCommandToken` skips `--user` without consuming `--quiet` and returns `'today'`, so `expect(ours).toBe(commanderDispatch(argv))` compares `'today'` to `undefined`. Production `index.ts` avoids this path by validating/stripping `--user` before Commander parses; either mirror that stripping/validation in `commanderDispatch`, or remove this case from the agreement table.

doistbot (commented):
This PR fixes argv misrouting in the lazy command loader by walking argv left-to-right—stepping over global flags and their values and stopping at the first command token—instead of scanning the whole of argv, and hardens token resolution with `Object.hasOwn` to close the prototype-chain lookup path. Security review found no new trust-boundary or exploit concerns in the changed parsing, lazy-loading, and completion-server paths.

Few things worth tightening:
- One commander-agreement test case doesn't hold: with `--user <ref>` mirrored in the test's Commander program, `td --user --quiet task today` makes Commander treat `--quiet` as the option-argument and throw, so `commanderDispatch` returns `undefined` while `findCommandToken` still resolves `today`. Either mirror the stripping/validation production `index.ts` does before Commander parses, or drop that case from the agreement table.
- Worth noting the agreement test's hand-copied Commander mirror can't fully guarantee drift detection against future root options; if a root option with a value is added without a matching `VALUE_FLAGS` entry, consider an assertion that ties the two together so the stated protection is actually enforced.

<details>

- **<a href="#">P3</a> src/lib/command-token.test.ts:90**: The comment overclaims what this test can detect. `commanderDispatch` is a hand-copied mirror of `index.ts`'s options, so if someone adds a value-taking root option to `index.ts` only (the most likely drift), the mirror lacks it too and the agreement tests still pass: `findCommandToken` returns the option's value as the token, the mirrored commander errors on the unknown flag, and both yield `undefined`. The same overclaim appears in the `VALUE_FLAGS` doc in command-token.ts ("the tests check the two agree"). Reword to say the mirror must be kept in step manually, or actually assert the mirror's declared options against the real ones.
- **<a href="#">P3</a> src/lib/command-token.ts:38**: `valueFlags` is speculative API surface: both production call sites use the default set, and only the test passes a custom one. The helper is already coupled to the root options through `VALUE_FLAGS`, so drop the parameter (and the caller-supplied-set test) until a real second caller needs it.
- **<a href="#">P3</a> src/lib/command-token.test.ts:7**: `CommandToken.index` is never read by either call site (`index.ts` uses only `.token`), yet almost every assertion here also pins the exact `index` value. This locks in an unused implementation detail and makes every test edit more brittle. Drop the field from the return type, or assert only `token`.
- **<a href="#">P3</a> src/index.ts:209**: The `Object.hasOwn` / `constructor` fix in `resolveCommandName` has no test. The `findCommandToken` suite would still pass if someone reverted this to `in`, leaving the prototype-chain regression this PR fixes unprotected (e.g. `completion-server` resolving `constructor` to `Object` and crashing). Add a small test around resolution, for example by exporting the resolver from `command-token.ts` and covering `constructor`/`toString`.
- **<a href="#">P3</a> src/lib/command-token.test.ts:42**: This test duplicates coverage: `findCommandToken(['--accessible', 'task'])` is already asserted at line 12, and boolean-flag handling is pinned against commander by the agreement cases (`['--accessible', 'task', 'list']`, `['--no-spinner', 'today']`). Drop this test or fold it into the 'returns the first positional argument' test.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F532&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Affb059b0-add6-11f1-954b-2f5272eccfe3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Affb059b0-add6-11f1-954b-2f5272eccfe3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/command-token.test.ts
scottlovegrove: This case passes, and the reasoning here does not hold: commander consumes the following token as the value of `--user <ref>` even when it looks like a flag. It only raises `optionMissingArgument` when the option is the last entry in argv.

Verified directly against commander 15 with a program declaring the same root options:

```
["--user","--quiet","today"] -> dispatched: today   user = "--quiet"
["--user","today"]           -> dispatched: (none)  (commander printed help)
["--user"]                   -> threw commander.optionMissingArgument
```

`findCommandToken` returns `today` for the first case because `--quiet` starts with a dash and so is not taken as a value, which is deliberate — a forgotten value is commander's to report, and stepping over the flag here would hide the command token behind it. Both sides answer `today`, so the assertion compares `'today'` to `'today'`. The agreement suite is green; it would have failed loudly otherwise, which is what it is for.

Not changing this.

henningmu (commented):
✅

PR #533, opened 2026-09-11
feat(extensions): the extension command group, list and exec

scottlovegrove:
Add `registerExtensionGroup`, which puts `extension` (alias `ext`) on a program the host owns, with `list` and `exec` to start with. Everything it needs comes from the manager it is handed — colours, accessible mode, the binary name, the trust warning — so it stays free of host imports and moves to cli-core with the rest of the directory.

`list` renders the five columns from the spec, folding the notes that apply to few installs (pinned, shadowed, not executable, a manifest from a newer format) into the kind column rather than giving each one a column that is empty for almost every row. `--json` reports where the entry sits as well as where it points, which differ for a local install.

`exec` reads its arguments from argv rather than from commander, which consumes any root option it recognises before the action runs. `passThroughOptions` would be the obvious fix and is not usable: it requires `enablePositionalOptions` on the parent, which makes ordinary commands reject global flags they accept today.

### DIFF
--- a/src/lib/extensions/commands.test.ts
+++ b/src/lib/extensions/commands.test.ts
@@ -7,6 +7,7 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { writeFakeGitRepo, writeFixtureExtension } from '../../test-support/extension-fixture.js'
 import { registerExtensionGroup } from './commands.js'
 import { createExtensionManager, type ExtensionManager } from './manager.js'
+import { run as runProcess } from './run.js'
 
 describe('registerExtensionGroup', () => {
     let root: string
@@ -176,7 +177,35 @@ describe('registerExtensionGroup', () => {
             expect(lines().join('\n')).toContain('manifest needs a newer td')
         })
 
-        it('shows the pin, and the short SHA for a clone', async () => {
+        it('shows the short SHA of a clone', async () => {
+            const dir = await writeFixtureExtension(extensionsDir, 'standup', {})
+            await runProcess('git', ['init', '--quiet', '--initial-branch=main'], { cwd: dir })
+            await runProcess('git', ['add', '.'], { cwd: dir })
+            await runProcess(
+                'git',
+                [
+                    '-c',
+                    'user.email=t@example.com',
+                    '-c',
+                    'user.name=T',
+                    'commit',
+                    '--quiet',
+                    '-m',
+                    'init',
+                ],
+                { cwd: dir },
+            )
+            const head = (
+                await runProcess('git', ['rev-parse', 'HEAD'], { cwd: dir })
+            ).stdout.trim()
+
+            await run(makeManager(), ['extension', 'list'])
+            const row = lines().find((line) => line.startsWith('standup'))
+            expect(row).toContain(head.slice(0, 8))
+            expect(row).toContain('git')
+        })
+
+        it('shows the pin, and the repository a clone came from', async () => {
             const dir = await writeFixtureExtension(extensionsDir, 'standup', {})
             await writeFakeGitRepo(dir, 'https://github.com/example/td-standup.git')
             await writeFixtureExtension(extensionsDir, 'goals', {
@@ -264,6 +293,14 @@ describe('registerExtensionGroup', () => {
                 '--raw',
             ])
             expect((await reported()).args).toEqual(['--json', '-q', '--accessible', '--', '--raw'])
+        })
+
+        it('forwards --help rather than answering it', async () => {
+            // The only way to reach the usage of an extension a built-in
+            // command is hiding, so commander must not take the flag first.
+            await writeFixtureExtension(extensionsDir, 'goals', {})
+            await run(makeManager(), ['extension', 'exec', 'goals', '--help'])
+            expect((await reported()).args).toEqual(['--help'])
         })
 
         it('runs an extension with no arguments at all', async () => {
--- a/src/lib/extensions/commands.test.ts
+++ b/src/lib/extensions/commands.test.ts
@@ -1,0 +1,296 @@
+import { mkdir, mkdtemp, readFile, rm, symlink } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { captureConsole, captureStream, createTestProgram } from '@doist/cli-core/testing'
+import type { Command } from 'commander'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { writeFakeGitRepo, writeFixtureExtension } from '../../test-support/extension-fixture.js'
+import { registerExtensionGroup } from './commands.js'
+import { createExtensionManager, type ExtensionManager } from './manager.js'
+
+describe('registerExtensionGroup', () => {
+    let root: string
+    let dataDir: string
+    let extensionsDir: string
+    let warnings: string[]
+    let out: ReturnType<typeof captureConsole>
+
+    function makeManager(
+        options: { reserved?: string[]; accessible?: boolean } = {},
+    ): ExtensionManager {
+        return createExtensionManager({
+            binName: 'td',
+            envPrefix: 'TD',
+            version: '5.3.1',
+            dataDir,
+            stateDir: join(root, 'state'),
+            configDir: join(root, 'config'),
+            hostPath: '/host/dist/index.js',
+            reservedNames: () => options.reserved ?? [],
+            officialSource: { host: 'github.com', owner: 'Doist' },
+            officialLabel: 'Todoist',
+            isAccessible: () => options.accessible ?? false,
+            warn: (message) => warnings.push(message),
+            log: () => undefined,
+        })
+    }
+
+    function makeProgram(manager: ExtensionManager): Command {
+        return createTestProgram((program) => {
+            registerExtensionGroup(program, manager)
+        })
+    }
+
+    /** Run the group the way the CLI would, argv included. */
+    async function run(manager: ExtensionManager, args: string[]): Promise<void> {
+        const argv = ['node', 'td', ...args]
+        vi.spyOn(process, 'argv', 'get').mockReturnValue(argv)
+        await makeProgram(manager).parseAsync(argv)
+    }
+
+    /** Everything written to stdout, as lines. */
+    function lines(): string[] {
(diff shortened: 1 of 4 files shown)


src/lib/extensions/commands.ts
```diff
@@ -0,0 +1,201 @@
...
+        // Whatever follows the name belongs to the extension. The arguments
+        // are taken from argv rather than from these declarations, which exist
+        // so that `--help` describes the shape.
+        .allowUnknownOption()
+        .allowExcessArguments()
+        .action(async (name: string) => {
+            // `require` first, so an unknown name is a clean error rather than
+            // a failed spawn.
+            const target = await manager.require(name)
+            process.exitCode = await manager.dispatch(target.name, argsAfter('exec', name))
```
doistbot, line 200: <a href="#">P2</a>  `exec` promises that everything after the extension name passes through untouched, but `--user` is removed before this code runs: `src/index.ts` rewrites `process.argv` via `stripUserFlag` whenever `--user` appears anywhere. So `td extension exec goals --user foo` drops the flag (and the account selection is never forwarded to `manager.dispatch`). Add a host-level carve-out so `--user` after `extension exec <name>` is preserved for the extension, or stop claiming full passthrough for this flag.

src/lib/extensions/commands.ts
```diff
@@ -0,0 +1,201 @@
...
+            'after',
+            `
+An extension is an executable named \`${binName}-<name>\` that ${binName} runs as
+\`${binName} <name> …\`, passing everything after the name through untouched. It can
+be written in any language.
+
+${manager.trustWarning}
+
+Examples:
+  ${binName} extension install owner/${binName}-goals
```
doistbot, line 172: <a href="#">P2</a>  This help text and the `list` empty-state message (line 134) tell users to run `extension install`, `extension upgrade`, and `extension remove`, but `registerExtensionGroup` only registers `list` and `exec`. Running those examples after this PR fails as an unknown command. Remove the examples/instructions until those subcommands land, or register them here.

src/lib/extensions/commands.ts
```diff
@@ -0,0 +1,201 @@
...
+        )
+
+    extension
+        .command('list')
+        .description('List installed extensions')
+        .option('--json', 'Output as JSON')
+        .action((options) => listExtensions(manager, options))
+
+    extension
+        .command('exec <name> [args...]')
```
doistbot, line 186: <a href="#">P2</a>  Commander registers a built-in help option on every command, so `td ext exec goals --help` (or `-h`) prints the exec subcommand's help and the action never runs. That is the same flag-eating problem the `argsAfter` workaround exists to solve, and it contradicts the contract in the group help text: "passing everything after the name through untouched". An extension that ships its own `--help` can never receive it. Add `.helpOption(false)` to the exec command so `--help` is treated as a plain argument.

src/lib/extensions/commands.ts
doistbot, line 200: <a href="#">P2</a>  This forwards untrusted arguments into the current Windows `.cmd` dispatch path, which invokes `cmd.exe /c` with the executable path and args. `cmd.exe` interprets metacharacters such as `&` and `|`, so a caller that supplies an argument like `& whoami` to a trusted batch extension can execute an additional command. Escape arguments for `cmd.exe` before dispatching batch files, or reject `.cmd` extensions rather than passing their argv to `cmd /c`.

src/lib/extensions/commands.test.ts
```diff
@@ -0,0 +1,296 @@
...
+                    tag: 'v0.3.0',
+                    pinned: true,
+                    asset: 'a',
+                    installedAt: '2026-09-07T10:12:00Z',
+                },
+            })
+
+            await run(makeManager(), ['extension', 'list'])
+            const rendered = lines().join('\n')
+            expect(rendered).toContain('binary (pinned)')
```
doistbot, line 225: <a href="#">P2</a>  The test title promises the short SHA for a git clone, but the only assertion (`example/td-standup`) checks the SOURCE column. `writeFakeGitRepo` writes just `.git/config` with no HEAD, so `git rev-parse HEAD` fails and the VERSION column renders as `—`; the short-SHA rendering is never exercised. Create a real git fixture and assert the 8-character SHA (as `upgrade.test.ts` does), or rename the test to match what it actually asserts.

doistbot (commented):
Adds a host-agnostic `extension` (alias `ext`) command group with `list` and `exec`, delegating discovery, resolution, and spawning to the extension manager.

Few things worth tightening:
- `exec`'s full passthrough contract doesn't hold: the host strips `--user` from argv before dispatch, and Commander still intercepts `--help`/`-h` — both should reach the extension untouched.
- Help and empty-state text point to `install`/`upgrade`/`remove` subcommands that aren't registered yet.
- On Windows, untrusted args reach the `.cmd` dispatch path via `cmd.exe /c`, where metacharacters like `&` can run extra commands — escape or reject batch extensions.
- The short-SHA list test never actually exercises SHA rendering (the fixture has no real git repo), so either fix the fixture or rename the test.

<details>

- **<a href="#">P3</a> src/lib/extensions/commands.ts:45**: `shorten` only collapses paths when they start with `${homedir()}/`. On Windows, both `homedir()` and the paths produced by `resolveLocalTarget` use backslashes, so local extension paths are never shown as `~/...`. Use a separator-aware prefix check or `path.relative(homedir(), path)` when Windows display is intended.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:196**: This `require` call is redundant: `manager.dispatch(selector, …)` already calls `requireExtension` internally (manager.ts, `dispatch`) and throws the same `EXTENSION_NOT_FOUND` CliError with the same hints before anything is spawned. So `await manager.require(name)` just runs the extension discovery scan twice per exec, and the inline comment's rationale ("require first, so an unknown name is a clean error") doesn't hold — dispatch already guarantees that. Simplify to `process.exitCode = await manager.dispatch(name, argsAfter('exec', name))` and drop the comment.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:211**: `vi.unstubAllEnvs()` runs after the assertion, so if the assertion at lines 208-210 fails, `HOME` stays stubbed for the rest of the file. Move the unstub to an `afterEach` (or wrap the body in `try/finally`) so cleanup also happens on failure.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:277**: The test title says it accepts the owner-qualified form, but only `td-goals` (the directory-name form) is passed. `Doist/td-goals` is never run through `exec`, so the claimed coverage is missing. Either add a second run for the owner-qualified selector and assert it resolves, or narrow the title to the directory-name form.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F533&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A02904820-add7-11f1-8fbb-9fe918f69a30%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A02904820-add7-11f1-8fbb-9fe918f69a30&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/commands.ts
scottlovegrove: Real, and fixed — though in #537, where the stripping actually happens. 7b440cc skips the pre-parse `--user` handling entirely when the command token is the `extension` group, since those arguments belong to the extension rather than to td.

Verified against a built CLI: `td extension exec hello --user alice --json` now reports `args: ["--user","alice","--json"]`.

The consequence is that `--user` is no longer accepted by the group's other subcommands, where it meant nothing anyway. #539 records that in the spec.

src/lib/extensions/commands.ts
scottlovegrove: This is the stack boundary rather than a defect: `install`, `upgrade` and `remove` are the next commit, #534, and the group is never released without them. Splitting the help text across the two PRs would leave it in a shape neither branch wants.

src/lib/extensions/commands.ts
scottlovegrove: Done in c9d22eb — `exec` now declares `.helpOption(false)`, so `td extension exec goals --help` reaches the extension. You are right that this was the one form that could ask a shadowed extension for its usage.

There is a test asserting the fixture receives exactly `['--help']`.

src/lib/extensions/commands.ts
scottlovegrove: Real, and fixed in c9d22eb. Node quotes an argument only when it contains a space, so `&whoami` did arrive at `cmd.exe` bare.

Every argument is now quoted into a single command line written for `cmd.exe`, spawned with `windowsVerbatimArguments` so node does not quote it again. Arguments containing `%`, `!` or a newline are refused rather than escaped: cmd expands those from inside quotes and no escaping holds in every context a batch file can create — the same choice Rust's standard library makes for `.bat`.

Four tests cover it, including `&whoami` and the embedded-quote case. Worth saying plainly that this is not verified on a real Windows machine; the spec still has extensions as best-effort there.

src/lib/extensions/commands.test.ts
scottlovegrove: Done in c9d22eb. The test now builds a real repository with `git init` and a commit, and asserts the first eight characters of `rev-parse HEAD` in the VERSION column. The fake-repo case kept its own test for the SOURCE column, which is what it was really checking.

craigcarlyle:
✅

PR #534, opened 2026-09-11
feat(extensions): install, upgrade and remove commands

scottlovegrove:
Complete the management surface. `install` reports where the extension landed, saying "Linked" for a local directory, since that install is a symlink to a working copy rather than a copy of anything. The trust warning is left to the manager, which prints it before any step that can run code from the repository — printing it here as well would double it, and a local install does not warrant it at all.

`upgrade` shows one row per extension with what changed or why nothing did, wrapped in the standard dry-run block when previewing. Naming extensions and passing `--all` together is a usage error rather than a silent preference for one; passing neither shows the usage, so a bare invocation never reaches the network.

`remove` says plainly that a local install's own directory was left alone, which is the part a user most needs to be sure of.

### DIFF
--- a/src/lib/extensions/commands.ts
+++ b/src/lib/extensions/commands.ts
@@ -12,10 +12,11 @@
  */
 
 import { homedir } from 'node:os'
-import { formatJson, printEmpty } from '@doist/cli-core'
+import { join } from 'node:path'
+import { CliError, formatJson, printEmpty } from '@doist/cli-core'
 import type { Command } from 'commander'
 import type { ExtensionManager } from './manager.js'
-import type { ExtensionListing } from './types.js'
+import type { ExtensionListing, RemoveResult, UpgradeResult } from './types.js'
 
 /** Neither column has a value worth showing, so say so once, the same way. */
 const NOTHING = '—'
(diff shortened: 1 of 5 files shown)


doistbot (commented):
This PR completes the extension management surface with `install`, `upgrade`, and `remove` commands, with nice touches like the "Linked" wording for local installs and a plain note that removal leaves a local working copy untouched.

Few things worth tightening:
- Add the new `extension` subcommands and their flags to `SKILL_CONTENT` and regenerate `skills/todoist-cli/SKILL.md`, so agents can discover and invoke them (CI's skill-sync check will catch this).
- Run git error output (stderr from `pull`/`fetch`/`reset` surfaced as `UpgradeResult.detail`) through `redactUrl` before display — a failed fetch on a remote with embedded credentials could otherwise echo the token to the terminal and logs.
- Also strip ANSI/OSC control sequences from captured git diagnostics before printing them, so a hostile remote can't write raw escape bytes into the user's terminal.

<details>

- **<a href="#">P3</a> src/lib/extensions/commands.ts:310**: A bare `upgrade` (`names` empty and no `--all`) calls `upgradeCmd.help()`, which prints human-readable help and exits successfully even when `--json` was passed. A machine consumer running `td extension upgrade --json` therefore gets non-JSON stdout with exit code 0 instead of a parseable result or error. When `options.json` is set, throw a `CliError` here so the global handler renders it as JSON, rather than falling through to `help()`.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:248**: `RemoveResult` already carries the exact path that was deleted (`result.path`, set to `extension.entryPath` in remove.ts). Rebuilding it here as `join(manager.extensionsDir, `${manager.binName}-${result.name}`)` duplicates the naming rule and silently assumes `dirName` is always `${binName}-${name}` — true for names discovered through `toCommandName`, but not for an entry in the extensions directory that lacks the prefix. Using `shorten(result.path)` is simpler and always correct.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:221**: This duplicates the empty-state handling that `printEmpty` already provides; the helper is imported above and used for `listExtensions`'s empty case. Reuse it here so the message goes through the same quiet/output handling instead of an unconditional `console.log`: `printEmpty({ options, message: 'No extensions installed.' })`.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:303**: The test title says the bare invocation shows usage, but the assertion only checks that the action rejects — any thrown error would satisfy it. Capture stdout (as the group-help test above does with `captureStream('stdout')`) and assert the usage text so the behavior is actually pinned.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:312**: This mkdir/writeFixtureExtension/symlink block is repeated four times across the upgrade and remove tests to create a linked local install. Extract one helper (like the `localSource` helper already in the install tests) and call it from each test to remove the copy-paste.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:249**: These local-install tests assume POSIX symlinks. On Windows `installLocal` records the target in a plain path file instead of a symlink (see `install.ts`/`discover.ts`), so `lstat(entry).isSymbolicLink()` fails there; the same assertion repeats at line 282, and the manual `symlink(...)` setups at lines 314, 324, 343, and 356 need elevated rights on Windows. Mirror `install.test.ts` by skipping these symlink assertions on `win32` or by building the path-file representation.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F534&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A05ee6740-add7-11f1-9c16-e2993970a3ab%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A05ee6740-add7-11f1-9c16-e2993970a3ab&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/commands.ts
```diff
@@ -182,6 +290,36 @@ Examples:
         .option('--json', 'Output as JSON')
         .action((options) => listExtensions(manager, options))
 
+    extension
+        .command('install <source>')
```
doistbot, line 294: <a href="#">P2</a>  Add the new extension management subcommands and their flags to `SKILL_CONTENT`, then regenerate `skills/todoist-cli/SKILL.md`. The skill is the command reference used by agents, and it currently contains no `td extension` commands, so agents cannot discover or correctly invoke install, upgrade, or remove.

src/lib/extensions/commands.ts
```diff
@@ -152,6 +153,113 @@ async function listExtensions(
...
+    upgraded: 'upgraded',
+    'up-to-date': 'up to date',
+    skipped: 'skipped',
+    'would-upgrade': 'would upgrade',
+}
+
+/** What changed, or why nothing did. */
+function describeOutcome(result: UpgradeResult): string {
+    if (result.from && result.to) return `${result.from} → ${result.to}`
+    return result.detail ?? ''
```
doistbot, line 195: <a href="#">P2</a>  `describeOutcome` returns `result.detail` verbatim, and `upgrade` prints it for skipped rows. For git upgrades that detail is the raw stderr tail from `git pull`/`git fetch`/`git reset` (see `pull`/`resetToRemote` in `git.ts`), which is not run through `redactUrl` the way `clone` and `checkout` errors are. A failing fetch on a private-repo remote whose URL carries `user:token@` credentials can make git echo that URL in its error, so this line would print the token to the terminal and any captured logs. Apply `redactUrl` to those git error strings before they become `UpgradeResult.detail`, or redact the detail here before display.

src/lib/extensions/commands.ts
```diff
@@ -152,6 +153,113 @@ async function listExtensions(
...
+            manager.theme.yellow(`[dry-run] Would upgrade ${count(changing, 'extension')}:`),
+        )
+    }
+
+    const indent = options.dryRun ? '  ' : ''
+    for (const line of alignRows(
+        results.map((result) => [
+            result.name,
+            OUTCOME_LABELS[result.outcome],
+            describeOutcome(result),
```
doistbot, line 237: <a href="#">P2</a>  Sanitize process diagnostics before rendering them in the terminal. `result.detail` can be the raw stdout/stderr captured from `git pull`; a hostile non-GitHub remote can emit ANSI/OSC control sequences in a failed fetch, and `extension upgrade` now writes those bytes directly to the user's terminal. Strip terminal control characters (and preserve/redact safe diagnostics) before including `detail` in display rows.

src/lib/extensions/commands.ts
scottlovegrove: Lands in #539, the last commit of this stack, which adds the `### Extensions` section to `SKILL_CONTENT` and regenerates `skills/todoist-cli/SKILL.md`. Holding it there keeps the regeneration to one commit rather than three.

src/lib/extensions/commands.ts
scottlovegrove: Real, and fixed in f4e34b5. `pull`, `fetch` and `reset` now build their error through the same `redactUrl` that `clone` and `checkout` already used.

One thing worth noting for anyone reading this later: current git already elides credentials from that particular message — `https://alice:s3cret@host/x.git` comes back as `unable to access 'https://host/x.git/'`. The redaction is still right as defence in depth, and it makes the four call sites consistent, but it is not the only thing standing between a token and the terminal.

src/lib/extensions/commands.ts
scottlovegrove: Done in f4e34b5, and one layer lower than suggested: control characters other than tab and newline are stripped in `outputHints`, so this covers the install and dependency-install paths too rather than only what `upgrade` prints. New tests in `run.test.ts` cover the C0 range, the C1 range, and that tabs and newlines survive.

PR #535, opened 2026-09-11
feat(extensions): pass-through commands and the unknown-command hint

scottlovegrove:
Register one command per installed extension, so `--help` lists them under their own heading and name completion works with no special cases. A name a built-in already uses is skipped rather than replacing it; `list` reports it as shadowed and `exec` still runs it.

`passThroughOptions` is what this looks like it should use, and cannot: commander requires `enablePositionalOptions` on the parent for it, which makes every ordinary command reject the global flags it accepts today. The arguments are read from argv instead, which is stronger anyway — the host can short-circuit before commander parses at all.

The unknown-command hint is added at `outputError`, because commander writes that error and exits without the parse promise ever rejecting. Going through `outputError` keeps commander's own wording and its "did you mean" suggestion intact. It appears only for someone with no extensions installed, who is the only person it can tell anything.

### DIFF
--- a/src/lib/extensions/commands.ts
+++ b/src/lib/extensions/commands.ts
@@ -351,6 +351,17 @@ export type ExtensionCommands = {
 }
 
 /**
+ * How commander opens the message it writes for an unknown command.
+ *
+ * Matching on the rendered text is not ideal — the error code would be
+ * steadier — but `outputError` is handed the finished string and nothing else,
+ * and it is the only hook that runs before commander exits. Naming the prefix
+ * at least puts it somewhere a commander upgrade can be checked against, and
+ * the tests pin it.
+ */
+export const UNKNOWN_COMMAND_PREFIX = 'error: unknown command'
+
+/**
  * Point at the feature from the one place someone will be looking when they
  * have mistyped a command, and only when they have no extensions — anyone who
  * has installed one does not need telling.
@@ -367,7 +378,7 @@ function addInstallHint(program: Command, binName: string): void {
     program.configureOutput({
         outputError(message, write) {
             base(message, write)
-            if (message.startsWith('error: unknown command')) {
+            if (message.startsWith(UNKNOWN_COMMAND_PREFIX)) {
                 write(
                     `\nRun \`${binName} extension install <owner/repo>\` to add commands from extensions.\n`,
                 )
--- a/src/lib/extensions/commands.ts
+++ b/src/lib/extensions/commands.ts
@@ -16,7 +16,7 @@ import { join } from 'node:path'
 import { CliError, formatJson, printEmpty } from '@doist/cli-core'
 import type { Command } from 'commander'
 import type { ExtensionManager } from './manager.js'
-import type { ExtensionListing, RemoveResult, UpgradeResult } from './types.js'
+import type { DispatchOptions, ExtensionListing, RemoveResult, UpgradeResult } from './types.js'
 
 /** Neither column has a value worth showing, so say so once, the same way. */
 const NOTHING = '—'
@@ -30,13 +30,14 @@ const NOTHING = '—'
  * it requires `enablePositionalOptions` on the parent, which makes ordinary
  * commands reject the global flags they accept today.
  *
- * The search starts after the subcommand token so that a value elsewhere on
- * the line cannot be mistaken for the name.
+ * `marker` is the subcommand the name follows, so that a value elsewhere on
+ * the line cannot be mistaken for it. An extension invoked directly has no
+ * marker: its name is the first thing on the line that is not a global flag.
  */
-function argsAfter(marker: string, name: string): string[] {
-    const markerIndex = process.argv.indexOf(marker, 2)
-    if (markerIndex === -1) return []
-    const nameIndex = process.argv.indexOf(name, markerIndex + 1)
+function argsAfter(name: string, marker?: string): string[] {
+    const start = marker ? process.argv.indexOf(marker, 2) : 1
+    if (start === -1) return []
+    const nameIndex = process.argv.indexOf(name, start + 1)
     return nameIndex === -1 ? [] : process.argv.slice(nameIndex + 1)
 }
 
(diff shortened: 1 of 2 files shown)


src/lib/extensions/commands.ts
```diff
@@ -332,8 +333,103 @@ Examples:
...
+ * wording, and its "did you mean" suggestion, exactly as they are.
+ */
+function addInstallHint(program: Command, binName: string): void {
+    const base = program.configureOutput().outputError
+    if (!base) return
+
+    program.configureOutput({
+        outputError(message, write) {
+            base(message, write)
+            if (message.startsWith('error: unknown command')) {
```
doistbot: <a href="#">P2</a>  The hint is gated on commander's rendered error string (`message.startsWith('error: unknown command')`). That is message-based error classification: if a commander upgrade rewords this line, the hint silently stops appearing even though the error is still an unknown command, and the test only pins today's wording. Commander exposes a stable code (`commander.unknownCommand` on the `CommanderError` when `exitOverride()` is enabled); prefer that, or at least keep the matched prefix in one named constant so a wording change fails loudly instead of quietly dropping the feature.

src/lib/extensions/commands.ts
```diff
@@ -332,8 +333,103 @@ Examples:
...
+    if (extensions.length === 0) addInstallHint(program, manager.binName)
+
+    return {
+        names,
+        dispatch: (name, args, options) => manager.dispatch(name, args, options),
+    }
+}
+
+/** The whole feature in one call, for a host that wants it that way. */
+export async function registerExtensionCommands(
```
doistbot, line 443: <a href="#">P1</a>  Wire this into the executable before exposing the combined registrar. There is no production caller for `registerExtensionCommands` (or `registerExtensionPassThrough`): `src/index.ts` never creates an extension manager or registers these commands. As shipped, installed extensions remain unknown commands and the new help group and install hint are unreachable.

doistbot (commented):
This PR registers each installed extension as a real commander command so they show up in help and completion, routes unknown-name invocations through to the extension via argv, and adds an install hint on unknown commands. Few things worth tightening:

- `registerExtensionCommands`/`registerExtensionPassThrough` have no production caller yet — wiring them into `src/index.ts` is needed before the new help group and hint are reachable.
- The hint is gated on commander's rendered error string (`message.startsWith('error: unknown command')`); commander exposes a stable `commander.unknownCommand` code via `exitOverride()`, which would keep the hint alive if a commander upgrade rewords the message (or keep the matched prefix in one named constant so a wording change fails loudly).

<details>

- **<a href="#">P3</a> src/lib/extensions/commands.ts:38**: The no-marker branch sets `start` to 1 and then scans with `process.argv.indexOf(name, 2)`, which does not skip global options that consume a value. For `td --progress-jsonl goals goals list`, the first `goals` is the flag's path value, so the extension receives `['goals', 'list']` instead of `['list']`. `--user` is stripped before parse, but `--progress-jsonl` is not. The comment claims the name is "the first thing on the line that is not a global flag," but the code doesn't implement that; the scan needs the same flag-value-skipping behavior used to locate the command token.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:367**: This also matches unknown-subcommand errors, since commander emits the same 'error: unknown command' wording at the subcommand level and subcommands inherit the program's outputError. With no extensions installed, `td task updt` (a typo for `task update`) prints commander's 'did you mean update?' suggestion followed by 'Run `td extension install ...`', which misdirects the user. Consider gating on root-level: hint only when the quoted token is `process.argv[2]`, the same argv-reading approach the actions already use.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:424**: `dispatch` is a no-op closure over `manager.dispatch`: same signature, no added behavior, and `manager.dispatch` doesn't use `this`. Return `dispatch: manager.dispatch` (or have the host call the manager directly) instead of wrapping it.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:369**: This hint re-hardcodes the `Run \`${binName} extension install <owner/repo>\`` command already built in `listExtensions`. Extract one shared helper/constant for the install command so the two messages can't drift.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:478**: The new `makeManager` duplicates the one in the `registerExtensionGroup` describe block above — same `createExtensionManager` call with only `isAccessible`/`warn` handling differing. Consider hoisting a single factory to file scope that takes those as options, so the manager shape (binName, dirs, officialSource, …) lives in one place.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:564**: `XX_REPORT` is set and deleted inline, so if an assertion (or `parseAsync`) throws before line 576, the variable leaks into the rest of the file. The `exec` describe already isolates this with `beforeEach`/`afterEach`; move this test's env setup/teardown there (and restore `process.exitCode` in the same hook) so a failing run can't leave shared state behind.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F535&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A092bdf00-add7-11f1-8f4c-7b714e86be15%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A092bdf00-add7-11f1-8f4c-7b714e86be15&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/commands.ts
scottlovegrove: Done in 039a4b5, though not with the code. `outputError` is handed the finished string and nothing else — the `CommanderError` never reaches it — so matching on text is the only option at that hook.

What the commit does is name the prefix as `UNKNOWN_COMMAND_PREFIX` and add a test that drives commander to the error and asserts it still writes that text. A reword now fails there rather than silently dropping the hint, which was the real risk.

src/lib/extensions/commands.ts
scottlovegrove: Lands in #537, two commits up: that is where `src/index.ts` builds the manager and calls `registerExtensionPassThrough`. Nothing in this stack is released before then.

pawelgrimm (commented):
🚀

PR #536, opened 2026-09-11
feat(auth): read TD_USER as the fallback for --user

scottlovegrove:
A process the CLI starts should be able to pass its account on to the nested `td` calls it makes, without repeating `--user` on every one. Read `TD_USER` as the fallback, between the flag and the configured default: the flag is this invocation saying what it wants, the variable is the surrounding context saying who it is, and the specific instruction wins.

`TODOIST_API_TOKEN` still short-circuits ahead of both, and an unresolvable `TD_USER` reports the same `UserNotFoundError` an unresolvable `--user` does, so neither silently falls back to some other account. An empty value counts as unset.

Useful outside extensions too: `TD_USER=alice td task list` now works for anyone.

### DIFF
--- a/src/lib/auth.test.ts
+++ b/src/lib/auth.test.ts
@@ -242,6 +242,77 @@ describe('lib/auth', () => {
         await expect(resolveActiveUser({ ref: '222' })).resolves.toMatchObject({ id: '222' })
     })
 
+    it('TD_USER names the account when no --user was given', async () => {
+        vi.stubEnv('TD_USER', 'd@e.f')
+        setConfig({
+            config_version: 2,
+            users: [
+                { id: '111', email: 'a@b.c' },
+                { id: '222', email: 'd@e.f' },
+            ],
+        })
+        entryFor(keyring, 'user-222').token = 'token-222'
+
+        const { resolveActiveUser } = await import('./auth.js')
+
+        await expect(resolveActiveUser()).resolves.toMatchObject({ id: '222' })
+    })
+
+    it('an explicit ref beats TD_USER', async () => {
+        vi.stubEnv('TD_USER', '222')
+        setConfig({
+            config_version: 2,
+            users: [
+                { id: '111', email: 'a@b.c' },
+                { id: '222', email: 'd@e.f' },
+            ],
+        })
+        entryFor(keyring, 'user-111').token = 'token-111'
+
+        const { resolveActiveUser } = await import('./auth.js')
+
+        await expect(resolveActiveUser({ ref: '111' })).resolves.toMatchObject({ id: '111' })
+    })
+
+    it('TODOIST_API_TOKEN still short-circuits ahead of TD_USER', async () => {
+        vi.stubEnv('TODOIST_API_TOKEN', 'env-token-123456')
+        vi.stubEnv('TD_USER', '222')
+        setConfig({
+            config_version: 2,
+            users: [{ id: '222', email: 'd@e.f' }],
+        })
+
+        const { resolveActiveUser } = await import('./auth.js')
+
+        await expect(resolveActiveUser()).resolves.toMatchObject({ id: 'env', source: 'env' })
+    })
+
+    it('treats an empty TD_USER as unset rather than as an account name', async () => {
+        vi.stubEnv('TD_USER', '')
+        setConfig({
+            config_version: 2,
+            users: [{ id: '111', email: 'a@b.c' }],
+        })
+        entryFor(keyring, 'user-111').token = 'stored-token'
+
+        const { resolveActiveUser } = await import('./auth.js')
+
+        await expect(resolveActiveUser()).resolves.toMatchObject({ id: '111' })
+    })
+
+    it('reports an unresolvable TD_USER the same way an unresolvable --user is reported', async () => {
+        vi.stubEnv('TD_USER', 'nope')
+        setConfig({
+            config_version: 2,
+            users: [{ id: '111', email: 'a@b.c' }],
+        })
+
+        const { resolveActiveUser } = await import('./auth.js')
+        const { UserNotFoundError } = await import('./users.js')
+
+        await expect(resolveActiveUser()).rejects.toBeInstanceOf(UserNotFoundError)
+    })
+
     it('throws UserNotFoundError when ref does not match', async () => {
         setConfig({
             config_version: 2,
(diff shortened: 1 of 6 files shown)


src/lib/auth.ts
```diff
@@ -76,7 +76,12 @@ export async function resolveActiveUser(opts: { ref?: string } = {}): Promise<Re
 
     const config = await readConfig()
     const users = getStoredUsers(config)
-    const ref = opts.ref ?? getRequestedUserRef()
+    // The flag before the environment: a flag is this invocation saying what
+    // it wants, while `TD_USER` is the surrounding context saying who it is,
+    // and the specific instruction should win. `||` for the variable, because
+    // `TD_USER=` is someone clearing it, not naming an account called the
+    // empty string.
+    const ref = opts.ref ?? getRequestedUserRef() ?? (process.env[USER_ENV_VAR] || undefined)
```
doistbot, line 84: <a href="#">P2</a>  `td auth status` misses this fallback. In `commands/auth/status.ts`, `fetchLive` decides between the `store.active()` snapshot and a full re-resolve via `getApi()` using only `getRequestedUserRef() !== undefined`. With `TD_USER` set (no `--user`), the snapshot's token — the default user's — is used for `api.getUser()`, so status renders the default account while every other command acts as the `TD_USER` account. This is the same "wrong account" mismatch `createTodoistTokenStore` already guards against for `TODOIST_API_TOKEN` (see the comment in `auth-store.ts`). Fix: treat `TD_USER` like `--user` in that check, e.g. `getRequestedUserRef() !== undefined || Boolean(process.env[USER_ENV_VAR])`.

src/lib/auth.test.ts
```diff
@@ -242,6 +242,77 @@ describe('lib/auth', () => {
...
+            users: [
+                { id: '111', email: 'a@b.c' },
+                { id: '222', email: 'd@e.f' },
+            ],
+        })
+        entryFor(keyring, 'user-111').token = 'token-111'
+
+        const { resolveActiveUser } = await import('./auth.js')
+
+        await expect(resolveActiveUser({ ref: '111' })).resolves.toMatchObject({ id: '111' })
```
doistbot, line 302: <a href="#">P2</a>  This asserts `opts.ref` beats `TD_USER`, but production callers invoke `resolveActiveUser()` with no args — the real `--user` flag reaches the resolver via `getRequestedUserRef()`. The actual `--user` beats `TD_USER` contract is therefore untested: swapping `getRequestedUserRef()` and `process.env[USER_ENV_VAR]` in the fallback chain would change behavior for `td --user alice` with `TD_USER` set and no test would fail. Drive this through the flag instead: set `process.argv` to include `--user 111`, reset the global-args cache, and call `resolveActiveUser()` without `{ ref }`.

doistbot (commented):
Reads `TD_USER` as a fallback for `--user` in `resolveActiveUser`, sitting between the flag and the configured default, with `TODOIST_API_TOKEN` still short-circuiting and empty values treated as unset. Nice, focused change.

Few things worth tightening:
- `td auth status` doesn't honor `TD_USER`: with the variable set, `fetchLive` uses the default account's snapshot token, so status can render a different account than the one every other command acts as. Treating `TD_USER` like `--user` in that check would close the gap.
- The test asserts `opts.ref` beats `TD_USER`, but production callers pass no args and the flag arrives via `getRequestedUserRef()` — so the real flag-beats-env contract is untested. Driving the test through `process.argv` + `resolveActiveUser()` would cover the behavior that actually matters.

<details>

- **<a href="#">P3</a> src/lib/auth-store.ts:26**: `src/lib/skills/content.ts` still documents the old resolution order in its Multi-user section (`--user <ref>` > `user.defaultUser` > the only stored account), so the generated `skills/todoist-cli/SKILL.md` omits the new `TD_USER` fallback between the flag and the default. Update `content.ts` and run `npm run sync:skill` so agents get the correct precedence.
- **<a href="#">P3</a> src/commands/doctor.ts:341**: `TD_USER` is duplicated as a raw string here while `USER_ENV_VAR` is exported from `../lib/auth.js`, which this file already imports from. Interpolate `USER_ENV_VAR` so the message can't drift if the env var is renamed.
- **<a href="#">P3</a> src/lib/auth.test.ts:290**: This test doesn't pin the `|| undefined` behavior it names. `''` is already falsy in the later `ref ? ...` branches, so the resolver falls back to the single stored user even if the guard is removed or changed to `??`. It's effectively the existing single-user resolution test plus a stubbed env var, adding no additional regression signal. Consider dropping it or replacing it with a case that can actually distinguish empty-as-unset.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F536&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0c89d710-add7-11f1-8e4c-4d0a281b628b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0c89d710-add7-11f1-8e4c-4d0a281b628b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
scottlovegrove: Good catch, and fixed in 28de3c5 — `userOverride` now also checks `USER_ENV_VAR`, so `TD_USER` sends `status` down the re-resolve path exactly as `--user` does. There is a test asserting `createApiForToken` is not called and the override account's email is printed.

src/lib/auth.test.ts
scottlovegrove: Fair, and fixed in 28de3c5. The new test puts `--user 111` into `process.argv`, resets the global-args cache and calls `resolveActiveUser()` with no arguments.

I checked it actually bites: swapping `getRequestedUserRef()` and `process.env[USER_ENV_VAR]` in the fallback chain fails it.

PR #537, opened 2026-09-11
feat(td): wire extensions into the CLI

scottlovegrove:
Installed extensions now appear in `td --help` and run as `td <name> …`, and `td extension` manages them.

Dispatch happens before commander parses, so everything after the name reaches the extension exactly as typed — `td goals --json --user alice` is three arguments for `goals`, not td's output mode and account. The global-args parser is told where td's own arguments end so that nothing past the name changes how td behaves, and that boundary is set for extensions only: applying it to a built-in would quietly turn off JSON mode for every subcommand flag.

Discovery is gated on the command token not naming a built-in. A built-in always wins over an extension of the same name, so for `td task list` there is nothing on disk that could change the outcome, and the extension system is not loaded at all — verified with strace: zero reads of the extensions directory and zero extension modules. `--help`, completion, an extension being run, and a mistyped command are the cases that do need the answer.

The three error-reporting call sites in the entry point become one `reportFatal`, now shared with the pre-parse extension path so a failure looks the same whichever raised it.


src/lib/command-token.ts
```diff
@@ -68,3 +68,22 @@ export function findCommandToken(
...
+ * which offers them, an extension being run, and a mistyped command, which
+ * could be either.
+ *
+ * It lives here, beside the token scan and away from anything that imports the
+ * extension system, so that asking the question costs nothing.
+ */
+export function needsExtensionLookup(argv: string[], builtIn: string | undefined): boolean {
+    if (builtIn) return false
+    // `--version` prints one line and lists nothing.
+    return !argv.some((arg) => arg === '--version' || arg === '-V')
```
doistbot: <a href="#">P2</a>  `needsExtensionLookup` treats `--version`/`-V` anywhere in argv as td's own version flag. With an extension installed, `td goals --version` skips discovery, so `goals` never runs and `--version` is not passed through — contradicting the pass-through contract where everything after the name reaches the extension (`td goals --help` does dispatch). Only skip the lookup for version when there is no command token, or when the flag precedes the token.

src/index.ts
```diff
@@ -220,6 +245,44 @@ program.hook('preAction', (_thisCommand, actionCommand) => {
...
 })
 
+const rawArgs = process.argv.slice(2)
+const { token: commandToken, index: commandTokenIndex } = findCommandToken(rawArgs)
+const builtInCommand = commandToken ? resolveCommandName(commandToken) : undefined
+
+let extensions: ExtensionCommands | undefined
+if (needsExtensionLookup(rawArgs, builtInCommand)) {
+    const { setUpExtensionDispatch } = await import('./commands/extension/dispatch.js')
+    extensions = await setUpExtensionDispatch(program)
```
doistbot: <a href="#">P2</a>  `setUpExtensionDispatch` is awaited outside any `reportFatal` catch. `manager.discover()` rethrows real directory errors (e.g., EACCES reading the extensions directory), so this becomes an unhandled top-level rejection with a raw stack instead of the formatted error every other path now produces. Wrap the discovery/registration await in the same `reportFatal` handling used for dispatch.

src/index.ts
```diff
@@ -183,6 +203,11 @@ const commands: Record<string, [string, () => Promise<(p: Command) => void>, str
         'Manage CLI configuration',
         async () => (await import('./commands/config/index.js')).registerConfigCommand,
     ],
+    extension: [
```
doistbot, line 202: <a href="#">P2</a>  This adds a new top-level `extension`/`ext` command, but `src/lib/skills/content.ts` is not updated. AGENTS.md requires SKILL_CONTENT to stay in sync whenever commands are added, so add `td extension ...` to the Quick Reference and command docs, then run `npm run sync:skill`.

src/index.ts
```diff
@@ -220,6 +245,44 @@ program.hook('preAction', (_thisCommand, actionCommand) => {
     setActiveCommandPath(getActionCommandPath(actionCommand))
 })
 
+const rawArgs = process.argv.slice(2)
```
doistbot, line 244: <a href="#">P2</a>  `td extension exec goals --user alice` cannot preserve its extension arguments: the later pre-parse `stripUserFlag(originalArgs)` removes `--user alice` from `process.argv`, and the `exec` action's `argsAfter()` then reads that modified argv. Limit stripping to arguments before `exec`'s extension name (or preserve the original argv for `argsAfter`) so `exec` has the same passthrough contract as direct dispatch.

src/commands/extension/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        return realpathSync(entry)
+    } catch {
+        return entry
+    }
+}
+
+export function buildExtensionManager(program: Command): ExtensionManager {
+    // Snapshot the names already taken before any extension is registered, so
+    // an extension can never be measured as shadowing itself.
+    const reserved = new Set(
```
doistbot: <a href="#">P2</a>  This snapshot is taken after the lazy loader has removed the `extension` placeholder. Consequently, `td extension install .../td-extension` (and `td-ext`) is accepted even though those names are occupied by this group and its alias; subsequent `list` also reports it as unshadowed. Include the group's canonical name and alias in the reserved set, or construct the manager before removing its placeholder.

src/index.ts
```diff
@@ -220,6 +245,44 @@ program.hook('preAction', (_thisCommand, actionCommand) => {
...
+let extensions: ExtensionCommands | undefined
+if (needsExtensionLookup(rawArgs, builtInCommand)) {
+    const { setUpExtensionDispatch } = await import('./commands/extension/dispatch.js')
+    extensions = await setUpExtensionDispatch(program)
+}
+
+// Run the extension here rather than through commander, so that everything
+// after its name reaches it exactly as typed. Commander would consume any
+// global flag it recognises on the way past.
+if (commandToken && extensions?.names.has(commandToken)) {
```
doistbot, line 264: <a href="#">P2</a>  Reserve or explicitly exclude `completion-server` from extension dispatch. It is an internal hidden command invoked automatically when the shell requests completions, but it is not in the placeholder command registry when `buildExtensionManager()` snapshots reserved names. A valid `td-completion-server` extension can therefore be installed and will be spawned here on every TAB completion, inheriting the CLI environment (including any exported API token), instead of running `serverAction`. Add this internal name to the reserved set and/or handle the completion-server branch before extension dispatch.

doistbot (commented):
This PR wires installed extensions into `td --help` and pre-parse dispatch as `td <name> …`, adds a `td extension` management group, and consolidates the entry-point error reporting into a shared `reportFatal`. The argv boundary, built-in precedence, and error consolidation are solid overall; a few edge cases around dispatch and documentation are worth tightening.

Few things worth tightening:

- **Pre-parse edge cases:** the `--version`/`-V` short-circuit scans past the command token, so `td goals --version` skips discovery and the extension never runs — contrary to the pass-through contract. Also wrap the discovery/registration `await` in `reportFatal` so real directory errors (e.g., EACCES) render as formatted errors instead of unhandled rejections.

- **Reserved-name gaps in the snapshot:** the reserved-name set is taken after the `extension` placeholder is removed, so `td extension install` accepts names occupied by the group and its alias, and internal `completion-server` can be shadowed by an installed extension — spawning it on every TAB completion. Include those names in the reserved set or snapshot before the placeholder removal.

- **Passthrough contract for `extension exec`:** `td extension exec goals --user alice` loses its extension arguments because the later pre-parse `stripUserFlag` mutates argv before `argsAfter()` reads it. Limit stripping to arguments before the extension name (or preserve the original argv) so `exec` matches direct dispatch.

- **Duplication and test coverage:** direct-dispatch behavior now lives in both the pre-parse path and the pass-through handlers `registerExtensionPassThrough` attaches, which can drift — consolidate on one dispatcher. Relatedly, the core entry-point contract (`setHostArgvLength` to the command-token boundary, slicing args after the name) has no end-to-end test; add one or extract a pure helper to cover it.

- **Docs:** the new `td extension`/`td ext` command group needs to be added to `SKILL_CONTENT` in `src/lib/skills/content.ts`, then regenerate via `npm run sync:skill`.

<details>

- **<a href="#">P3</a> src/index.ts:261**: The dispatch branch ignores a root `--help`/`-h` that precedes the command token. `td --help goals` runs the `goals` extension, while `td --help task` prints td's help; the difference is the pre-parse dispatch. Check the host args before `commandTokenIndex` for help and defer to commander before dispatching (version before the token is already skipped by `needsExtensionLookup`, help is not).
- **<a href="#">P3</a> src/index.ts:206**: Update `CODEBASE.md` for the new `src/commands/extension/` command-group directory. The repository guidance requires map updates when a command group is added.
- **<a href="#">P3</a> src/index.ts:271**: `getRequestedUserRef() ?? (process.env[USER_ENV_VAR] || undefined)` duplicates the flag-then-`TD_USER` precedence logic already in `resolveActiveUser` (src/lib/auth.ts:84), including the `|| undefined` treatment of `TD_USER=`. Consider extracting a small exported helper (e.g. `resolveRequestedUserRef()` in `src/lib/auth.ts`) and using it in both places, so the precedence rules and their rationale live in one spot.
- **<a href="#">P3</a> src/lib/global-args.ts:81**: The new `hostArgvLength` doc block is inserted between `parseGlobalArgs`'s JSDoc and the function it documents, so the function's doc comment no longer sits immediately above it. Move the variable declaration (or the function JSDoc) so each doc block is adjacent to what it describes.
- **<a href="#">P3</a> src/index.ts:270**: This re-resolves an extension that `setUpExtensionDispatch` already discovered. `registerExtensionPassThrough` calls `manager.discover()` to build `extensions.names`, then this `extensions.dispatch(name, …)` routes through `manager.dispatch` → `findExtension` → `describe`, which re-reads that same extension's manifests, state, and (for git installs) `.git/config` on every `td <name>` invocation. Dispatch from the `Extension` object already in hand (or keep the discovered map) so the target extension is read once, not twice.
- **<a href="#">P3</a> src/commands/extension/index.test.ts:35**: This assertion adds no independent signal: `trustWarning` is built in `manager.ts` from `officialLabel`, which the previous line already pins to `'Todoist'`. It couples the test to the warning's wording without protecting the wording. Drop it, or assert the full warning text only if that exact user-facing string is itself a contract.
- **<a href="#">P3</a> src/commands/extension/index.test.ts:79**: The alias and group shape here are already covered by `src/lib/extensions/commands.test.ts` (`registerExtensionGroup`), and the exact subcommand list is registration wiring rather than behavior: adding or renaming a subcommand churns this assertion even though every subcommand's behavior is already tested there. Keep at most a one-line smoke check that this host wrapper registers the group with a real manager.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F537&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0fb80c90-add7-11f1-9941-68dc016524b4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0fb80c90-add7-11f1-9941-68dc016524b4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/index.ts
doistbot, line 264: <a href="#">P2</a>  This block reimplements the dispatch that `registerExtensionPassThrough` already attaches to each extension command (`argsAfter` + `manager.dispatch`). Since every direct `td <extension> …` invocation exits here before `program.parseAsync()`, those registered action handlers never run, so the same behavior now lives in two places and can drift. Consolidate on one path: have `setUpExtensionDispatch` register pass-through commands for help/completion only (no dispatch action) and keep this as the sole dispatcher, or pass the host's dispatch behavior into the pass-through registration.

src/index.ts
```diff
@@ -220,6 +245,44 @@ program.hook('preAction', (_thisCommand, actionCommand) => {
...
+    // Only the arguments before the name are td's. `--json` or `--user` after
+    // it belong to the extension, and must not change how td behaves or what
+    // account it passes on.
+    setHostArgvLength(commandTokenIndex)
```
doistbot: <a href="#">P2</a>  The entry-point contract this PR introduces — `td goals --json --user alice` passes `--json --user alice` to `goals` while td itself sees no global args — has no test. `setHostArgvLength` and `dispatchExtension` are tested in isolation, but not that `index.ts` sets the boundary to `commandTokenIndex` and passes `rawArgs.slice(commandTokenIndex + 1)`. A wrong index or slice (for example, including the extension name as arg 0) would pass the existing unit tests. Add a test that exercises this dispatch path, or extract a pure helper that computes the dispatch args/env and test that.

src/lib/command-token.ts
scottlovegrove: Real, and fixed in 7b440cc. `needsExtensionLookup` now takes the token index and only looks at arguments before it, so `td goals --version` reaches the extension while `td --version` still prints td's.

Verified against a built CLI: the first reports `args: ["--version"]` from the extension, the second prints `5.3.2`.

src/index.ts
scottlovegrove: Done in 7b440cc — the import and the `setUpExtensionDispatch` await are inside a `try`/`catch` that calls `reportFatal`, so an unreadable extensions directory is formatted like every other failure.

src/index.ts
scottlovegrove: Lands in #539, the last commit of this stack, which adds the `### Extensions` section and the Quick Reference line to `SKILL_CONTENT` and regenerates `SKILL.md`.

src/index.ts
scottlovegrove: Real, and fixed in 7b440cc. The pre-parse `--user` block is skipped entirely when the command token is the `extension` group, so argv reaches `exec` as typed.

Verified against a built CLI: `td extension exec hello --user alice --json` now reports `args: ["--user","alice","--json"]`.

src/commands/extension/index.ts
scottlovegrove: Right, and fixed in 7b440cc. `ALWAYS_RESERVED` names `extension`, `ext` and `completion-server` explicitly, so the snapshot no longer depends on what happens to be registered when the manager is built.

Verified: `td extension install ./td-extension` now fails with `EXTENSION_NAME_RESERVED`.

src/index.ts
scottlovegrove: This one was the worst of the set — thank you. Fixed in 7b440cc: `completion-server` is in `ALWAYS_RESERVED`, so it cannot be installed, and `registerExtensionPassThrough` now consults the manager's reserved names as well as the program's commands, so an extension placed in the directory by hand is not registered either.

Verified: `td extension install ./td-completion-server` fails with `EXTENSION_NAME_RESERVED`.

scottlovegrove: Agreed, and consolidated in 7b440cc. `registerExtensionPassThrough` now takes a `dispatch` function from the host, and both the registered action and the entry point go through it, so the account and the translated `TD_VERBOSE`/`TD_SPINNER` are applied whichever path runs. The entry point is down to one call.

src/index.ts
scottlovegrove: Done in 7b440cc. The env and account derivation is now `hostDispatchOptions(hostArgvLength)` in `src/commands/extension/dispatch.ts`, with its own tests: the boundary hiding `--json` and `--user` that come after the name, the flag beating `TD_USER`, an empty `TD_USER` naming nobody, and the verbosity and spinner translations.

The slice is covered by the same function taking the index, and the whole path is exercised end to end against a built CLI.

PR #538, opened 2026-09-11
feat(doctor): check installed extensions

scottlovegrove:
Adds a check per installed extension: the executable is there and can be run, no built-in command is hiding it, its manifests parse, the running td satisfies what it asks for, and its manifest format is one this version understands. Each says what to do about it.

Only problems are reported, after a single summary line, the way the Node version check stays quiet when there is nothing wrong. Someone with no extensions installed sees nothing at all.

The program is passed through to the check because which command names are taken is what decides whether an extension is shadowed, and that is only knowable from the program itself.

This gives `findManifestProblems` its first caller — it was written for exactly this and has been sitting unused.

### DIFF
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -20,15 +20,16 @@ import {
     isNewer,
 } from '../lib/update.js'
 import { getDefaultUserId, NoUserSelectedError } from '../lib/users.js'
+import { checkExtensions } from './extension/doctor.js'
 
-type CheckStatus = 'pass' | 'warn' | 'fail' | 'skip'
+export type CheckStatus = 'pass' | 'warn' | 'fail' | 'skip'
 
 interface DoctorOptions {
     json?: boolean
     offline?: boolean
 }
 
-interface DoctorCheck {
+export interface DoctorCheck {
     name: string
     status: CheckStatus
     message: string
@@ -355,18 +356,19 @@ async function checkStoredUsers(): Promise<DoctorCheck[]> {
     return checks
 }
 
-async function runDoctorChecks(options: DoctorOptions): Promise<DoctorCheck[]> {
+async function runDoctorChecks(options: DoctorOptions, program: Command): Promise<DoctorCheck[]> {
     return [
         checkNodeVersion(),
         await checkConfigFile(),
         ...(await checkStoredUsers()),
         await checkAuthentication(Boolean(options.offline)),
+        ...(await checkExtensions(program)),
         await checkForUpdates(Boolean(options.offline)),
     ].filter((check): check is DoctorCheck => check !== null)
 }
 
-export async function doctorAction(options: DoctorOptions): Promise<void> {
-    const checks = await runDoctorChecks(options)
+export async function doctorAction(options: DoctorOptions, program: Command): Promise<void> {
+    const checks = await runDoctorChecks(options, program)
     const summary = summarize(checks)
     const ok = summary.failed === 0
 
@@ -392,5 +394,8 @@ export function registerDoctorCommand(program: Command): void {
         .description('Diagnose common CLI setup and environment issues')
         .option('--json', 'Output diagnostic results as JSON')
         .option('--offline', 'Skip network checks against Todoist and npm')
-        .action(doctorAction)
+        // The program is passed on so the extension checks can see which
+        // command names are taken, which is what decides whether an extension
+        // is being shadowed.
+        .action((options: DoctorOptions) => doctorAction(options, program))
 }
--- a/src/commands/extension/doctor.ts
+++ b/src/commands/extension/doctor.ts
@@ -1,0 +1,114 @@
+/**
+ * What `td doctor` reports about installed extensions.
+ *
+ * Nothing here reaches the network: `list` is a directory read plus, for a
+ * clone, a `git rev-parse`. Someone with no extensions sees nothing at all.
+ */
+
+import type { Command } from 'commander'
+import packageJson from '../../../package.json' with { type: 'json' }
+import { findManifestProblems } from '../../lib/extensions/manifest.js'
+import { satisfiesRange } from '../../lib/extensions/version-range.js'
+import type { DoctorCheck } from '../doctor.js'
+import { buildExtensionManager } from './index.js'
+
+const NAME = 'extensions'
+const TD_VERSION = packageJson.version
+
+/**
+ * Windows paths are implemented but have not had a full test pass, so say so
+ * rather than letting someone discover it themselves.
+ */
+function windowsCheck(): DoctorCheck | null {
+    if (process.platform !== 'win32') return null
+    return {
+        name: NAME,
+        status: 'warn',
+        message: 'Extensions are experimental on Windows in this release',
+        details: { platform: process.platform },
+    }
+}
+
+export async function checkExtensions(program: Command): Promise<DoctorCheck[]> {
+    const manager = buildExtensionManager(program)
+
+    if (await manager.isEmpty()) return []
+
+    let listing: Awaited<ReturnType<typeof manager.list>>
+    try {
+        listing = await manager.list()
+    } catch (error) {
+        return [
+            {
+                name: NAME,
+                status: 'fail',
+                message: `Cannot read ${manager.extensionsDir}: ${error instanceof Error ? error.message : String(error)}`,
+                details: { extensionsDir: manager.extensionsDir },
+            },
+        ]
+    }
+
+    const checks: DoctorCheck[] = [
+        {
+            name: NAME,
+            status: 'pass',
+            message: `${listing.length} extension(s) installed in ${manager.extensionsDir}`,
+            details: { count: listing.length, extensionsDir: manager.extensionsDir },
+        },
+    ]
+
+    for (const entry of listing) {
+        const details = { name: entry.name, kind: entry.kind, dir: entry.dir }
+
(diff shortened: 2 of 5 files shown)


src/commands/extension/doctor.ts
```diff
@@ -0,0 +1,114 @@
...
+        status: 'warn',
+        message: 'Extensions are experimental on Windows in this release',
+        details: { platform: process.platform },
+    }
+}
+
+export async function checkExtensions(program: Command): Promise<DoctorCheck[]> {
+    const manager = buildExtensionManager(program)
+
+    if (await manager.isEmpty()) return []
```
doistbot: <a href="#">P2</a>  `isEmpty()` catches every `readdir` error and returns `true`, so a permissions error (EACCES), `ENOTDIR`, or a broken mount on the extensions directory makes `td doctor` silently report nothing as if no extensions were installed. The `try`/`catch` around `manager.list()` below is then unreachable for exactly the failures it was written to surface. Call `manager.list()` first and return `[]` only when `listing.length === 0` (a missing directory already yields `[]` from `discoverExtensions`), so genuine read failures still reach the fail branch.

src/commands/extension/doctor.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (entry.shadowed) {
+            checks.push({
+                name: NAME,
+                status: 'warn',
+                message: `${entry.name} is hidden by the built-in command of the same name. Run \`td extension exec ${entry.name}\`, or reinstall it under another name`,
+                details,
+            })
+        }
+
+        for (const problem of await findManifestProblems(entry.dir, manager.binName)) {
```
doistbot: <a href="#">P2</a>  Avoid unbounded reads of extension-controlled manifest paths here. A git extension can commit `.td-manifest.json` as a symlink to `/dev/zero` (or a FIFO) without executing any extension code; this first caller of `findManifestProblems` passes it to `readFile`, which never reaches EOF and can hang or exhaust memory when the user runs `td doctor`. Reject non-regular files (including symlinks) and enforce a manifest size limit before reading.

doistbot: <a href="#">P2</a>  These manifest checks run serially for every installed extension, even though each performs two independent filesystem reads. `manager.list()` already uses bounded concurrency for its per-extension I/O; collect `findManifestProblems` with bounded concurrency before emitting checks so a user with many extensions does not pay one manifest-read latency per extension.

src/commands/doctor.ts
```diff
@@ -355,18 +356,19 @@ async function checkStoredUsers(): Promise<DoctorCheck[]> {
...
 }
 
-async function runDoctorChecks(options: DoctorOptions): Promise<DoctorCheck[]> {
+async function runDoctorChecks(options: DoctorOptions, program: Command): Promise<DoctorCheck[]> {
     return [
         checkNodeVersion(),
         await checkConfigFile(),
         ...(await checkStoredUsers()),
         await checkAuthentication(Boolean(options.offline)),
+        ...(await checkExtensions(program)),
```
doistbot, line 365: <a href="#">P2</a>  This wiring makes the existing doctor.test.ts read the real user's extensions directory: those tests run `program.parseAsync(['node','td','doctor'])` without stubbing XDG_DATA_HOME, so on any machine with extensions installed `checkExtensions` adds real checks (and may shell out to `git rev-parse` for clone extensions). `supports json output and offline mode` asserts `parsed.summary.passed === 0` and `ok === true`, which a healthy installed extension breaks, and a broken one flips `ok` to false. Stub XDG_DATA_HOME/XDG_STATE_HOME to a temp dir in doctor.test.ts (the way the new extension/doctor tests do), or mock `checkExtensions` there.

doistbot (commented):
This PR wires a per-extension check into `td doctor` — verifying each installed extension's executable, manifest, shadowing, compatibility, and manifest format, reporting only problems after a single summary line, and finally giving `findManifestProblems` its first caller.

Few things worth tightening:
- The empty-directory short-circuit swallows `readdir` errors (EACCES, ENOTDIR, broken mounts) and silently reports success — call `manager.list()` first and only return `[]` on a genuinely empty listing, so the fail branch can actually surface real filesystem failures.
- Guard against unbounded reads of extension-controlled manifest paths: reject symlinks/non-regular files (e.g. a manifest symlinked to `/dev/zero`) and enforce a size limit before `readFile`, otherwise `td doctor` can hang on untrusted input.
- Run the manifest checks with bounded concurrency (as `manager.list()` already does) instead of serially per extension, so users with many extensions don't pay one read latency each.
- Stub `XDG_DATA_HOME`/`XDG_STATE_HOME` (or mock `checkExtensions`) in doctor.test.ts, since the existing tests currently read the real user's extensions directory and a healthy or broken installed extension can flip the `passed === 0`/`ok` assertions.

<details>

- **<a href="#">P3</a> src/commands/extension/doctor.ts:91**: This re-implements the `requires` compatibility decision that already exists in `ExtensionManager.warnIfIncompatible` (src/lib/extensions/manager.ts): the same `entry.requires?.[binName]` lookup plus `satisfiesRange(version, range)` predicate. The manager already has `version` and `satisfiesRange` in scope, so consider exposing the result on `ExtensionListing` (e.g. `requiresIncompatible`, computed in `list()`) or adding a manager helper, and have `checkExtensions` only format the message. That keeps one source of truth for the decision and removes the duplicated `packageJson`/`satisfiesRange` imports here.
- **<a href="#">P3</a> src/commands/extension/doctor.ts:81**: `manager.list()` has already read each extension's `td-extension.json` (and `.td-manifest.json` for binary installs) during its concurrent discovery pass. This loop then calls `findManifestProblems`, which re-reads those same files and awaits them one extension at a time. Surface the read/parse status from `list()` (or run the problem scan in the same concurrent pass) instead of reading the manifests twice.
- **<a href="#">P3</a> src/commands/extension/doctor.test.ts:125**: This pins the exact problem order (`fail` before `warn`), which is not what the test is about. Reordering the per-extension checks would fail this test even though the same three checks are still reported. Keep the length and summary-first assertions, then assert the remaining statuses as a set (e.g., sorted `['fail', 'warn']`) instead of an ordered array.
- **<a href="#">P3</a> src/commands/extension/doctor.ts:40**: The `manager.list()` throw path has no test: nothing covers that an unreadable extensions directory produces a FAIL check instead of crashing `td doctor`. A test that chmods the directory 0o000 (with the running-as-root guard the repo already uses in manifest.test.ts) would pin that contract.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F538&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A12afa1b0-add7-11f1-92d9-cf56b3fc5859%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A12afa1b0-add7-11f1-92d9-cf56b3fc5859&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/extension/doctor.ts
scottlovegrove: Right, and fixed in f526b6c. `list()` runs first and the check returns `[]` only when the list is genuinely empty, so a directory that cannot be read reaches the fail branch. There is a test that chmods the directory to `0o000` and asserts the fail check, skipped when running as root since the mode means nothing there.

src/commands/extension/doctor.ts
scottlovegrove: Done in f526b6c, in `readJsonFile` rather than at this call site, so it covers `list` too. The file is opened and checked before reading: anything that is not a regular file is reported as unreadable, as is anything over 1 MiB. A test symlinks `td-extension.json` to `/dev/zero` and asserts the warning rather than hanging.

scottlovegrove: Done in f526b6c — the manifest reads now go through `mapWithConcurrency` at 4, matching what `list()` already does for its per-extension work.

src/commands/doctor.ts
scottlovegrove: Correct, and worse than it looks: those tests were only passing because the `node:fs/promises` mock has just `readFile`, so discovery failed and `isEmpty` swallowed it. Once the fix above made read failures visible, they broke.

f526b6c mocks `./extension/doctor.js` in `doctor.test.ts` so those tests stay about doctor, and the real behaviour keeps its own tests in `src/commands/extension/doctor.test.ts`, which use a temp directory.

PR #539, opened 2026-09-11
docs: document td extension

scottlovegrove:
Adds an Extensions section to SKILL_CONTENT, with the two rules an agent has to know: everything after the extension name is forwarded verbatim, and td exits with the extension's exit code. Regenerates SKILL.md.

CODEBASE.md gains the argv token scan, the discovery gate and the pre-parse dispatch in the architecture flow, the pass-through command pattern beside the flat and group ones, and the new lib modules.

The spec catches up with what was built. `passThroughOptions` is recorded as unusable, with the reason, so nobody reaches for it again. Discovery is described as gated rather than unconditional. The manifest examples carry `manifestVersion`, and the paragraph claiming the extensions directory depends only on Node built-ins and cli-core is corrected — zod is a dependency, behind a dynamic import. Telemetry moves from the open question it was to a decision: there is no outbound request to tag when td spawns a child and waits, and the nested td calls an extension makes already report their own command paths.

## Trying it out

This is the top of the stack, so this branch has the whole feature. None of the earlier PRs is usable on its own.

```bash
git checkout feat/ext-cmd-9-docs
npm ci && npm run build && npm link
td extension install scottlovegrove/td-hello
```

Then:

```bash
td hello                              # runs it
td hello Scott --json                 # arguments pass through untouched
td hello --help                       # the extension answers, not td
td --help                             # lists it under "Extensions:"
td extension list                     # NAME / SOURCE / VERSION / KIND / OFFICIAL
td extension upgrade --all --dry-run
td doctor                             # reports on what is installed
td extension remove hello
```

`--user` before the name is td's and reaches the extension as `TD_USER`; after the name it belongs to the extension:

```bash
td --user you@example.com hello --json   # "actingAs": "you@example.com"
td hello --json --user someone           # "actingAs": null, flag passed through
TD_USER=you@example.com td hello --json  # "actingAs": "you@example.com"
```

Two caveats. `npm link` replaces your installed `td`, and the install writes to your real `~/.local/share/todoist-cli/extensions`. To keep it contained, skip the link and point the directories somewhere scratch:

```bash
XDG_DATA_HOME=/tmp/td-ext/data XDG_STATE_HOME=/tmp/td-ext/state \
  node dist/index.js extension install scottlovegrove/td-hello
```

Undo with `td extension remove hello`, then `npm unlink -g @doist/todoist-cli && npm i -g @doist/todoist-cli`.

`scottlovegrove/td-hello` ships no release assets, so it installs as a git clone — which is why `list` shows a short SHA rather than a tag. The release-binary and checksum paths have unit tests but no extension publishing real assets to try them against yet.

### DIFF
--- a/skills/todoist-cli/SKILL.md
+++ b/skills/todoist-cli/SKILL.md
@@ -28,7 +28,7 @@ metadata:
 - Mutating commands support `--dry-run` to preview actions without executing them.
 - Destructive commands typically require `--yes`.
 - `--quiet` / `-q` suppresses success messages. Create commands still print the bare ID for scripting (e.g. `id=$(td task add "Buy milk" --quiet)`).
-- Global flags: `--no-spinner`, `--progress-jsonl`, `-v/--verbose`, `--accessible`, `--quiet`, `--user <id|email>`.
+- Global flags: `--no-spinner`, `--progress-jsonl`, `-v/--verbose`, `--accessible`, `--quiet`, `--user <id|email>`. `TD_USER` is read as the fallback for `--user`.
 
 ## Authentication
 
@@ -82,12 +82,13 @@ td accounts use <id|email>             # set the default account (alias: td acco
 td accounts current                    # show the active account (--json/--ndjson supported)
 td accounts remove <id|email>          # delete an account and its token (--json/--ndjson supported)
 td --user <id|email> task list         # one-off override for any command
+TD_USER=<id|email> td task list        # same, for every nested call in a script
 td auth logout --user <id|email>       # log out a specific account
 ```
 
 `td accounts` is also available as `td user` / `td users` (back-compat aliases).
 
-Resolution order: `--user <ref>` > `user.defaultUser` from config > the only stored account. With multiple accounts and no default, commands error and ask for `--user` (or `td accounts use`). `<ref>` matches an exact id or email (case-insensitive on email). `TODOIST_API_TOKEN` still bypasses the resolver entirely.
+Resolution order: `--user <ref>` > `TD_USER` > `user.defaultUser` from config > the only stored account. With multiple accounts and no default, commands error and ask for `--user` (or `td accounts use`). `<ref>` matches an exact id or email (case-insensitive on email). `TODOIST_API_TOKEN` still bypasses the resolver entirely.
 
 ## Quick Reference
 
@@ -100,6 +101,7 @@ Resolution order: `--user <ref>` > `user.defaultUser` from config > the only sto
 - Templates and files: `td template ...`, `td attachment view <file-url>`, `td backup ...`
 - Help Center: `td hc locales/search/view`
 - Account and tooling: `td stats`, `td settings ...`, `td config view`, `td accounts ...`, `td completion ...`, `td view <todoist-url>`, `td doctor`, `td update`, `td changelog`
+- Extensions: `td extension install/list/upgrade/remove/exec` (alias: `td ext`)
 - Developer apps: `td apps list/view` (requires `td auth login --additional-scopes=app-management`)
 - Backups: `td backup list/download` (requires `td auth login --additional-scopes=backups`)
 - Billing: `td billing subscription/plan/prices/pricing` (requires `td auth login --additional-scopes=billing`)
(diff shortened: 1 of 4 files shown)


skills/todoist-cli/SKILL.md
```diff
@@ -392,6 +394,27 @@ The `billing` command surface is **read-only** and requires the `billing` OAuth
...
+td extension install ./td-scratch      # a local directory, symlinked, for development
+td extension list                      # --json for the full record of each one
+td extension upgrade --all             # --dry-run to preview; --force to pass a pin
+td extension remove goals              # --force for a clone with uncommitted changes
+td extension exec goals list --json    # bypass the built-in commands entirely
+```
+
+Two rules matter when calling one:
+
+- Everything after the extension name is passed through untouched. `td goals --json` is `--json` for `goals`, not for td. Global flags only apply before the name: `td --user alice goals list`.
```
doistbot, line 413: <a href="#">P2</a>  Preserve `--help` for `extension exec`. This rule reads as applying to the preceding `td extension exec goals ...` form, but `exec` keeps Commander’s default help option. `td extension exec task --help` therefore renders td’s `exec` help instead of forwarding `--help` to the shadowed extension—the only invocation form that can reach it. Add `.helpOption(false)` to `exec` (as on per-extension commands), or narrow this documentation.

doistbot (commented):
Docs-only PR that catches CODEBASE.md, the extensions spec, SKILL.md, and SKILL_CONTENT up with the implemented extension system — the passthrough/dispatch claims, manifest examples, zod note, and telemetry decision all check out against the code.

Few things worth tightening:
- The "everything after the extension name is forwarded verbatim" rule has one real gap: `td extension exec task --help` renders td's `exec` help instead of reaching the shadowed extension, since `exec` still keeps Commander's default help option. Either add `.helpOption(false)` to `exec` (as the per-extension commands already do) or narrow the doc to match.

<details>

- **<a href="#">P3</a> CODEBASE.md:74**: This step says extensions are discovered whenever the token does not name a built-in, but the actual gate in `needsExtensionLookup()` also skips discovery for `td --version` / `td -V`. For those invocations there is no command token, so this wording would incorrectly imply the extension system is loaded. Add the `--version` exclusion here (the spec's Startup cost section already states it correctly).

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F539&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A16510b60-add7-11f1-89fb-3dce96088827%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A16510b60-add7-11f1-89fb-3dce96088827&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

skills/todoist-cli/SKILL.md
scottlovegrove: Fixed the code rather than narrowing the documentation — c9d22eb on #533 adds `.helpOption(false)` to `exec`, so `td extension exec task --help` now forwards `--help` to the extension. The rule as written holds.

nvignola:
✅

PR #541, opened 2026-09-11
fix(extensions): stop a manifest range from breaking the extension

scottlovegrove:
Two findings from post-merge review of #521 and #527. Both reproduce against `next`; the first is a real breakage rather than a nit.

## `satisfiesRange` throws, and that is fatal

`version-range.ts` documents that anything it cannot parse counts as satisfied, because the check only ever drives a warning. That did not hold. The comparator pattern accepts any version starting with a digit, so `1.x` and `>=1.2.3.4` build a comparator that `parseVersion` then refuses, and the exception escapes.

`warnIfIncompatible` runs before every dispatch and again in `doctor`, so an extension whose manifest says `"requires": { "td": "1.x" }` is simply unrunnable, and it takes `doctor` down with it for everyone who has it installed:

```
$ td wild                  → Invalid version string: '1.x'    exit 1
$ td doctor --offline      → Invalid version string: '1.x'    exit 1   (no checks, no summary)
```

Evaluating a range is now wrapped, and anything the version parser refuses counts as satisfied. Ranges it can read are unaffected — `>=1.2` is still `true`, `^1.2` still `false`, `~2` still `false` — so no correct warning is lost.

After:

```
$ td wild                  → wild ran                          exit 0
$ td doctor --offline      → Doctor summary: 3 passed, 1 warnings, 1 skipped
```

## An empty authored manifest discarded the `package.json` metadata

A `td-extension.json` written for a later manifest format parses to `{}`, because every key it carries is one this version strips. That counted as the dedicated file winning, so the metadata the same author wrote under the `td` key of `package.json` — in a form this version *can* read — was thrown away.

The dedicated file is now taken only when it says something. This needed more than a guard: `package.json` was previously read only when the dedicated file was absent, so the fallback had nothing to fall back to. It is now read when the dedicated file yields nothing, and not eagerly otherwise.

```
before:  Extensions:  fb   Extension fb
after:   Extensions:  fb   Description from package.json
```

## Notes

The report on #527 listed `>=1.2`, `^1.2` and `~2` as throwing and `>=1.2.3.4` as silently truncating. Against `@doist/cli-core@1.4.0` the first three evaluate fine and the fourth throws — that version tolerates a two-part version but not a wildcard or a fourth part. The conclusion stands regardless; only the specific inputs differ.

Tests cover both: six range shapes that previously threw, the two-part forms that must keep evaluating, and the two manifest fallback cases.

### DIFF
--- a/src/lib/extensions/manifest.ts
+++ b/src/lib/extensions/manifest.ts
@@ -26,31 +26,50 @@ export async function pickAuthoredFields(value: unknown): Promise<AuthoredManife
 }
 
 /**
+ * The fields that carry meaning. `manifestVersion` is deliberately not one of
+ * them: it says which format a file is written in, not what the file has to
+ * say, and a file carrying nothing else has told this version of the CLI
+ * nothing it can use.
+ */
+const CONTENT_FIELDS = ['description', 'requires', 'completion'] as const
+
+function saysSomething(manifest: AuthoredManifest | undefined): boolean {
+    return manifest !== undefined && CONTENT_FIELDS.some((field) => manifest[field] !== undefined)
+}
+
+/**
  * Author metadata for an extension, from `<bin>-extension.json` or, for Node
  * extensions that would rather not carry a second file, the `<bin>` key of
- * `package.json`. The dedicated file wins when both exist.
+ * `package.json`.
+ *
+ * The dedicated file wins whenever it says anything this version understands.
+ * When it does not — the shape a file written for a later format takes, since
+ * every key it carries but the version marker is stripped here — the
+ * `package.json` metadata is used instead, rather than being shadowed by a
+ * file that turned out to be empty. The version marker survives that fall
+ * back, so `list` and `doctor` still report that some of the metadata is from
+ * a format this version cannot read.
  */
 export async function readAuthoredManifest(
     dir: string,
     binName: string,
 ): Promise<AuthoredManifest | undefined> {
     const dedicatedValue = await readJsonValue(join(dir, authoredManifestFileName(binName)))
+    const dedicated =
+        dedicatedValue === undefined ? undefined : await pickAuthoredFields(dedicatedValue)
 
-    if (dedicatedValue !== undefined) {
-        const dedicated = (await schemas()).parseAuthoredManifest(dedicatedValue)
-        // Taken only when it actually says something. A file written for a
-        // later format parses to an empty object here, because every key it
-        // carries is one this version strips, and letting that win would throw
-        // away the `package.json` metadata the same author wrote in a form
-        // this version can read.
-        if (dedicated && Object.keys(dedicated).length > 0) return dedicated
-    }
+    if (saysSomething(dedicated)) return dedicated
 
     const packageJson = await readJsonValue(join(dir, 'package.json'))
-    // Nothing to validate, so nothing to load a validator for.
-    if (!isRecord(packageJson)) return undefined
+    // Nothing left to validate, so nothing to load a validator for.
+    if (!isRecord(packageJson)) return dedicated
 
-    return (await schemas()).parseAuthoredManifest(packageJson[binName])
+    const fromPackage = await pickAuthoredFields(packageJson[binName])
+    if (!saysSomething(fromPackage)) return dedicated ?? fromPackage
+
+    return dedicated?.manifestVersion === undefined
+        ? fromPackage
+        : { ...fromPackage, manifestVersion: dedicated.manifestVersion }
 }
 
 /**
--- a/src/lib/extensions/manifest.ts
+++ b/src/lib/extensions/manifest.ts
@@ -35,18 +35,22 @@ export async function readAuthoredManifest(
     binName: string,
 ): Promise<AuthoredManifest | undefined> {
     const dedicatedValue = await readJsonValue(join(dir, authoredManifestFileName(binName)))
-    const packageJson =
-        dedicatedValue === undefined ? await readJsonValue(join(dir, 'package.json')) : undefined
 
-    // Nothing to validate, so nothing to load a validator for.
-    if (dedicatedValue === undefined && !isRecord(packageJson)) return undefined
-
-    const { parseAuthoredManifest } = await schemas()
     if (dedicatedValue !== undefined) {
-        const dedicated = parseAuthoredManifest(dedicatedValue)
-        if (dedicated) return dedicated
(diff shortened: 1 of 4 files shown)


src/lib/extensions/manifest.ts
```diff
@@ -35,18 +35,22 @@ export async function readAuthoredManifest(
...
     if (dedicatedValue !== undefined) {
-        const dedicated = parseAuthoredManifest(dedicatedValue)
-        if (dedicated) return dedicated
+        const dedicated = (await schemas()).parseAuthoredManifest(dedicatedValue)
+        // Taken only when it actually says something. A file written for a
+        // later format parses to an empty object here, because every key it
+        // carries is one this version strips, and letting that win would throw
+        // away the `package.json` metadata the same author wrote in a form
+        // this version can read.
+        if (dedicated && Object.keys(dedicated).length > 0) return dedicated
```
doistbot: <a href="#">P2</a>  The `Object.keys(dedicated).length > 0` check still discards `package.json` metadata when a later-format `td-extension.json` declares `manifestVersion`. `authoredManifestSchema` keeps that field rather than stripping it, so `{ "manifestVersion": 2 }` parses to a non-empty object and shadows package.json — yet that field is exactly how a future format is marked, so the fallback misses its main scenario. Gate the check on authored fields (`description`/`requires`/`completion`) or merge the package.json fields while retaining the version marker.

doistbot (commented):
This PR fixes two post-merge extension breakages: `satisfiesRange` now degrades to "satisfied" on unparseable version ranges (restoring its documented never-throw contract) instead of crashing dispatch and `doctor`, and an authored `td-extension.json` that strips to nothing no longer shadows usable `package.json` metadata, which is now read lazily only as a fallback.

Few things worth tightening:
- The emptiness check still lets a later-format manifest shadow `package.json` when it carries only `manifestVersion` (kept by `authoredManifestSchema`), so the fallback misses its main scenario — consider gating on authored fields like `description`/`requires`/`completion`, or merging package.json fields while preserving the version marker.

<details>

- **<a href="#">P3</a> src/lib/extensions/manifest.ts:46**: The docstring above `readAuthoredManifest` still says "The dedicated file wins when both exist" — no longer true, since the dedicated file now only wins when it yields at least one readable field. Worth updating so the contract statement matches the new fallback behavior.
- **<a href="#">P3</a> src/lib/extensions/manifest.ts:40**: Reuse the existing `pickAuthoredFields` helper from this same module for both parse sites (here and line 53) instead of inlining `(await schemas()).parseAuthoredManifest(...)`. It is the module's dedicated wrapper for exactly this operation and would keep authored-manifest parsing on a single code path.
- **<a href="#">P3</a> src/lib/extensions/version-range.ts:71**: This comment says `parseVersion` rejects anything that isn't a full `x.y.z`, but the new `^1.2`/`~2` cases (and `>=1.2`) only pass because it accepts two-part and one-part versions. Rephrase to say it rejects shapes such as wildcards and four-part versions.
- **<a href="#">P3</a> src/lib/extensions/version-range.test.ts:68**: The six ranges run inside one `for` loop, so the first failing input aborts the remaining assertions and the report only points at the shared `expect`, not the offending range. Use `it.each` (already used earlier in this file) so each range is an independently named case and a failure identifies its input.
- **<a href="#">P3</a> src/lib/extensions/manifest.test.ts:149**: The empty-object case at L149 is redundant with the later-format case at L131: both drive the same path — the dedicated file parses to a value with no keys, so `Object.keys(dedicated).length > 0` is false and the fallback runs. If that guard broke, both tests fail identically. Folding the two into one `it.each` over file contents ('a file written for a later format', 'an empty object') keeps the contract covered with half the fixture duplication.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F541&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Add45a000-adf0-11f1-857c-2626eb8c170f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Add45a000-adf0-11f1-857c-2626eb8c170f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/manifest.ts
scottlovegrove: Right, and the fix missed its own main case — thank you. Reproduced before changing anything:

```
td-extension.json:  {"manifestVersion": 2, "somethingFromTheFuture": true}
package.json:       {"td": {"description": "Description from package.json"}}

before:  Extensions:  mv   Extension mv
```

Fixed in 969f860 by gating on what the file says — `description`, `requires`, `completion` — rather than on key count. A version marker is not something to say: it records which format the file is written in, not what it has to tell this version of the CLI.

Took the merge option for the rest of your suggestion, since the two signals are independent and both worth keeping. The marker survives the fall back, so an author with a later-format manifest and a `package.json` description gets both:

```
after:   Extensions:  mv   Description from package.json
         td extension list →  mv   binary · manifest needs a newer td
```

The five follow-up notes are all in the same commit: the docstring no longer claims the dedicated file always wins, both parse sites go through `pickAuthoredFields`, the `parseVersion` comment now says wildcards and four-part versions rather than "not a full x.y.z" (two-part versions do parse, which is why `^1.2` still evaluates), and both the range and fallback tests use `it.each` so a failure names its input. The empty-object and later-format cases are folded into one `it.each`, with the `manifestVersion` variants as separate cases since they now take a different branch.

gnapse (approved):

PR #542, opened 2026-09-11
feat(extensions): scaffold a new extension with `td extension create`

scottlovegrove:
First half of `td extension create`, the phase 2 scaffolding command. `bash` and `node` templates are here; `compiled`, with the release workflow, follows separately.

## What it does

```
$ td extension create goals
Created td-goals from the bash template:
  README.md
  td-goals
  td-extension.json

Next:
  cd td-goals
  td extension install .   # links it here, so every edit is live
  td goals

Publish it as a repository named td-goals with the `td-extension` topic.
```

`--template bash|node`, `--description <text>`, `--json`.

## The bar

This works with nothing edited, for both templates — verified end to end against a build:

```bash
td extension create goals
cd td-goals
td extension install .
td goals          # Hello from td-goals, running under td 5.3.2.
td goals --json   # {"extension":"goals","hostVersion":"5.3.2"}
td goals --help   # the extension's own usage, not td's
```

That is why neither template calls the API. They demonstrate calling td back through `TD_NODE`/`TD_PATH` — the part worth teaching, since it is what works on Windows, where the `td` on PATH is a `.cmd` shim that cannot be spawned directly — but against `td --version`, so a fresh scaffold runs before anyone has logged in. The line that fetches real data is there, commented, one edit away.

## Templates are real files

Under `templates/extension/<kind>/`, so they can be read, linted and run as what they are rather than escaped inside template literals.

They ship through the `files` field and resolve with `new URL('../../../templates/extension/', import.meta.url)`, which lands on the same directory from `src/commands/extension/` under vitest and from `dist/commands/extension/` in an install — so there is no copy step in the build that can quietly not happen. `npm pack --dry-run` confirms all eight files are in the tarball.

Two names cannot be taken literally from the template directory: the executable is named after the extension, and npm rewrites a `.gitignore` inside a published package, so it travels as `gitignore` and is restored on the way out.

A placeholder nothing fills is refused rather than written through, so a typo in a template fails a test here instead of landing in someone's new repository. One test scaffolds each shipped template and asserts no `{{…}}` survives anywhere.

## Notes

- Names go through the same `validateExtensionName` and reserved-name check as `install`, so a name that could never be installed is never scaffolded.
- `create` is only registered when the host supplies `templatesDir`, so a CLI adopting `src/lib/extensions/` without templates simply does not offer it.
- `ExtensionManager` gains `version`, which the manifest's starting `requires` range needs. `PassThroughOptions` is renamed `ExtensionCommandOptions` now that it carries more than `dispatch` — no external consumers.

### DIFF
--- a/src/commands/extension/index.test.ts
+++ b/src/commands/extension/index.test.ts
@@ -4,8 +4,9 @@ import { join } from 'node:path'
 import { captureConsole, createTestProgram } from '@doist/cli-core/testing'
 import { Command } from 'commander'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { listTemplates } from '../../lib/extensions/create.js'
 import { writeFixtureExtension } from '../../test-support/extension-fixture.js'
-import { buildExtensionManager, registerExtensionCommand } from './index.js'
+import { buildExtensionManager, registerExtensionCommand, templatesDir } from './index.js'
 
 describe('buildExtensionManager', () => {
     let root: string
@@ -63,6 +64,15 @@ describe('buildExtensionManager', () => {
 
         const [entry] = await manager.list()
         expect(entry.shadowed).toBe(false)
+    })
+})
+
+describe('templatesDir', () => {
+    it('resolves to the templates the package actually ships', async () => {
+        // The path is relative to this module's own location, so moving the
+        // command file would break `extension create` at runtime while every
+        // test that recomputes the URL itself still passed.
+        await expect(listTemplates(templatesDir())).resolves.toEqual(['bash', 'node'])
     })
 })
 
--- a/src/commands/extension/index.test.ts
+++ b/src/commands/extension/index.test.ts
@@ -77,6 +77,7 @@ describe('registerExtensionCommand', () => {
         expect(group).toBeDefined()
         expect(group?.aliases()).toContain('ext')
         expect(group?.commands.map((command) => command.name()).sort()).toEqual([
+            'create',
             'exec',
             'install',
             'list',
--- a/src/lib/extensions/create.ts
+++ b/src/lib/extensions/create.ts
@@ -1,0 +1,164 @@
+/**
+ * Scaffolding a new extension.
+ *
+ * The templates are real files rather than strings in this module, so they can
+ * be read, linted and run as what they are. They belong to the host — a CLI
+ * adopting this directory brings its own — and arrive through `templatesDir`.
+ */
+
+import { chmod, mkdir, readdir, readFile, writeFile } from 'node:fs/promises'
+import { dirname, join } from 'node:path'
+import { CliError } from '@doist/cli-core'
+import { exists } from './fs-utils.js'
+import { toCommandName, toDirName, validateExtensionName } from './source.js'
+
+export type CreateOptions = {
+    /** Template to scaffold from. Defaults to the first one available. */
+    template?: string
+    /** Where to create the directory. Defaults to the working directory. */
+    directory?: string
+    description?: string
+}
+
+export type CreateResult = {
+    name: string
+    dirName: string
+    dir: string
+    template: string
+    /** Paths written, relative to the new directory. */
+    files: string[]
+}
+
+export type CreateContext = {
+    binName: string
+    /** Host version, for the `requires` range the manifest starts with. */
+    version: string
+    /** Directory holding one subdirectory per template. */
+    templatesDir: string
+    reservedNames: () => Iterable<string>
+}
+
+/**
+ * Two names cannot be taken literally from the template directory.
+ *
+ * The executable is named after the extension, which is only known here. And
+ * npm rewrites a `.gitignore` inside a published package, so it is carried
+ * under a name npm leaves alone and restored on the way out.
+ */
+function targetName(templateFile: string, dirName: string): string {
+    if (templateFile === 'executable') return dirName
+    if (templateFile === 'gitignore') return '.gitignore'
+    return templateFile
+}
+
+export async function listTemplates(templatesDir: string): Promise<string[]> {
+    const entries = await readdir(templatesDir, { withFileTypes: true }).catch(() => [])
+    return entries
+        .filter((entry) => entry.isDirectory())
+        .map((entry) => entry.name)
+        .sort()
+}
+
+function fill(content: string, values: Record<string, string>): string {
+    const filled = content.replace(/\{\{(\w+)\}\}/g, (whole, key: string) =>
+        key in values ? values[key] : whole,
+    )
+
+    // A placeholder nothing replaced is a typo in the template, and shipping it
+    // verbatim into someone's new repository is worse than refusing.
+    const leftover = filled.match(/\{\{\w+\}\}/)
+    if (leftover) {
+        throw new CliError(
+            'EXTENSION_TEMPLATE_INVALID',
+            `The template refers to ${leftover[0]}, which is not a value this command supplies.`,
+        )
+    }
+    return filled
+}
+
+export async function createExtension(
+    rawName: string,
+    options: CreateOptions,
+    context: CreateContext,
+): Promise<CreateResult> {
+    const { binName } = context
+
+    // Accept `goals` or `td-goals`, and hold both to the rules install uses, so
+    // a name that could never be installed is never scaffolded either.
+    const dirName = toDirName(binName, toCommandName(binName, rawName))
+    validateExtensionName(binName, dirName)
+
+    const name = toCommandName(binName, dirName)
+    if (new Set(context.reservedNames()).has(name)) {
+        throw new CliError(
+            'EXTENSION_NAME_RESERVED',
+            `"${name}" is the name of a built-in ${binName} command.`,
+            {
+                hints: [
+                    `An extension called ${name} could never be run as \`${binName} ${name}\`.`,
+                ],
+            },
+        )
+    }
+
+    const available = await listTemplates(context.templatesDir)
+    if (available.length === 0) {
+        throw new CliError(
+            'EXTENSION_TEMPLATE_INVALID',
+            `No templates found in ${context.templatesDir}.`,
+        )
+    }
+
+    const template = options.template ?? available[0]
+    if (!available.includes(template)) {
+        throw new CliError('EXTENSION_TEMPLATE_INVALID', `There is no "${template}" template.`, {
+            hints: [`Available templates: ${available.join(', ')}.`],
+        })
+    }
+
+    const dir = join(options.directory ?? process.cwd(), dirName)
+    if (await exists(dir)) {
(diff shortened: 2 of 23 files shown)


src/lib/extensions/create.ts
```diff
@@ -0,0 +1,164 @@
...
+        const relative = [
+            ...segments.slice(0, -1),
+            targetName(segments.at(-1) ?? '', dirName),
+        ].join('/')
+        const destination = join(dir, relative)
+
+        await mkdir(dirname(destination), { recursive: true })
+        await writeFile(
+            destination,
+            fill(await readFile(join(templateDir, templateFile), 'utf8'), values),
```
doistbot: <a href="#">P2</a>  `--description` is substituted verbatim into `td-extension.json` (and the node template's `package.json`) without JSON escaping. A description containing `"`, `\`, or a newline produces invalid JSON, so `readJsonValue` silently ignores the manifest — dropping the starting `requires` range and description — and `td extension doctor` reports the file unreadable. Escape the value for JSON context, or reject JSON-breaking characters, instead of doing a raw string replacement.

src/lib/extensions/create.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    const template = options.template ?? available[0]
+    if (!available.includes(template)) {
+        throw new CliError('EXTENSION_TEMPLATE_INVALID', `There is no "${template}" template.`, {
+            hints: [`Available templates: ${available.join(', ')}.`],
+        })
+    }
+
+    const dir = join(options.directory ?? process.cwd(), dirName)
```
doistbot: <a href="#">P2</a>  This checks `dir` and only later creates parent paths recursively before writing files, so it is vulnerable to a check-then-use symlink race. In a shared writable CWD such as `/tmp`, an attacker can create `td-<name>` after `exists()` returns false and place (for example) `README.md` as a symlink to a victim-owned file; `writeFile()` then follows it and overwrites that target. Claim the destination atomically with a non-recursive `mkdir(dir)` (mapping `EEXIST` to `EXTENSION_ALREADY_EXISTS`) before writing, rather than probing it separately.

src/commands/extension/index.test.ts
```diff
@@ -77,6 +77,7 @@ describe('registerExtensionCommand', () => {
         expect(group).toBeDefined()
         expect(group?.aliases()).toContain('ext')
         expect(group?.commands.map((command) => command.name()).sort()).toEqual([
+            'create',
```
doistbot, line 90: <a href="#">P2</a>  This asserts `create` is registered, but nothing verifies that `templatesDir()` actually resolves to the shipped templates. If the `../../../templates/extension/` URL breaks (e.g. the command file moves or the depth changes), `td extension create` would register and then fail with "No templates found", while every test still passes — the shipped-template tests in `create.test.ts` recompute the same URL instead of calling this function. Add an assertion here (or nearby) that `templatesDir()` contains the `bash` and `node` directories.

doistbot (commented):
This PR scaffolds new extensions via `td extension create`, with `bash` and `node` templates shipped as real files, verified end to end against `td extension install` and the resulting extension commands.

Few things worth tightening:
- `--description` is substituted verbatim into `td-extension.json` and the node template's `package.json`; unescaped quotes, backslashes, or newlines yield invalid JSON, which silently drops manifest fields (e.g. `requires`) and trips up `extension doctor`. Escape or reject JSON-breaking characters.
- The exists-check-then-write flow has a symlink race in shared directories: an attacker could swap the destination path for a symlink between the check and `writeFile`, redirecting writes outside the extension dir. Claim the destination atomically with non-recursive `mkdir` (mapping `EEXIST` to `EXTENSION_ALREADY_EXISTS`) instead.
- Nothing asserts `templatesDir()` actually resolves to the shipped `bash`/`node` directories — the tests recompute the same URL, so a resolution break would pass CI while `extension create` fails at runtime. Add a registration-time assertion.
- A failure mid-render leaves a partially written destination directory behind with no cleanup.

<details>

- **<a href="#">P3</a> src/lib/extensions/create.ts:69**: The leftover-placeholder check runs on the filled string, so a user-supplied description containing `{{anything}}` (e.g. `--description 'uses {{NAME}}'`) trips it and scaffolding fails with "The template refers to {{NAME}}...", even though every template placeholder was resolved. Check for unresolved placeholders on the original template content, or make substituted values inert before the check.
- **<a href="#">P3</a> src/lib/extensions/create.ts:141**: When a template file fails mid-loop (e.g. `fill` rejects a leftover placeholder), files already written stay on disk inside the new directory. Retrying then fails with `EXTENSION_ALREADY_EXISTS` — "Choose another name, or remove the directory first" — which is confusing since the failure was the template's, not the user's. The shipped templates are covered by the placeholder test, but hosts bringing their own templates hit this on a typo. Consider reading and filling every template file into memory first, then writing, so the write loop can't fail halfway.
- **<a href="#">P3</a> src/lib/extensions/create.ts:138**: `relative(templateDir, join(entry.parentPath, entry.name))` from `node:path` strips the template prefix without relying on `templateDir.length + 1`. Use the standard API instead of the manual offset.
- **<a href="#">P3</a> src/lib/extensions/create.ts:92**: This validate-and-reserve sequence duplicates the same check in `preflight` (src/lib/extensions/install.ts): same code, message, and outcome. Extract a shared helper and have both call sites pass their own hint text, so `install` and `create` can't drift.
- **<a href="#">P3</a> src/lib/extensions/commands.ts:181**: `manager.reservedNames` already has the `reservedNames: () => Iterable<string>` shape, so the extra arrow is a redundant pass-through. Pass it directly: `reservedNames: manager.reservedNames`.
- **<a href="#">P3</a> src/lib/extensions/create.ts:55**: `listTemplates` swallows every `readdir` failure with `.catch(() => [])`, so a missing or unreadable `templatesDir` surfaces as `No templates found in ...` while the real I/O error is lost. Let the error propagate or wrap it in a `CliError` that preserves the cause, so a misconfigured host fails with an accurate message.
- **<a href="#">P3</a> templates/extension/node/README.md:32**: This says td runs `npm ci --omit=dev`, but a freshly scaffolded extension ships no lockfile, so `installDependencies` runs `npm install --omit=dev --no-package-lock` instead; `npm ci` only kicks in once the author commits a package-lock.json. Worth matching the actual behavior: "td runs `npm ci --omit=dev` when a lockfile is present, otherwise `npm install`".
- **<a href="#">P3</a> src/lib/extensions/create.test.ts:177**: This assertion can never fail. `createExtension` already throws `EXTENSION_TEMPLATE_INVALID` when `fill` encounters any `{{WORD}}` it cannot resolve, so any file that reaches this loop has no `\{\{\w+\}\}` left. Either drop the per-file check, or tighten it to something `fill` does not enforce (e.g. `not.toContain('{{')`) so it also catches space-separated or punctuation-containing placeholders.
- **<a href="#">P3</a> src/lib/extensions/commands.test.ts:279**: `create` is the only extension subcommand here without a `--json` test, even though it is a declared machine-readable contract and the sibling install/upgrade/remove/list commands all assert their JSON shape. Add a test that runs `extension create goals --json` and asserts the parsed `{ name, dirName, dir, template, files }` result.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F542&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab6219410-adf6-11f1-86ec-dd1d20cafbfd%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab6219410-adf6-11f1-86ec-dd1d20cafbfd&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/extensions/create.ts
scottlovegrove: Confirmed and fixed in 1d1c6f1 — this was the worst of the three, because the failure is silent:

```
$ td extension create quoted --description 'He said "hi"'
td-extension.json →  "description": "He said "hi"",
                     INVALID: Expected ',' or '}' after property value
```

An unreadable manifest is ignored rather than reported, so the starting `requires` range disappeared along with the description.

Values are now escaped for the file they land in — JSON string escaping for anything `.json`, raw elsewhere, since a README wants the text as typed. After:

```
description = "He said \"hi\" \\ and left"
requires    = {"td":">=5.3.2"}
```

I also took `{{DESCRIPTION}}` out of both executables rather than escaping it there. In the node template it sat inside a template literal, so a backtick or `${` in a description would have broken the generated JavaScript, and there is no file extension to key an escaper on. The description still reaches the manifest, `package.json` and the README, and the usage line in the executable is something an author edits anyway.

src/lib/extensions/create.ts
scottlovegrove: Fixed in 1d1c6f1, and your suggestion is the right shape — the probe is gone entirely rather than being tightened.

`mkdir(dir)` without `recursive` now claims the destination, with `EEXIST` mapped to `EXTENSION_ALREADY_EXISTS`, so the check and the claim are the same operation and there is no window between them.

One thing that fell out of fixing it: the write loop moved after the claim, and everything is read and filled before any of it is written. That closes a separate hole in the same area — a template that failed to render halfway through used to leave a partial directory, so the retry failed with "already exists" for something that was the template's fault rather than the user's. There is a test for that now.

src/commands/extension/index.test.ts
scottlovegrove: Right — the shipped-template tests recompute the same URL, so they would have gone on passing while `td extension create` failed at runtime. That is exactly the kind of gap worth pointing at.

Fixed in 1d1c6f1 with an assertion here that calls the function itself:

```ts
await expect(listTemplates(templatesDir())).resolves.toEqual(['bash', 'node'])
```

Kept on this side rather than in `create.test.ts`, since resolving the host's own templates is the host's concern — the library tests should stay runnable wherever that directory ends up.

gnapse (approved):

PR #543, opened 2026-09-11
fix: omit status suffixes from abbreviated user names

### LINKED ISSUE #508: formatUserShortName assumes `First Last`, so a status marker yields an initial from the status
Found while fixing #506 (surrogate splitting, PR #507). Same function, different and more frequent defect, so filing it separately rather than folding it into that PR.

`formatUserShortName` assumes a display name is `First Last` and takes the last whitespace-delimited token as the surname. Many display names carry a status marker instead, so the initial comes from the status rather than the name.

Ten shapes drawn from real workspace display names, anonymised, through shipped 5.2.1:

```
"Name | 🌴"                                       "Name \ud83c."    <- invalid, that's #506
"name (OOO Friday)"                               "name F."
"Firstname Surname (OOO Sept 4-7)"                "Firstname 4."
"Firstname Surname (OOO until Tuesday, Sept 8)"   "Firstname 8."
"Firstname Surname (working half days sept 1-4)"  "Firstname 1."
"Firstname Surname - OOO back Tue"                "Firstname T."
"Firstname OOO - Back on September 8"             "Firstname 8."
"Firstname Surname (Back 7 Sept)"                 "Firstname S."    <- from "Sept)", not the surname
"Firstname 🏖️ (🔙 31 Aug)"                        "Firstname A."
"Firstname Surname"                               "Firstname S."
```

One of ten is invalid. Eight of ten render an initial taken from a status word. One is correct. In five of them a real surname is sitting in the string, ignored because it is not the final token.

The marked line is the one worth attention: `S.` looks like a correct surname initial and comes from `Sept)`. So this cannot be caught by eye, which is likely why it has not been filed despite being the common case.

#506 is the visible symptom because it throws. This one is silent, and by these shapes considerably more frequent.

No fix proposed, since the right behaviour is a product question rather than a mechanical one. The options are not obviously equal: stripping a trailing parenthesised or dash-delimited segment before abbreviating would cover most of these, but `Firstname OOO - Back on September 8` has no surname at all, and `name (OOO Friday)` is lowercase and single-token underneath. Falling back to the first token alone may read better than a confidently wrong initial.

Happy to send a PR if you decide on the behaviour you want.

gnapse:
## Overview

Abbreviated assignee names currently take their initial from the final word, so `Ada Lovelace (OOO Friday)` appears as `Ada F.`. This removes common status suffixes before using the first name and the second token’s initial, following the web client’s token order.

Trailing parenthesized text, suffixes separated by spaced pipes or dashes, and a trailing standalone `OOO` are omitted. When the second token’s first grapheme has no letters, only the first name is shown: `Name | 🌴` becomes `Name`. Hyphenated names stay intact, and initials retain complete Unicode graphemes, including combining accents. These rules affect abbreviations only; parenthesized nicknames and decorative second tokens are intentionally omitted.

Closes #508.

## Verification

- All 1,930 tests pass, including all reported name shapes, accented initials, mixed suffixes, punctuation initials, and completed-task assignee output.
- Type-check, build, and formatting pass. Lint reports only the existing whitespace warning in `src/commands/user/aliases.ts`.
- Verified on Node 24.20.0.

### DIFF
--- a/src/lib/collaborators.ts
+++ b/src/lib/collaborators.ts
@@ -132,18 +132,18 @@ const nameSegmenter = new Intl.Segmenter('en', { granularity: 'grapheme' })
 export function formatUserShortName(fullName: string): string {
     // Require spaces around separators so hyphenated names stay intact.
     let name = fullName.trim().split(/\s+[|–—-]\s+/)[0]
-    // A display name can carry more than one trailing parenthesized suffix.
-    while (/\s+\([^()]*\)$/.test(name)) {
-        name = name.replace(/\s+\([^()]*\)$/, '').trimEnd()
+    // Remove mixed suffixes repeatedly, regardless of their order.
+    const statusSuffix = /\s+(?:\([^()]*\)|OOO)$/i
+    while (statusSuffix.test(name)) {
+        name = name.replace(statusSuffix, '').trimEnd()
     }
-    name = name.replace(/\s+OOO$/i, '')
 
     // Match the web client's first/second-token rule after removing statuses.
     const [firstName, secondName] = name.split(/\s+/)
-    if (!secondName || !/\p{L}/u.test(secondName)) return firstName
+    if (!secondName) return firstName
 
     const initial = nameSegmenter.segment(secondName)[Symbol.iterator]().next().value?.segment
-    return initial ? `${firstName} ${initial}.` : firstName
+    return initial && /\p{L}/u.test(initial) ? `${firstName} ${initial}.` : firstName
 }
 
 export interface FormatAssigneeOptions {
--- a/src/lib/collaborators.ts
+++ b/src/lib/collaborators.ts
@@ -2,7 +2,6 @@ import { isWorkspaceProject, type TodoistApi } from '@doist/todoist-sdk'
 import { getCurrentUserId, type Project, type Task } from './api/core.js'
 import { CliError } from './errors.js'
 import { extractId, isIdRef } from './refs.js'
-import { firstCodePoint } from './text.js'
 
 export interface CollaboratorInfo {
     id: string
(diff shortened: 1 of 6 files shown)


src/lib/collaborators.ts
```diff
@@ -127,16 +126,24 @@ export class CollaboratorCache {
...
-    const firstName = parts[0]
-    // By code point, not `[0]`: indexing by code unit takes half of an astral
-    // character, so a last name starting with an emoji yields a lone surrogate.
-    const lastInitial = firstCodePoint(parts[parts.length - 1])
-    return `${firstName} ${lastInitial}.`
+    name = name.replace(/\s+OOO$/i, '')
+
+    // Match the web client's first/second-token rule after removing statuses.
+    const [firstName, secondName] = name.split(/\s+/)
+    if (!secondName || !/\p{L}/u.test(secondName)) return firstName
```
doistbot: <a href="#">P2</a>  A second token that starts with punctuation but contains letters passes this check and becomes the initial. `Ada (nickname) Lovelace` now renders as `Ada (.` (the previous final-token logic returned `Ada L.`), and `Ada (nickname) OOO` also renders `Ada (.` because the trailing-`OOO` strip runs after the trailing-paren loop and leaves the middle nickname as the second token. Strip parenthesized tokens before tokenising (not only trailing ones), or test that the second token's first grapheme is a letter rather than any letter in the token.

doistbot (commented):
Nice cleanup — abbreviated assignee names now drop status suffixes (parenthesized text, spaced pipes/dashes, trailing OOO) and use grapheme-safe first/second-token initials, matching the web client, with solid edge-case handling for hyphenated names, combining accents, and astral characters.

Few things worth tightening:
- A second token that starts with punctuation but contains letters still becomes the initial, e.g. `Ada (nickname) Lovelace` renders as `Ada (.` — consider stripping parenthesized tokens before tokenising, or requiring the second token's first grapheme to be a letter.
- Relatedly, `Ada (nickname) OOO` also renders `Ada (.` because the trailing-`OOO` strip runs after the trailing-paren loop, leaving the middle nickname as the second token.
- Worth adding test coverage for non-trailing parenthesized nicknames once the fix lands.

<details>

- **<a href="#">P3</a> src/lib/collaborators.ts:5 (removed line)**: Removing this import leaves `firstCodePoint` in `src/lib/text.ts` with no production callers — only its own tests reference it. Since this PR switches to `Intl.Segmenter`, consider deleting the helper and its tests in the same change so the shared-utilities catalog stays accurate.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F543&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Acffeeca0-ae16-11f1-832b-20d510d02a25%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Acffeeca0-ae16-11f1-832b-20d510d02a25&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove:
I wonder if this should have gone in the SDK instead so it can be consumed by the MCP as well 🤔

gnapse:
> I wonder if this should have gone in the SDK instead so it can be consumed by the MCP as well 🤔

Probably yes. But I didn't think about it and I don't have time to look into it right now.

PR #554, opened 2026-09-21
fix(task): Strip leading `@` from `--labels` values

### LINKED ISSUE #550: `--labels` accepts an invalid `@`-prefixed label instead of stripping or rejecting it
### Summary
`td task add --labels "@Foo"` attaches the label literally as `@Foo`. Since `@` is display syntax and never part of a stored label name, this produces an **invalid** label: it renders as `@@Foo` in the UI and never appears in the UI's label-management list. The CLI accepts it silently — no strip, no warning, no error.

### Repro (td 5.3.6)
```
td task add "repro" --labels "@ZZTEST_Mode:Foo" --json
# -> task labels: ["@ZZTEST_Mode:Foo"]

td label list --json
# -> appears under sharedLabels as "@ZZTEST_Mode:Foo"
```
In the Todoist UI the task shows `@@ZZTEST_Mode:Foo`, and the label does **not** appear in the label-management list — confirming it is not a valid label.

### Actual
The CLI stores an invalid label name (leading `@`).

### The fix (maintainers' call)
A leading `@` in a `--labels` value is never valid. The CLI should not accept it as-is. Either option resolves it:
- **Strip** the leading `@` so `@Foo` and `Foo` resolve to the same label; or
- **Reject** with a clear error.

We don't have a strong preference — the key point is that silently storing `@Foo` is wrong.

### Note — inconsistency inside the CLI
`td label remove-shared "@ZZTEST_Mode:Foo"` already strips the `@` and looks up `ZZTEST_Mode:Foo`:
```
Error: LABEL_NOT_FOUND
Shared label "ZZTEST_Mode:Foo" not found.
```
So the CLI normalises a leading `@` on `remove-shared`, but not on the `--labels` apply path.

### Bundled skill guidance
The package ships a `todoist-cli` agent skill (`SKILL.md`). Until the CLI handles this, the skill should teach agents the rule explicitly: never pass a `@` to `--labels` (or an MCP `labels` array); the `@` belongs only in `td task quickadd` strings, where the parser strips it. Ideally the skill says to strip a leading `@` from any label value before applying.

### Environment
- `td` 5.3.6

engfragui:
## Summary

`td task add/update --labels "@Foo"` previously stored the label literally as `@Foo`. Since `@` is display-only syntax and never part of a stored label name, this produced an **invalid** label: it renders as `@@Foo` in the UI and never appears in label management. The CLI accepted it silently — no strip, no warning.

This was also inconsistent within the CLI: `td label remove-shared "@Foo"` already strips the `@`, but the `--labels` apply path did not.

## Fix

Chose the **strip** approach from the two the issue offered, matching how quickadd and `label remove-shared` already normalize a leading `@`.

- Added a shared `parseLabels()` helper (`src/commands/task/helpers.ts`) that splits on commas, trims, strips a leading `@`, and drops empty entries (bare `@`, trailing comma).
- `task add` and `task update` both route `--labels` through it. `@Foo` and `Foo` now resolve to the same label.

## Reference 

- Fixes #550

## Test Plan

- Checkout this branch
- Build todoist-cli locally

```
npm run build
```

- Create a task with label `@ZZTEST_Foo`

```
node dist/index.js task add "repro" --labels "@ZZTEST_Foo" --json
```

- [ ] Verify that the task has label `ZZTEST_Foo` (without `@`)

Screenshot 2026-09-21 at 08 59 32

- Create a task with label `ZZTEST_Foo`

```
node dist/index.js task add "repro2" --labels "ZZTEST_Foo" --json
```

- [ ] Verify that the task has label `ZZTEST_Foo` (without `@`)

Screenshot 2026-09-21 at 09 01 16

- Update a task to have labels `@ZZTEST_Work` and `ZZTEST_Urgent`

```
node dist/index.js task update id:<task-from-step-1> --labels "@ZZTEST_Work, ZZTEST_Urgent" --json
```

- [ ] Verify that the task has label `ZZTEST_Urgent` and `ZZTEST_Work` (without `@`)

Screenshot 2026-09-21 at 09 03 43


src/commands/task/update.ts
```diff
@@ -47,7 +47,7 @@ export async function updateTask(ref: string, options: UpdateOptions): Promise<v
     if (options.labels === false) {
         args.labels = []
     } else if (options.labels) {
-        args.labels = options.labels.split(',').map((l) => l.trim())
+        args.labels = parseLabels(options.labels)
```
doistbot: <a href="#">P2</a>  `parseLabels` can return `[]` for a truthy `--labels` value such as `--labels "@"` or `--labels ","`, and `updateTask` then assigns `args.labels = []`. For `task update`, an empty labels array is the same payload as `--no-labels`, so these inputs now silently remove all labels instead of being ignored or rejected. The dry-run preview also prints `Labels: ` (blank) rather than `(remove all)`, hiding the destructive effect. Only set `args.labels` when `parseLabels` returns a non-empty array (or error on an all-empty value), and keep the dry-run preview consistent with that behavior.

doistbot (commented):
This PR adds a shared `parseLabels()` helper so `task add`/`task update` strip a leading `@` from `--labels` values, fixing the invalid-label issue in a way consistent with quickadd and `label remove-shared`, with tests, dry-run preview updates, and skill content kept in sync.

Few things worth tightening:
- All-empty `--labels` values (e.g. `"@"` or `","`) currently produce `labels: []` on `task update`, which is the same payload as `--no-labels` — these inputs silently remove all labels; consider ignoring/rejecting them instead, and only setting `args.labels` when the parsed list is non-empty.
- Keep the dry-run preview consistent with that fix so the destructive effect isn't hidden behind a blank `Labels:` line.

<details>

- **<a href="#">P3</a> src/commands/task/helpers.ts:13**: `parseLabels` re-implements the leading-`@` normalization already present in `src/commands/label/helpers.ts` (three separate `startsWith('@') ? …slice(1)` sites). The “`@` is display-only” rule now lives in two command-group helper files. Extract one shared helper (e.g. `stripLabelAtPrefix` in `src/lib`) and call it from both places so the normalization can’t drift between commands.
- **<a href="#">P3</a> src/commands/task/task.test.ts:1862**: This test is nested under `describe('task update --no-labels')`, but it exercises `--labels` preview, not `--no-labels`. That makes the suite name misleading and hides it from the existing `task --dry-run` preview suite, where it belongs. Move it there (or split it out).
- **<a href="#">P3</a> src/commands/task/add.ts:148**: The add dry-run preview now joins the parsed labels (`args.labels?.join(', ')`), so `task add --dry-run --labels "@work"` should display `work` without the `@`. No test covers this path — the update dry-run preview got a test, but add can regress independently. Extend the existing `task add --dry-run previews without calling API` test to assert the normalized label output.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftodoist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftodoist-cli%2Fpull%2F554&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A12345590-b58a-11f1-84f6-c31fa2455b03%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A12345590-b58a-11f1-84f6-c31fa2455b03&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

engfragui:
Thanks for the review — addressed in 8d51ce6:

**Main points**
- **All-empty `--labels` silently clearing labels:** `task add`/`task update` now only set `args.labels` when at least one real label survives parsing, so values like `"@"` or `","` are a no-op instead of matching the `--no-labels` payload. Clearing labels stays exclusively `--no-labels`. Added tests for both paths.
- **Dry-run consistency:** the preview already joins `args.labels`, so with the no-op fix an all-empty value simply omits the `Labels:` line (no change shown), consistent with the non-destructive behaviour.

**Follow-up notes**
- **P3 (duplicated normalization):** extracted `stripLabelAtPrefix` into `src/lib/labels.ts` (with its own unit test) and reused it from `task` `parseLabels` and all three `label/helpers.ts` sites, so the "`@` is display-only" rule lives in one place.
- **P3 (misplaced test):** moved the `--labels` preview test out of the `task update --no-labels` suite into the `task --dry-run` suite.
- **P3 (add dry-run uncovered):** added a `task add --dry-run` test asserting the `@` is stripped in the preview.

Full suite green (1941 tests), type-check and lint/format clean.

src/commands/task/helpers.ts
```diff
@@ -1,8 +1,18 @@
...
+import { stripLabelAtPrefix } from '../../lib/labels.js'
 
 export type DurationArgs = { duration?: number; durationUnit?: 'minute' | 'day' }
 
+// Parses a comma-separated --labels value into label names, stripping the
+// display-only leading `@` from each (see stripLabelAtPrefix).
+export function parseLabels(value: string): string[] {
+    return value
+        .split(',')
+        .map((label) => stripLabelAtPrefix(label.trim()).trim())
```
gnapse: Maybe `stripLabelAtPrefix` should take care of trimming its input. That way we do not need to remember to do it when calling it.

Also, after trimming the input, we do not need to trim its output.

gnapse (approved):

src/commands/task/helpers.ts
engfragui: Makes sense, done in 08bd0e5 ✅

